Issue #341, opened 2015-01-05
http connections broken without ACE


paulfitz:
Commit b2ef02864ad1bef13e6a643fdcb56592b84e0731 appears to have broken http connections for yarp compiled without ACE.  To replicate:

 * Compile yarp without ACE.
 * Start the name server.
 * Follow the link mentioned on "Name server can be browsed at" line.
 * Result is blank (should be a webpage, compare with result when compiled with ACE).

paulfitz:
Result is blank in chrome, or "The connection to the server was reset while the page was loading." in firefox.

drdanz:
Actually on my computer I can see the webpage for a second before the connection is reset, and using wget I get the html, but it hangs.

It looks like that the cause is the ``l_linger`` value of the ``linger`` struct that is set to ``0``.

```c++
    struct linger lval;
    lval.l_onoff = 1;
    lval.l_linger = 0;
    stream.set_option (SOL_SOCKET, SO_LINGER, &lval, sizeof(linger));
```
Changing it to any 

According to the man page

> SO_LINGER
> Sets or gets the SO_LINGER option.  The argument is a linger structure.
>
>```c
>      struct linger {
>          int l_onoff;    /* linger active */
>          int l_linger;   /* how many seconds to linger for */
>      };
>```
>
> When  enabled, a close(2) or shutdown(2) will not return until all queued messages for the socket have been successfully sent or the linger timeout has been reached.  Otherwise, the call returns immediately and the closing is done in the background.  When the socket is closed as part of exit(2), it always lingers in the background.


Setting l_onoff to ``1`` and l_linger to ``0`` causes the connection to be aborted and any pending data is immediately discarded when ``close()`` or ``shutdown()`` are called, and therefore the webpage is not displayed by the browser. I'm not sure of the reason why this works for ACE and does not work without ACE, I was expecting it to be a pure wrapper of the system call. Also I'm not sure of the reason why we enable it...
I think we can either remove the ``SO_LINGER`` option for SKIP_ACE (or for both cases) or change the ``l_linger`` value to a non-zero value (either for SKIP_ACE only or for both cases). @paulfitz opinion?

paulfitz:
[...]

PR #342, opened 2015-01-05
[test] add integration test for http protocol against curl

paulfitz:
This is a test for the problem described in #341. I wasn't able to replicate it with a yarp-to-yarp connection, I think yarp is too forgiving, so I show it here against `curl`. This test currently fails for me when compiling without ACE. /cc @drdanz


paulfitz:
added a fix for #341 (stripped non/dysfunctional linger code) and activated the http integration test on travis

PR #345, opened 2015-01-09
qos for yarp

apaikan:
seems to be ready!


lornat75:
We should first check why travis is failing?

apaikan:
why all the #ifdef YARP_HAS_ACE are removed from DgramTwoWayStream.cpp? :o

drdanz:
Should be fixed now... Probably a failed rebase

paulfitz:
Nice clean patch, really interesting feature!

PR #349, opened 2015-01-13
setRefTorque/s implemented using streaming instead of RPC

francesco-romano:
This PR should change how the `setRefTorque` and `setRefTorques` handle the request passing from RPC to streaming.

### DIFF
--- a/src/libYARP_dev/include/yarp/dev/ControlBoardInterfaces.h
+++ b/src/libYARP_dev/include/yarp/dev/ControlBoardInterfaces.h
@@ -807,5 +807,8 @@ public:
 #define VOCAB_IMP_PARAM   VOCAB3('i','p','r')
 #define VOCAB_IMP_OFFSET  VOCAB3('i','o','f')
 
+#define VOCAB_TORQUES_DIRECTS VOCAB4('d','t','q','s') //This implements the setRefTorques for the whole part
+#define VOCAB_TORQUES_DIRECT VOCAB3('d','t','q') //This implements the setRefTorque for a single joint
+#define VOCAB_TORQUES_DIRECT_GROUP VOCAB4('d','t','q','g') //This implements the setRefTorques with joint list
 #endif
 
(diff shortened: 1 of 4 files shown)


francesco-romano:
@lornat75 @barbalberto @traversaro

Do you think this is ok?
Should we / I implement also the get? Or they are not needed?

lornat75:
Fine with me. get should be in the new streaming port @barbalberto

randaz81:
Please merge after IMotorEncoder branch

2015-01-13 22:07 GMT+01:00 Lorenzo Natale <notifications@github.com>:

barbalberto:
the getTorque is already in the streming port, the getRefTorque is not but I guess it is not important, right?

francesco-romano:
@barbalberto it was just a question in order to have a "clean" patch. I agree with you.

francesco-romano:
@randaz81 I will wait for https://github.com/robotology/yarp/pull/352 then I will rebase.

francesco-romano:
I rebased as I promised.

Waiting to be merged.

barbalberto:
This change willl create a problem similar to #343 since the controlboardwrapper'1' has not been updated. @francesco-romano can you update that one as well, please?

Then, if 2 PCs have different yarp version, the set torque will not work, and the user will see 
  yError("Unrecognized message while receiving on command port (%s)\n",str.c_str());
on the receiver (robot) side. Streaming gives no feedback so there is no way I can think of to warn the sender

francesco-romano:
@barbalberto @lornat75 
The old control board wrapper does not implement the `setRefTorque*` methods (they are all `return false`).
What should we do?

Regarding the second question: different yarp versions.
I agree with you. This is an issue. And it happened also to me while testing this patch.
I don't know how to solve this. Maybe there should be a sort of protocol versioning and some way to check this.

randaz81:
I'd reconsider the idea of removing the already deprecated
controlboardwrapper1 and force the users to an update (giving a deadline of
a 1-2months for example), it is becoming really complicate to maintain it.

2015-01-16 10:44 GMT+01:00 Francesco Romano <notifications@github.com>:

lornat75:
@francesco-romano If the old controlboard does not implement setRefTorque* methods it means that existing code that uses the old controlboard does not need them, so I would not consider this a problem.

We should implement the protocol check anyway. I believe it is not difficult to do.

barbalberto:
Yes, I assumed the old wrapper had the torque interface implemented, but waas not the case, my fault. So I'll accept the pull request.
Version check of the protocol is still an issue, but will open a dedicated thread.

lornat75:
For protocol check: see https://github.com/robotology/yarp/pull/349#event-220076524

PR #361, opened 2015-01-16
added mutex to ensure multi-thread safety

randaz81:
The current version of extendedIntputStatePort is not thread safe.
Examining the two following lines, the contents of the shared vector last_wholePart may be modified by another thread before the std::copy(), or maybe during the copy.

bool ret = extendedIntputStatePort.getLast(last_wholePart, lastStamp, localArrivalTime);
std::copy(last_wholePart.jointVelocity.begin(), last_wholePart.jointVelocity.end(), spds);

In the proposed patch, i added a mutex to ensure multi-thread safety.

### DIFF
--- a/src/libYARP_dev/src/modules/RemoteControlBoard/RemoteControlBoard.cpp
+++ b/src/libYARP_dev/src/modules/RemoteControlBoard/RemoteControlBoard.cpp
@@ -269,6 +269,7 @@ protected:
     // from the YARP .thrift file
 //  yarp::os::PortReaderBuffer<jointData>           extendedInputState_buffer;  // Buffer storing new data
     StateExtendedInputPort                          extendedIntputStatePort;  // Buffered port storing new data
+    Semaphore extendedPortMutex;
     jointData last_singleJoint;     // tmp to store last received data for a particular joint
 //    yarp::os::Port extendedIntputStatePort;         // Port /stateExt:i reading the state of the joints
     jointData last_wholePart;         // tmp to store last received data for whole part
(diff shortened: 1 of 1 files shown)


barbalberto:
Mutex is already contained in the StateExtendedInputPort class: the writing in the structure by *onRead* callback and getting back the data through the *getLast* function internally calls the mutex.
The last_wholePart variable used in the copy is an internal helper data, not shared with others and not touched when receiving data from the port.
If you think it is not enough, we can merge this. Maybe I'm missing something...

barbalberto:
Ok, I after a small chat I got your point. Let's add this mutex, then I'll think of a way to optimize the code and place a single mutex to to the job instead of having 2 of them.

This mutex shall:
- protect the writing from the onRead  (ok)
- protect the reading from the getLast (ok)

- protect the last_wholePart from concurrent access in case the module instantiating the remoteControlBoard has multiple threads calling different getXXX  (missing part added by the patch)

randaz81:
I was thinking the same, then I noticed that unfortunately the mutex internal to getLast() is not sufficient to protect data consistency in case of modules that perform multiple concurrents "gets". 
For example, controlBoardDataDumper open multiple interfaces on multiple concurrent threads, each of them requesting for a particular data type (encoders, torques etc). 
After some time, the controlBoardDataDumper crashes, complaining that the last_wholePart vector changed size (this make me think that the crash happened because during the std::copy(), another getLast() just completed)

barbalberto:
Or, another solution could be get rid of the temp struct last_wholePart and have different functions returning the data of interest, like getLastEncoder, getLastVelocity ... 
I'll think a bit, thanks @randaz81 for noticing and patching +1

PR #365, opened 2015-01-20
[YARP_OS] add hooks for checking heap allocation in regression tests

paulfitz:
This is some support for memory checks I'll be adding in later PRs, basically adding tests that yarp does not allocate memory during performance-sensitive operations.

It relies on overriding new/delete.  This appears to conflict with something ACE is doing.  Reserving these tests for a no-ace build is fine for my purposes.  None of this code will be run unless YARP is compiled with a special `YARP_TEST_HEAP` flag (marked as advanced).

### DIFF
--- a/src/libYARP_OS/CMakeLists.txt
+++ b/src/libYARP_OS/CMakeLists.txt
@@ -385,6 +385,12 @@ if (YARP_USE_READLINE)
     target_link_libraries(YARP_OS LINK_PRIVATE ${Readline_LIBRARY})
 endif ()
 
+option(YARP_TEST_HEAP "Replace new/delete operators with test harness (may be incompatible with ACE)" FALSE)
+mark_as_advanced(YARP_TEST_HEAP)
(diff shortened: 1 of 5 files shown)


drdanz:
@paulfitz what about malloc/free allocation? There are a few around YARP_OS, it might be worth checking for these as well...

paulfitz:
@drdanz I did look at catching malloc/free but the easy-to-do method was deprecated and the blessed method was more awkward than I wanted to get into. However maybe I missed something. It seemed easier to remove any malloc/frees in the code I'll be testing (I don't think there are any in those parts but will check)

drdanz:
@paulfitz Have you seen [this](https://stackoverflow.com/questions/27137527/overload-symbols-of-running-process-ld-preload-attachement/27304692#27304692)?
Also [this tool (Heaptrack)](http://milianw.de/blog/heaptrack-a-heap-memory-profiler-for-linux) looks interesting.

PR #371, opened 2015-01-21
[YARP_OS] reduce memory churn for envelope messages 

paulfitz:
This eliminates one set of unnecessary memory allocations associated with sending envelope messages (e.g. timestamps), see #353. There's more to do for that issue, this is just one piece.  The PR also adds a bunch of documentation to the `BufferedConnectionWriter` class, since I needed to upgrade it a bit.


drdanz:
Looks good to me... thanks.

PR #376, opened 2015-01-23
[CMake] Add YARP_ADD_IDL Function

drdanz:
Add a new ``YARP_ADD_IDL`` function that takes one or more IDL files and generate code at build time.
Files will be regenerated whenever the IDL file changes.

```cmake
yarp_add_idl(<var> <file> [file [...]])
```

The ``<var>`` variable, will contain the generated files, and can be added to the an ``add_executable`` or ``add_library call``. For example:

```cmake
set(THRIFT_FILES file1.thrift
                 file2.msg
                 file3.srv)
yarp_add_idl(THRIFT_GEN_FILES ${THRIFT_FILES})
add_executable(foo main.cpp ${THRIFT_GEN_FILES})
```

### DIFF
--- a/conf/YarpIDL.cmake
+++ b/conf/YarpIDL.cmake
@@ -1,5 +1,6 @@
-# Copyright: (C) 2012 IITRBCS
-# Authors: Elena Ceseracciu, Paul Fitzpatrick
+# Copyright (C)  2012 Robotics Brain and Cognitive Sciences, Istituto Italiano di Tecnologia
+# Copyright (C)  2014 iCub Facility, Istituto Italiano di Tecnologia
+# Authors: Elena Ceseracciu, Paul Fitzpatrick, Daniele E. Domenichelli
 # CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
 
 # yarp_idl_to_dir
@@ -12,10 +13,29 @@
 #   yarp_idl_to_dir(foo.thrift foo)
 #   yarp_idl_to_dir(foo.thrift foo SOURCES HEADERS)
 #   yarp_idl_to_dir(foo.thrift foo SOURCES HEADERS INCLUDE_PATHS)
+#
+#
+# yarp_add_idl
+# ------------
+#
+# Take one or more IDL files and generate code at build time.
+# Files will be regenerated whenever the IDL file changes.
+#
+#   yarp_add_idl(<var> <file> [file [...]])
+#
+# The <var> variable, will contain the generated files, and can be
+# added to the an add_executable or add_library call. For example:
+#
+#   set(THRIFT_FILES file1.thrift
+#                    file2.msg
+#                    file3.srv)
+#   yarp_add_idl(THRIFT_GEN_FILES ${THRIFT_FILES})
+#   add_executable(foo main.cpp ${THRIFT_GEN_FILES})
+
 
 
 # Avoid multiple inclusions of this file
-if(COMMAND yarp_idl_to_dir)
+if(COMMAND yarp_add_idl)
   return()
 endif()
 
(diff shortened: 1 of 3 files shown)


paulfitz:
Looks pretty nifty, thanks for doing this. Style note, I'd feel better if the per-idl logic were factored out into yarp_idl_${family}_to_file_lists helper functions, the little parsers feel distracting in there.

paulfitz:
also, would be good to add some tests, see `src/idls/*/tests/demo/` (when testing locally these tests need to be turned on explicitly with `TEST_yarpidl_*` since they may not work under an IDE).

drdanz:
> I'd feel better if the per-idl logic were factored out into yarp_idl_${family}_to_file_lists helper functions, the little parsers feel distracting in there.

Looks reasonable, I will do it...

> also, would be good to add some tests, see src/idls/*/tests/demo/ (when testing locally these tests need to be turned on explicitly with TEST_yarpidl_* since they may not work under an IDE).

Do you want me to replace the current tests, or just to add more? Anyway I tested them with those demos and with a few more examples...

BTW I forgot to mention that the ros version is still failing sometimes for the TimeTick/Header issue we discussed on IRC. We might want to fix this issue as well before merging.

paulfitz:
I'd say add rather than replace, since there's been no talk yet of deprecating yarp_idl_to_dir. The additions could just repeat at least one of the tests you did offline. This is just to protect against future unintentional breakage of the macro.

Just to make sure I understand, the yarp_add_idl macro would be for internal use only, is that correct? If put in the hands of general users, It would miss stuff from nested ROS dependencies (ordinarily found via rosmsg) and thrift's include feature, unless I'm missing something myself.

drdanz:
I'd like to use it externally too... If we want it internal only, then it should be in some other file.
The includes from thrift work for me, I tried and if I understand it correctly, the files for the included thrift file are *not* generated, but just used as if you had them. So as long as you add both the includer and the included file to the macro, everything should work.
I don't know about the ROS nested stuff, I should do more tests...

paulfitz:
the thrift includes I mean are like this one http://stackoverflow.com/questions/20872683/how-to-include-thrift-files-from-different-modules/20873187#20873187 - sorry for the ambiguity, I don't mean the `yarp.includefile` thingie.

drdanz:
Yes, that works for me because the included files are read and used inside the generated source code, but no extra files are generated...

drdanz:
Updated.
I also fixed the TimeTick/Header by parsing the file, even though I'd recommend to move them into some YARP library in a second time.

paulfitz:
I'm happy with merging this for internal use in building yarp. But @drdanz you say that if it is for internal use only, you'd want it in another file.

I'm not yet happy with merging this for external use by users of yarp. Two things I don't yet understand support for.

 * Thrift include files. Included files could contain arbitrary thrift definitions, leading to new generated files. I feel like we are talking at cross-purposes about this somehow. Yarp's unit tests are not testing includes since yarp modifications to thrift don't touch them at all, so you won't see them there. I could add some if that will help us converge.
 * Ros types can refer to other types in any ros package, and those references will also lead to new generated files. You won't see this in yarp's unit tests, since it can't rely on Ros being present, but the integration tests should be checking this.

paulfitz:
Btw I'm ok with other people pushing the merge button, since in practice no user will notice this macro until we advertise it - just saying why I personally haven't done it yet :-)

drdanz:
@paulfitz do you mind adding some tests to the demo covering the cases you described?

lornat75:
This feature looks reasonable to me, @paulfitz I do not understand your motivations against accepting it, the discussion is mixed between the feature itself and tests (sorry if I miss something).

paulfitz:
@lornat75 I believe the yarp_add_idl macro may fail to correctly predict the files that will eventually be generated in the cases I listed, leading to build errors. If used internally during yarp build, we can simply take care how we use it. Not willing to merge since the macro appears to be intended for external use. 

@drdanz requests for tests is sensible and will help clarify discussion, I'll do that as soon as I can.

elen4:
If I remember correctly, the Thrift generator indeed does not generate code for included files, it expects them to be explicitly generated by the user and found at compile-time. I don't know if this is the intended behaviour or a bug. I think that it would make sense to have the same behaviour as yarpidl_rosmsg, which instead might end up generating other files, as @paulfitz mentioned, if ROS is present (I am not sure what happens if ROS is not present, I guess it breaks the build - not very nice, but this is unrelated to the current issue)

lornat75:
Wouldn't be enough to specify in the documentation that the macro does not resolve nested include files and that you have to manually list all files there?

drdanz:
> @drdanz requests for tests is sensible and will help clarify discussion, I'll do that as soon as I can.

Also corner cases that could confuse the parser are appreciated ;)

paulfitz:
For the ROS case, there's a test in https://github.com/robotology/yarp/tree/master/tests/integration/ros. You need to have ROS installed to compile this. The .msg file lists 4 types, and at least 2 others not mentioned (`sensor_msgs/RegionOfInterest` and `sensor_msgs/PointField`) get pulled in as dependencies.

For the thrift case, @elen4 is right, no code is generated for types in includes unless `-r` is specified to thrift, which we don't currently do (I was assuming the opposite default). I think that `-r` will be what the user wants in several situations, but it would be unfair to block this PR based on that, so I waive my thrift-based objection.

paulfitz:
I think the code in this PR could be good when combined with a method for managing IDL code across packages.

For example, a package `Foo` could define (in the cmake sense) a `Foo_IDLS` variable (along with the usual `Foo_INCLUDE_DIRS`, `Foo_LIBRARIES` etc). Then, the IDLs in that package could be made available for reference from local IDLs by doing:
```cmake
include_idls(${Foo_IDLS})
```

drdanz:
Sorry, I didn't have time to work on this PR, I will do it as soon as possible.

PR #377, opened 2015-01-26
yarpdatadumper and yarpdataplayer

drdanz:
* Import dataDumper and dataSetPlayer from robotology/icub-main repository preserving history (renamed yarpdatadumper and yarpdataplayer-gtk)
* Import yarpdataplayer-qt (thanks @aitek4iit)
* Fix build system, warnings and documentation


src/yarpdataplayer-gtk/include/msvc/dirent.h
```diff
@@ -0,0 +1,372 @@
+/*****************************************************************************
```
paulfitz: Can this be rationalized with existing dirent stuff? E.g. ACE has dirent (and ACE is required to compile yarp with MSVC), other platforms have it natively, and @apaikan also has a copy of the Ronkko implementation at `src/libYARP_manager/include/yarp/manager/ymm-dir.h`

src/yarpdataplayer-qt/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+  set(yarpdataplayer_qt_THRIFT_FILES dataSetPlayer.thrift)
+  if(MSVC)
+    set(yarpdataplayer_qt_RC_FILES QtDataSetPlayer_resource.rc)
+  endif()
+
+  qt5_add_resources(yarpdataplayer_qt_QRC_GEN_SRCS ${yarpdataplayer_qt_QRC_FILES})
+  qt5_wrap_ui(yarpdataplayer_qt_UI_GEN_SRCS ${yarpdataplayer_qt_UI_FILES})
+
+  include(YarpIDL)
+  yarp_idl_to_dir(${yarpdataplayer_qt_THRIFT_FILES}
```
paulfitz: @drdanz weren't you going to substitute in your new yarp_all_idl or whatever it is caused, and then not include the generated code?

src/yarpdataplayer-qt/msvc/dirent.h
```diff
@@ -0,0 +1,372 @@
+/*****************************************************************************
```
paulfitz: another copy of dirent??

src/yarpdataplayer-qt/utils.cpp
```diff
@@ -0,0 +1,427 @@
+/*
```
paulfitz: Isn't this code in the -gtk player too?

src/yarpdataplayer-qt/CMakeLists.txt
drdanz: Yes, but I opened the merge-request since I'm waiting for #376 to be merged to start the review and to check if it is building on all platforms... :)

paulfitz:
I'd be ok with merging this. Not super happy about the code duplication related to dirent though, so if you have time to clean that up it'd be better. The level of gtk/qt copy/pasting is a bit icky (at least at a quick glance) but on the assumption that gtk will be going away at some point, that'll be a problem that eventually fixes itself.

drdanz:
The dirent stuff was just there, I simply imported, I'd be happy if someone wants to remove it, though.
About the duplicate code, I agree it is redundant but I hope that the GTK code will disappear one day...

PR #389, opened 2015-01-30
[yarpdataplayer-gtk] Do not build without opencv

### LINKED ISSUE #388: yarpdataplayer-gtk compile failure without opencv
yarpdataplayer-gtk tries and fails to compile if opencv is not present

```
[ 86%] Building CXX object src/yarpdataplayer-gtk/CMakeFiles/yarpdataplayer-gtk.dir/src/main.cpp.o
In file included from /root/yarp2/src/yarpdataplayer-gtk/include/utils.h:32:0,
                 from /root/yarp2/src/yarpdataplayer-gtk/include/main_window.h:26,
                 from /root/yarp2/src/yarpdataplayer-gtk/src/main.cpp:27:
/root/yarp2/src/yarpdataplayer-gtk/include/worker.h:22:16: fatal error: cv.h: No such file or directory
compilation terminated.
```

drdanz:
Fixes #388

@paulfitz I cannot test the fix right now (working with the smartphone) Can you please test it and eventually merge it?
If it works, I think that the same needs to be done for the qt version...

### DIFF
--- a/src/yarpdataplayer-gtk/CMakeLists.txt
+++ b/src/yarpdataplayer-gtk/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Authors: Vadim Tikhanoff
 # CopyPolicy: Released under the terms of the GNU GPL v2.0.
 
-if(CREATE_YARPDATAPLAYER AND YARP_HAS_GTK2)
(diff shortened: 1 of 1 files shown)


paulfitz:
Tthanks! Works for me.

PR #396, opened 2015-02-05
Added inclusion of locale.h.

francesco-romano:
It defines LC_NUMERIC and other functions and types

I don't know which of the `SKIP_ACE=True` and `YARP_USE_READLINE=False` is causing this, but I could not compile without including `#include <clocale>`

### DIFF
--- a/src/libYARP_OS/src/BottleImpl.cpp
+++ b/src/libYARP_OS/src/BottleImpl.cpp
@@ -21,6 +21,8 @@
 #include <yarp/os/impl/PlatformStdlib.h>
 #include <yarp/os/impl/PlatformStdio.h>
 
+#include <clocale>
+
 using namespace yarp::os::impl;
 using namespace yarp::os;
 using namespace yarp::os::impl;
@@ -715,7 +717,6 @@ String StoreDouble::toStringFlex() const {
 
     // YARP Bug 2526259: Locale settings influence YARP behavior
     // Need to deal with alternate versions of the decimal point.
-#ifdef LC_NUMERIC
  	struct lconv * lc=localeconv();
  	size_t offset = str.find(lc->decimal_point);
  	if (offset!=String::npos){
@@ -723,12 +724,6 @@ String StoreDouble::toStringFlex() const {
  	} else {
         str += ".0";
     }
-#else
-    // maintain old YARP behavior if locale info not available
-    if (YARP_STRSTR(str,".")==String::npos) {
-        str += ".0";
-    }
-#endif
 
     int ct = 0;
     for (size_t i=str.length(); i>=1; i--) {
@@ -750,7 +745,6 @@ String StoreDouble::toStringFlex() const {
 void StoreDouble::fromString(const String& src) {
     // YARP Bug 2526259: Locale settings influence YARP behavior
     // Need to deal with alternate versions of the decimal point.
-#ifdef LC_NUMERIC
     String tmp = src;
     size_t offset = tmp.find(".");
     if (offset!=String::npos) {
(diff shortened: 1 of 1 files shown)


drdanz:
@francesco-romano sorry, I don't see the need for this, I tried all of the combinations of these 2 flags and all of them work for me... All ``LC_NUMERIC`` related parts are like this:

```c++
// YARP Bug 2526259: Locale settings influence YARP behavior
// Need to deal with alternate versions of the decimal point.
#ifdef LC_NUMERIC
  // ...
#else
  // ...
#endif
```

What exactly is your error here? On which platform/compiler?

francesco-romano:
Sorry for the few details I provided.
This is my cmake command:
```bash 
cmake .. -GXcode -DSKIP_ACE:BOOL=TRUE -DCREATE_SHARED_LIBRARY:BOOL=FALSE -DYARP_USE_READLINE:BOOL=FALSE -DYARP_WRAP_STL_STRING:BOOL=FALSE
```

The error is
~~~
yarp/src/libYARP_OS/src/BottleImpl.cpp:728:9: Use of undeclared identifier 'YARP_STRSTR'
~~~

drdanz:
Ok, I still cannot reproduce it, but the real problem is not that <clocale> is missing but that ``YARP_STRSTR`` is not defined anywhere and this ``#else``

```c++
#ifdef LC_NUMERIC
 // ...
#else
    // maintain old YARP behavior if locale info not available
    if (YARP_STRSTR(str,".")==String::npos) {
        str += ".0";
    }
#endif
```

is not reached in any of our tests.

aa2b36e9385137eab8f6edb858c59cdfb5397833 seem to be the commit that caused this...

Anyway, this doesn't seem a "complete" solution to the problem.
@paulfitz what do you think, should we just include ``clocale`` and remove the ``#ifdef`` or restore the ``YARP_STRSTR`` definition?

Also ``YARP_STRSTR`` is used in ``src/yarpserver/src/server_test.cpp``, that is never built because included in a ``if(TEST_BUILD)`` and the TEST_BUILD variable is not defined anywhere, running ``-DTEST_BUILD=ON`` shows that this file is full of errors...

francesco-romano:
@drdanz Yes, the error is the missing definition. 
But the problem is that it goes on the `#else` because `LC_NUMERIC` is not defined, which seems quite strange to me.
Do you think `locale` should not be defined?

drdanz:
@francesco-romano If it is always defined, I don't see the reason for keeping the ``#ifdef``...

francesco-romano:
@drdanz I don't get it. That macro is defined in `locale.h`. If it is not defined it means that the header is not included. Am I right?
So the question is.. Is this wanted or not?

drdanz:
@francesco-romano If we include ``locale.h`` then the LC_NUMERIC macro is always defined and therefore this ``#ifdef #else #endif`` does not make sense because the ``#else`` part is never going to be built... So I'm ok with including ``locale.h`` if we remove this ``#ifdef #else #endif``, otherwise the right fix is to fix the else part...

paulfitz:
good catch @drdanz.  All remaining instances of `YARP_STRSTR` are strays. For the BottleImpl.cpp one, stripping that code and including `locale.h` looks right. Everything related to `server_test.cpp` can be stripped too, that is old and unused, there are far better tests of the server in the regular yarp tests now.

francesco-romano:
@drdanz  @paulfitz  I removed the `YARP_STRSTR` call in `BottleImpl.cpp`

paulfitz:
thanks @francesco-romano. With `clocale` included, the `#ifdef LC_NUMERIC` is not needed, it will always be defined - could you strip that stuff out too if you have a chance?

francesco-romano:
@paulfitz done

paulfitz:
Thanks @francesco-romano!

drdanz:
@paulfitz I updated to ``yarp::os::Contact`` and fixed the YARP_STRSTR in ``server_test``. I don't know if it was worth, but it took just a few minutes...
Anyway at the moment it is not executed during tests, I don't know if it is worth including it.

PR #397, opened 2015-02-05
Added a setNameServerContact 

francesco-romano:
It is now possible to explicitly set the server address without using the detect method.

This addresses https://github.com/robotology/yarp/issues/394

### DIFF
--- a/src/libYARP_OS/include/yarp/os/Contact.h
+++ b/src/libYARP_OS/include/yarp/os/Contact.h
@@ -81,6 +81,14 @@ public:
     Contact addHost(const ConstString& host) const;
 
     /**
+     * Returns a new Contact with the previous information plus the updated
+     * port number
+     * @param portnumber port number
+     * @return a new contact the the port number updated
+     */
+    Contact addPort(int portnumber) const;
+
+    /**
      * Constructor for a socket contact.
      * Creates an unnamed contact, with information about
      * how to reach it using socket communication.
@@ -231,6 +239,19 @@ public:
     float getTimeout() const;
 
     ConstString getRegName() const;
+
+    /**
+     * Set the host to be the input parameter
+     * @param the new host
+     */
+    void setHost(const ConstString& host);
+
+    /**
+     * Set the port to be the input parameter
+     * @param the new port
+     */
+    void setPort(int port);
+
 
 private:
     ConstString regName;
--- a/src/libYARP_OS/include/yarp/os/Network.h
+++ b/src/libYARP_OS/include/yarp/os/Network.h
@@ -530,6 +530,14 @@ public:
                                     bool& scanNeeded,
                                     bool& serverUsed);
 
+    /** 
+     * Set explicitly the nameserver information 
+     *
+     * @param nameServerContact the NameServer contact information (e.g. IP, port)
+     * @return true if succeed. False otherwise
+     */
+    static bool setNameServerContact(Contact &nameServerContact);
+
     /**
      *
      * Search for a configuration file in YARP's standard config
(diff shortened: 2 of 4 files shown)


francesco-romano:
@paulfitz first implementation.
Feedbacks are welcome.

BTW: A `Contact` has a lot of setters, but none of them is able to directly set a port?
For my test I used `Contact::byConfig`.

paulfitz:
that looks about right, there's `NameConfig::setNamespace()` and `NameConfig::toFile()` that could do bits of that.

About `Contact` setters, yeah, that's kind of goofy. Feel free to add what you'd like to see there.

francesco-romano:
Do you mean something like this? 
```
    NameConfig nameConfig;
    nameConfig.setNamespace(nameServerContact.getName());
    nameConfig.setAddress(nameServerContact);
    return nameConfig.toFile();
```
?

I'm still having issues testing this, because the `Contact::byConfig` is not able to retrieve the `port_number` (`Value = 10000`, but `asInt()` returns 0) and randomly the socket creation fails.
I'll implement the `setPort()` method hoping to solve this issue.

francesco-romano:
@paulfitz  I pushed the changes in `Contact`. Not yet on the `Network` because I did not understand if they are correct (waiting for you).

About `Contact`: I just discover it's immutable.
So this snippet makes sense?
```cpp
Contact serverContact = Contact::byName("/fra");
    serverContact = serverContact.addHost([[self getIPAddress] UTF8String]);
    serverContact = serverContact.addPort(10000);
```

paulfitz:
@francesco-romano `Contact` was intended to be immutable, but it the resulting code has always felt clunky to me and I've regretted trying to go the immutable route with it. In recent work I went ahead and added setters (`setTimeout`, `setNested`). I think setHost/setPort would be fine, but addHost/addPort is also fine, whatever feels right to you.

Not sure I understand exactly what you are saying in https://github.com/robotology/yarp/pull/397#issuecomment-73212618.  Two reference points that could be useful are `cmdNamespace` and `cmdConf` in [Companion.cpp](https://github.com/robotology/yarp/blob/master/src/libYARP_OS/src/Companion.cpp) if you want to follow exactly what YARP currently does when setting namespace + server contact information. (This is all assuming you actually want to update config files. If you don't care about config files, there'd be easier ways)

francesco-romano:
About the `Contact`. I can also add a setter for host and port if you want. 

I'll take a look at the Companion better (I already did it a bit). 
I don't know about the configuration file. For my case I can either setup yarp every time or save the configuration file (but the user is not able to see it). What about a more general case? I think I should keep the same behaviour of the other functions.

I'll keep you updated :)

francesco-romano:
@paulfitz I've updated `Contact` to have both `addPort` and `setPort` and `setHost`.

I've also updated the `NetworkBase` method.
If I understood correctly the use case:
```cpp
NetworkBase::setNameServerName("/test"); //this is equivalent to $yarp namespace /test. It also creates a new config file.
Contact serverContact = Contact::byName("/test");
    serverContact.setHost("192.168.1.35");
    serverContact.setPort(10000);
    NetworkBase::setNameServerContact(serverContact);
//The above lines instead retrieve the configuration for the namespace, update the config and save back to the file.
```

Is this use case (and the implemented logic) correct?

paulfitz:
@francesco-romano looks good to me. I think it would also be reasonable for `setNameServerContact` to call `setNameServerName` if the contact it is passed contains a port name.

francesco-romano:
@paulfitz 
Do you want me to add these two lines to the method?
```cpp
if (nameServerContact.getName() != "")
        nameConfig.setNamespace(nameServerContact.getName());
```

paulfitz:
I think it makes sense. If it makes sense to you too, then please do add. Thanks!

francesco-romano:
@paulfitz done

PR #404, opened 2015-02-12
More Improvements to the Plugin System

drdanz:
* Simplify the implementation of YARP_PREPARE_PLUGIN using CMakeParseArguments
* Fix dll export for plugins when YARP_FORCE_DYNAMIC_PLUGINS is ON
* Generate dynamic plugins as MODULE instead of SHARED. Since MODULES cannot be linked by other CMake targets, they should not be known by the "master" plugin library and by the executable generated by YARP_ADD_PLUGIN_LIBRARY_EXECUTABLE, therefore they can be used only as plugins.


barbalberto:
Do you mean that the executable generated by  YARP_ADD_PLUGIN_LIBRARY_EXECUTABLE (like yarpdev or icubmoddev) will not be linked against the plugins but instead it will use the plugin feature to find the modules?
I found very handy in the past to have this binary that will know for sure where the libraries are (because it is linked to) as a quick check in case something does not work properly. 
This helped me a lot to understand is the problem of not loading a library is due to a wrong config file or environment setting for the plugin feature, something in the library code or the device was simply not compiled at all and so  on... 

So, is there a specific reason for this? Otherwise I'll keep the linking, at least for debug purposes.

If I didn't get it right, please forget the comment :)

drdanz:
@barbalberto thanks for the comment, that was also one of my issues, and the main reason why I opened the pull request ;)

I believe that you can still use the static version for that kind of debug... When building statically, all the plugins are linked as it used to be. Eventually we could change the ``YARP_FORCE_DYNAMIC_PLUGINS`` into a flag that can be set to ``STATIC``, ``SHARED``, or ``MODULE``. The first 2 options would be linked to the library and the 3rd will not. Or in addition to ``YARP_FORCE_DYNAMIC_PLUGINS`` (that already appears only when ``BUILD_SHARED_LIBS`` is disabled) we could have ``YARP_FORCE_STATIC_PLUGINS`` that could appear when ``BUILD_SHARED_LIBS`` is enabled. @barbalberto, @paulfitz what do you think? 

The main reason for wanting this is being able to remove plugin libraries after building and still being able to run the executable.
If the master library is linked to the plugin, then any executable linking to the master library will crash if the plugin is not available, and this makes it a *plug-in-but-not-out*, besides packagers life a :fire: hell :fire:. :smiling_imp:
At the moment in YARP we are doing:

```cmake
set (YARP_LINK_PLUGINS TRUE)
if (YARP_FORCE_DYNAMIC_PLUGINS OR BUILD_SHARED_LIBS)
  set (YARP_LINK_PLUGINS FALSE)
endif()
```

And when ``YARP_LINK_PLUGINS`` is ``FALSE``, ``libYARP_init`` does not link against the master plugin libraries, and since these are not linked anywhere it works, but this is a workaround.
For iCub everything is linking to the icubmod library and therefore it will not work when you remove one.

Another minor reason for wanting this is that at the moment we are installing the plugins with

```cmake
        install(TARGETS ${LIBNAME}
                EXPORT YARP
                COMPONENT runtime
                RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
                LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
                ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
```

The ``EXPORT YARP`` is unwanted, since these will end in the generated ``YARPConfigTargets.cmake`` and CMake will complain if they are not found, but unfortunately we cannot remove it, because the master library is linked to it and CMake will complain if they are not in the same export, even though on YARP we could probably remove the plugin master library as well, since the ``YARP_LINK_PLUGINS`` should do the trick (but still it won't work out of the box for other packages using YARP)...

drdanz:
@paulfitz @barbalberto @lornat75 Any issue with this? Otherwise I'll just go on and merge, since I have more stuff coming...

lornat75:
fine by me

paulfitz:
Ok with me if the tests related to finding plugins pass for static+shared build.

barbalberto:
Ok, I got your point and yes I think I can use static for debugging.

PR #405, opened 2015-02-12
Do not omit namespace when preparing carriers and devices

drdanz:
At the moment this works only because we add "using namespace"s in generated files but I believe that we should not endorse this. Using the full name of the class is more correct and makes it easier to understand what we are passing to the macro.

Also in order to be sure that the class name is always full, one day I would like to remove the "using namespace" from the generated files. Even though this will break some compatibility, it will make it more robust and less error prone.


drdanz:
@paulfitz opinions on this?

paulfitz:
:+1:

PR #407, opened 2015-02-13
Add flag rgbON to support Asus Xtion Pro

David-Estevez:
OpenNI2DeviceServer crashed with Asus Xtion Pro because it couldn't
find the RGB camera (this sensor does not have any), so a flag to
disble the RGB stream was added.

### DIFF
--- a/src/modules/openni2/OpenNI2DeviceDriverServer.cpp
+++ b/src/modules/openni2/OpenNI2DeviceDriverServer.cpp
@@ -18,49 +18,56 @@ yarp::dev::OpenNI2DeviceDriverServer::~OpenNI2DeviceDriverServer(void)
 {
 }
 
-void yarp::dev::OpenNI2DeviceDriverServer::openPorts(string portPrefix, bool userTracking, bool camerasON){
-    
+void yarp::dev::OpenNI2DeviceDriverServer::openPorts(string portPrefix, bool userTracking, bool camerasON, bool rgbOn){
+
     string receivingPortName = portPrefix+":i";
     receivingPort = new BufferedPort<Bottle>();
     receivingPort->open(receivingPortName.c_str());
-    
+
     if(userTracking){
         string skeletonPortName = portPrefix+PORTNAME_SKELETON+":o";
         skeletonPort = new BufferedPort<Bottle>();
         skeletonPort->open(skeletonPortName.c_str());
     }
-    
+
     if(camerasON){
         depthFramePort = new BufferedPort<yarp::sig::ImageOf<yarp::sig::PixelMono16> >();
         string strTemp = portPrefix+PORTNAME_DEPTHFRAME+":o";
         depthFramePort->open(strTemp.c_str());
+    }
+
+    if (camerasON && rgbOn){
         imageFramePort = new BufferedPort<yarp::sig::ImageOf<yarp::sig::PixelRgb> >();
-        strTemp = portPrefix+PORTNAME_IMAGEFRAME+":o";
+        string strTemp = portPrefix+PORTNAME_IMAGEFRAME+":o";
         imageFramePort->open(strTemp.c_str());
     }
+
 }
 
 void yarp::dev::OpenNI2DeviceDriverServer::sendSensorData(){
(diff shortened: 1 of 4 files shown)


paulfitz:
thanks @David-Estevez! @kt10aan are you available to review this?

kt10aan:
thanks @David-Estevez :)

Regarding your PR, there is a really major problem:
You deleted the printVideoModes functionality and all the code associated with it! Instead of properly taking all the cases (e.g, if RGB camera is off, and do probe but only for Depth video modes). 

There is also a really minor problem that you change the words in the output of the driver. This in itself is problematic but it is also done in a rather inconsistent way (e.g.,you replace "create" with "start", but only once). 

Having said that, the following are more important:

The camerasON flag was supposed to roughly do what the rgbOn flag you added is doing, and thus, allowing for use with the Asus sensors that lack RGB cameras, and also for those users that simply didn't want the RGB stream. 

As it is a remnant of the old driver, and as I did not have an Asus without RGB cam, it remained unfinished (and no one pushed me hard enough :D)

I think a better approach is to merge the noCameras and the rgbON switches (as the depthStream is needed for userTracking anyway) into one, something like --noRGB (or whatever the naming specialist, Giovanni, will recommend :D).

kt10aan:
@David-Estevez, at https://github.com/kt10aan/yarp I have committed a similar --noRGB option, that  takes into account all the other options, replaces the old noCameras option and it reduces redundant code.

Could you be so kind to test it, since you have the RGB-less ASUS?
Especially try to:
- Record an .oni file
- Play an .oni file
- Even better, if you have, play an .oni file that has RGB info in it
- Use the --syncFrames and the --imageRegistration option with the --noRGB

Please note that you need to change the YARP_DATA_DIRS to point to the kt10aan/yarp, otherwise you will still be using the robotology/yarp.

gsaponaro:
Hi @David-Estevez! It's nice to have more people using this piece of software. @kt10aan: next week I can test the new option on our Asus w/RGB.

Regarding naming conventions, I agree with @kt10aan that ```--noRGB``` should disable RGB input (thus RGB output too). And yes, the option that was called ```--noCameras``` was poorly named, making it unclear that it referred to output only. In the end, the most important thing is that running the driver with no options probes for the available cameras and starts the program smoothly.

gsaponaro:
```--noRGB``` option works fine (including in combination with other options) with my Asus Xtion Pro with RGB camera.

**Comment by David-Estevez** [comment]
Hi everyone!

Thanks for your comments. It is true that I accidentally erased the printVideoModes functionality, I'll be more careful the next time.

I have tested the code provided by @kt10aan with my device in all the required cases (Except playing a .oni file with RGB info, because I don't have any currently) and it works without any problem.

Thank you all for your help!

**Comment by kt10aan** [comment]
@David-Estevez, thanks for the testing!.

I also made another commit today at the kt10aan/yarp repository. The goal is for the driver to not crash if it is used with an RGBless Xtion, even if the --noRGB option is not used.

Since I do not have an RGBless Xtion, could you please test it? Just start the driver with:
```
yarpdev --device OpenNI2DeviceServer 
```

No need for  ``--noRGB``.

Remember though to point your YARP_DATA_DIRS to kt10aan/yarp etc etc...

gsaponaro:
Here is an .oni file with RGB and depth, if you need it for testing:

http://we.tl/Z65XIsCgA8 (215MB, link will expire on 24 February)

David-Estevez:
@gsaponaro  I have tried your oni file and it works perfectly

@kt10aan It also works without using the --noRGB flag, and notifies the user the RGB stream was not found

kt10aan:
@David-Estevez, many thanks!!
So, to summarize, with the updates I did:
- Sensors without RGB cameras are properly initialised and any RGB-related functionality is disabled gracefully
- Sensors with RGB cameras have the option to use ``--noRGB`` and disable RGB-related functionality

I will tidy up my commits (rebase/squash/etc) and will create a PR. @paulfitz, I think you can close this PR

paulfitz:
thanks all :-)

David-Estevez:
Ok!! Thank you all!!

kt10aan:
Created the PR #420

PR #413, opened 2015-02-18
[test] run integration tests via ctest

paulfitz:
This runs the various integration tests on travis using the regular ctest machinery, resulting in some simplification.  For now it seems necessary to run those tests only under clang, since some swig targets are now too bulky for gcc with the memory travis provides. Example run: https://travis-ci.org/robotology/yarp/builds/51133895

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ matrix:
   exclude:
     - compiler: gcc
       env: TRAVIS_BUILD_TYPE=Profile TRAVIS_CMAKE_GENERATOR="Unix Makefiles" TRAVIS_WITH_ACE=false TRAVIS_WITH_INTEGRATION_TESTS=true TRAVIS_BLESSED=false
-    - compiler: gcc
+    - compiler: clang
       env: TRAVIS_BUILD_TYPE=Profile TRAVIS_CMAKE_GENERATOR="Unix Makefiles" TRAVIS_WITH_ACE=true TRAVIS_WITH_INTEGRATION_TESTS=true TRAVIS_BLESSED=true
     - compiler: clang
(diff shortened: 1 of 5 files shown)


drdanz:
Awesome, If I understand correctly it should fix also the annoying issue with coveralls and multiple builds and perform the coverage analysis only for one build...

I don't remember why I changed it to use ``if [ "k$TRAVIS_WITH_XXX" = "ktrue" ]`` but ``if $TRAVIS_WITH_XXX`` should work, perhaps I was confusing it with ``if [ $TRAVIS_WITH_INTEGRATION_TESTS ]``.

Anyway :+1:, will merge as soon as the builds are successful.

paulfitz:
Yes, I did try to make it report to coveralls for a single "blessed" build to avoid the random-percentage problem.

drdanz:
@paulfitz It looks like that coveralls is not working, any idea why?

drdanz:
Actually it looks like that the travis build exits before finishing...

paulfitz:
oh this could be a gcc vs clang thing...

drdanz:
@paulfitz also there is a bug in the matrix... :1234:
From my limited knowledge of gcov, it does not work with clang, therefore the "blessed" version should be gcc (that was the reason for the ``if [ "$CXX" = "g++" ]``) but it looks like that gcc is skipping the blessed, while clang is not...

paulfitz:
Hmm then this needs some work, gcc build doesn't appear to have enough memory on travis anymore to build the bindings (python at least).

paulfitz:
@drdanz in the absence of a better idea, I switched the main integration test job to use gcc (for coverage), while forcing the swig tests to use clang to compile the bindings. This is the only way I could squeeze them into whatever amount of memory travis is now giving.

drdanz:
@paulfitz Unfortunately it looks like that we hit the line limit for travis log :sad:, but it looks like that the build was successful, the raw log looks ok. I don't know if building the bindings using clang could be a problem, but it looks ok, so I'll just go on and merge...

PR #420, opened 2015-02-23
OpenNI2 device detects and handles RGB-less sensors

kt10aan:
With this PR:

- Sensors with no RGB cameras are correctly initialised and RGB functionality is disabled
- The ``--noCameras`` option has been replaced with ``--noRGB``. This disables all RGB-related functionality
- Removed a warning during compilation
- Some general cleaning of white lines etc.

Commits squashed and rebased :)
Documentation will update shortly.

Many thanks to @David-Estevez for his work and testing!


paulfitz:
thanks @kt10aan, @David-Estevez

PR #423, opened 2015-02-25
More plugin changes

drdanz:
See commit messages


paulfitz:
A lot of great changes here.  Travis is showing one possible issue in the `integration::runtime` test:
```
82: ###############################################################
82: ## Check shared fakebot is findable
82: device fakebot
82:   fakebot (type device) (name fakebot) (library yarp_fakebot) (part fakebot)
82:   found by linking to /home/travis/build/robotology/yarp/build/runtime_tests/fakebot/lib/yarp/yarp_fakebot.so
82:   found by linking to /home/travis/build/robotology/yarp/build/runtime_tests/fakebot/lib/yarp/yarp_human.so
82: ###############################################################
82: ## Check shared fakebot is startable
82: yarpdev: ***ERROR*** could not find device <fakebot>
82: yarpdev: ***ERROR*** device not available.
82: Suggestions:
82: + Do "yarpdev --list" to see list of supported devices.
82: + Or append "--verbose" option to get more information.
82: Failure: Could not start fakebot
82/84 Test #82: integration::runtime ................***Failed  608.07 sec
```

I'll look more carefully at this tomorrow.

barbalberto:
A small issue with portaudio was reported this morning, using yarpdev command (with plugin system)
 yarpdev --device portaudio --id 1 --c 2 --read --rate 48000
 from tutorial page http://wiki.icub.org/wiki/Grabbing_sound_with_PortAudio

the device get stuck because:
- constructor gets called
- destructor is called
    - destructor internally calls the close, but no open was called before so device structures were not initilized yet and it hangs up.

Removing the close() call from the destructor it correctly opens following the patter:
- constructor
- destructor
- constructor (2nd time)
- open
..... use it ...   (actually it does not work on the icub, but this is a separeted issue)
- close
- destructor (2nd time)

(I have not pushed any changes yet)

Using this pattern we need to take a bit more care when writing devices to verify that both the previous flows of execution and the following one are safe.

- constructor
- open
... use it ...
- destructor (without the close)

In this case the destructor should call the close() function, so I think the destructors should be something like:
~device()
{ 
    if(device_was_opened)   --> set if open was succesful
         close();
}

I think that most of the devices we have were not wrote with this careful check in mind, so we need a quick check to verify if they are compliant.
What do you guys think about it ?

paulfitz:
The `close()` method on a `DeviceDriver` should complete even if `open()` has not been previously called, and any instances where that does not happen should (in my opinion) be treated as bugs and fixed.

lornat75:
I agree.

drdanz:
> The `close()` method on a `DeviceDriver` should complete even if `open()` has not been previously called, and any instances where that does not happen should (in my opinion) be treated as bugs and fixed.

+1

barbalberto:
Yes, I agree as well. 
So we need to check that all devices do behave that way, and add this policy in the wiki page http://wiki.icub.org/yarpdoc/dev-how-to.html so that at least it is stated somewhere.

drdanz:
My personal opinion is that we don't have to explain *how to write a C++ class that allocates memory* in our documentation, there are lots of tutorials for that... :smiling_imp:

drdanz:
@paulfitz: There was a dumb error in "Fix plugin search paths on windows" (a wrong ``a = x ? dll_name : name`` that should have been ``a = x ? name : dll_name``  that worked in many cases because ``dll_name`` and ``name`` were the same), integration tests should be fixed now.

I also rebased the branch over the latest master.

drdanz:
Hmm, looks like I broke something else...

drdanz:
Looks like it's not my fault but that this commit 683bf09e54eb5ab8c81b4f403c454b88f78d364a is causing the segmentation fault

drdanz:
Thanks @barbalberto for fixing the issue with ``dev::PolyDriverTest``
I rebased the changes over the latest master, hopefully travis and appveyor will be happy now...

paulfitz:
Good, thanks @drdanz, @barbalberto!

Issue #113, opened 2014-01-22
Broken IAnalog interfaces


elen4:
Commit 394f752585f845ed7562df5483f99c5a8ee2631c changed the client implementation of the yarp::dev::IAnalog interface, which means that ALL server implementations (i.e., many of iCub drivers) have to be changed accordingly. @EnricoMingo is there a VERY good reason for this change?

drdanz:
@barbalberto Was there any progress on this?

barbalberto:
The code with the refactor of the portname with the new policy is on a branch to be tested.

lornat75:
is anybody going to test it? Can you write down a wiki page with a summary of the policy?

barbalberto:
I'm testing it

barbalberto:
I did a small change to improve back compatibility with the skinManager.
The branch is currently on the red iCub to test it with old code that does not use the analogClient.

The new policy will look for parameter 'name' which will set the full port name (as client already does); if not found will fall back for the old parameters style 'robotName' + 'part' and will write the following warning message:

[WARNING] AnalogServer device:
************************************************************************************************
  AnalogServer is using deprecated parameters for port name! It should be:
       name:         full name of the port, like /robotName/deviceId/sensorType:o
       period:       refresh period of the broadcasted values in ms (optional, default 20ms)
************************************************************************************************ 

The  user code will work as before, just with the warning message.
If the user changes the config file to use the new parameter, he/she will get rid of the message. Note that he/she can still use the same old ugly portnames as before (just by typing it 'as is' in the config file) and the client (when used) will still be able to reach the server.
With the new policy the user can have a better port name as he/she wishes.

Shall I document only the new policy or also the deprecated one?

lornat75:
The new policy is enough.

barbalberto:
I think this issue can be closed due to new policy

randaz81:
I suggest to close this issue after adding the `name` parameter to all the robots in the repos. We are currently having a lot of warning in the log because we are still using the old deprecated mode.
Davide Tome can help running on the repo the same script used for the controlBoardWrappers.

barbalberto:
@randaz81 : can we close this issue now?

randaz81:
@davidetome completed the task succesfully

PR #429, opened 2015-03-04
Analog wrapper port name

barbalberto:
Moved analogSensorClient into folder and separated into header and cpp
Implemented new policy for portname using --name option instead of robot+part following issue #113 
It was tested on the red iCub - I did also connection test with custom config file with both new policy / both old policy and mixed configuration and it works, provide the config files are correct in the mixed config.


pattacini:
What does it mean in terms of current icub port names?
Any example?

barbalberto:
Short answer is: nothing immediatly because old behaviour is preserved, the update to the new policy requires a small change in config file.

With old policy, the port name was build as follow:
streaming port name:  param robot + param part + "analog:o" (last part enforced in code)
rpc port name :            param robot + param part + "analog:o/rpc:i" (last part enforced in code)

new policy is:
streaming port name:  param name  (used as is)
rpc port name :            param name + "/rpc:i"  (last part enforced in code)

If you use ' param robot + param part + "analog:o"  ' as name for the new policy, again it is back compatible, but then you can get rid of the ugly  'analog:o/rpc:i' since this part is now under user control.
You can also make the portname more clear by adding for example the type of data exchanged, like:

name   /icub/left_arm/forceTorque

will open
/icub/left_arm/forceTorque
/icub/left_arm/forceTorque/rpc:i

Of course, in this case you have to change the client accordingly. The clients already used this policy so now they are symmetric.
A warning message will be printed if old policy is used..

pattacini:
many thanks @barbalberto for the resume :)
if in the future we are going to change those port names via the new params, please remember to let the community know, since sometimes there are direct connections to analog ports made in the code (because are somehow deemed to be always available).

barbalberto:
If there are no other comments I'll merge it

barbalberto:
closing this pull request without merging  'cause I've updated the code with more stuff, I'll open a new one later

PR #451, opened 2015-03-19
fixing warnings

lornat75:


src/yarpmotorgui-qt/partitem.cpp
```diff
@@ -537,7 +536,7 @@ void PartItem::onSendPid(int jointIndex,Pid newPid)
 }
 void PartItem::onCalibClicked(JointItem *joint)
 {
-    const int jointIndex = joint->getJointIndex();
+    joint->getJointIndex();
```
drdanz: I don't know what it does, but this call looks useless now...

src/yarpmotorgui-qt/mainwindow.cpp
```diff
@@ -103,7 +103,8 @@ MainWindow::MainWindow(QWidget *parent) :
     partToolBar->setAllowedAreas(Qt::AllToolBarAreas);
 
     QFontMetrics metrics(f);
-    int w = metrics.width(globalLabel);
+    // int w = metrics.width(globalLabel);
+    metrics.width(globalLabel);
```
drdanz: Also this call looks useless now...

lornat75: it seems so but it may still have side-effects, I wanted to have minimal impact...

src/yarpmotorgui-qt/partitem.cpp
lornat75: it seems so but it may still have side-effects, I wanted to have minimal impact...

drdanz: But the point for the warnings is that they may hide an issue, I'm against just hiding them without fully understand what's happening. Sometimes who wrote the code changed his mind or left it after deleting some code, but some other time these are real bugs, for example he might have forgotten to do something with the variable, or even worse a wrong variable might be used instead of the right one.

In my opinion just hiding this kind of warnings is wrong, one should understand what the code does first, and then hide it only if there is a good reason to do it...

In this case, I think that we should ask the question "why `jointIndex` isn't used?" For example, maybe the number of the joint should be in the MessageBox... if this is not the case then what does `getJointIndex()` do? According to the code is just a `return jointIndex;` therefore it does nothing useful if you don't assign it to anything.
On the other hand `joint` will be unused here if this line is removed.

So, if I was to fix this warning, I'd either use `jointIndex` in the message box or remove the whole line and mark the `joint` variable as `YARP_UNUSED(joint)` or `Q_UNUSED(joint)`, since it is Qt stuff...

lornat75: I am fine with Q_UNUSED(joint) if this makes you sleep better.

The author of that code should also have a look, remove the function call and test it. If the author does not have time we can live with this.

lornat75:
Closing this pull request because already addressed via e23040e37df411f2d1f01101f69438cb96842394

PR #461, opened 2015-03-28
[ros] omit final null character when sending strings, fixes #452

### LINKED ISSUE #452: [YARP to ROS] Sending string to a ROS topic adds a "\x00" to the end of the string
I have code on Github that reproduces this behavior but maybe you can test it with some code of yours.

I'm adding the string to the bottle like this:

    bottleToSend.add("the string");

To run it I'm doing the following:

    // terminal 1
    $ roscore
    // terminal 2
    $ yarpserver --ros
    // terminal 3
    $ yarpidl_rosmsg --web true --name /typ@/yarpidl
    // terminal 4
    $ roslaunch my_package my_launch
    // terminal 5
    $ ./my_yarp_code
    // terminal 6
    $ rostopic echo /the_topic
        field: "the string\x00"

I'll gladly provide the instructions to run my code in order to test this. Actually the repositories containing  it have almost all the instructions needed.

Cheers!

paulfitz:
For compatibility with an ancient version of yarp, strings have been up to now transmitted with a null character included. More recent versions of yarp will accept either a string with or without a null termination. This commit stops transmitting that null termination, but preserves the ability to receive a null terminated string for compatibility between programs using different versions of yarp.

The motivation for this change is for ROS compatibility, see #452, but it will also be a step towards removing an awkward corner case where yarp will not faithfully transmit a non-null-terminated string that happens to have a null character embedded as its last character [as opposed to a null-terminated string :-)].

### DIFF
--- a/src/idls/rosmsg/src/RosTypeCodeGenYarp.cpp
+++ b/src/idls/rosmsg/src/RosTypeCodeGenYarp.cpp
@@ -393,7 +393,6 @@ bool RosTypeCodeGenYarp::writeField(bool bare, const RosField& field) {
     fprintf(out,"    // *** %s ***\n", field.rosName.c_str());
     if (field.rosType=="string") {
         // strings are special; variable length primitive
-        string tweak = bare?"":"+1";
         if (field.isArray) {
             if (!bare) {
                 fprintf(out,"    connection.appendInt(BOTTLE_TAG_LIST|BOTTLE_TAG_STRING);\n");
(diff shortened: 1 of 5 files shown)


mikearagao:
Hey @paulfitz, any idea of when will this PR be merged? I'll try to test the  branch now!

lornat75:
If I understand correctly if I send a string with an extra \0 this will get removed, right?

drdanz:
:+1: Looks good to me.

paulfitz:
@lornat75 yes, if you for example tried to transmit a string of length 3 whose third character happened to be `\0`, that last character would get removed, and you'd receive a string of length 2. This is not great behavior, but is something yarp already does, rather than a change in behavior. What this PR does is stop yarp transmitting an additional `\0` null termination after every string, which has not been necessary for some time.

lornat75:
Fine by me, it seems a step forward -- ideally at some point we should remove it completely.

Issue #462, opened 2015-03-29
Update thrift generation to make use of yError()


pattacini:
What about changing the generation of thrift files by relying on the new `yInfo()`, `yError()` ... defines/routines?

see https://github.com/robotology/icub-main/commit/e9a4053bbaf1d5c3d786c81c5ca6ffd45eeebfad

While updating, please take care of removing the ending char `'\n'`

/cc @randaz81 @paulfitz

PR #465, opened 2015-03-31
thrift generated files will use yError() in place of fprintf()

pattacini:
Thrift auto-generated files will employ `yError()` and no longer `fprintf()` in order to stream the messages throughout the network and make them be displayed by means of `yarplogger`.

see #462


pattacini:
Once merged, we have possibly to regenerate thrift files within yarp.

lornat75:
Considerd it is in the yarp specific generator it seems acceptable to me.

PR #482, opened 2015-04-22
Fix: RF module calls close if configure fails

barbalberto:
I found that robotInterface was crashing if the configure function fails. It turns out that I had to explixitly call the close function in order to clean up things.
 So me and @drdanz thought that maybe it'll be easier for user if the RFModule will call the close function by his own if the configure fails, like in the patch proposed.

@paulfitz @lornat75 what di you think about it? Do you see any drawback?

Maybe also the stopModule could call the close?

### DIFF
--- a/src/libYARP_OS/src/RFModule.cpp
+++ b/src/libYARP_OS/src/RFModule.cpp
@@ -365,6 +365,7 @@ int RFModule::runModule() {
 int RFModule::runModule(yarp::os::ResourceFinder &rf) {
     if (!configure(rf)) {
         ACE_OS::printf("Module failed to open\n");
+        close();
         return false;
     }
     return runModule();


pattacini:
Lots of user code works without this assumption.
I'm afraid this change would have a quite big impact.
I would thus change `robotInterface` in the meantime to then have time to better asses this PR.

lornat75:
I agree with @pattacini we should be careful. On the other hand this would affect only cases in which there is a failure opening a RFModule -- which should be limited.

Actually, looking at the code I think there could be problems if close() is not called in all cases in which configure() fails; configure is called implicitly within runModule(ResourceFinder &) or explicitly by the user. The first case is covered by this PR; the second should be handled by calling close() in the destructor.

paulfitz:
I think this change makes sense. There could be custom `close` methods out there that will be surprised by this, and crash, but for the reason @lornat75 notes the consequences should be limited.

pattacini:
We would need to update tutorials and documentation to clearly point out this change.

pattacini:
Btw, my opinion is that crash is always a crash, even if it has limited effect and causing crashes and bother to users for such a minimal improvement should be avoided.

lornat75:
I would propose to also call close in the destructor of RFModule.

The documentation should then specify that close() should handle and cleanup all the initialization that is made in configure().

drdanz:
So, if I understand correctly, some module assume that if configure fails, then the close function is never called, but if the module starts at some point the close will be called, am I right? What is the reason for this? Does the close method do nasty stuff?
@pattacini can you point me to one of the modules that assume this?

Perhaps instead of calling close, we could have another virtual method empty by default that is executed here, and that we can override explicitly calling close?

Or we could simply add a parameter like
```diff
-yarp::os::RFModule::runModule(yarp::os::ResourceFinder &rf);
+yarp::os::RFModule::runModule(yarp::os::ResourceFinder &rf, bool closeOnFail = false);
```
so that the default logic is unchanged.

@lornat75 I wouldn't recommend calling close in the destructor, I believe that the close should be called by the `runModule` before exiting (as it is now). If we want to change it, at least, we should ensure that it is called only once, since calling close twice may lead to unexpected behaviours. For example the robotInterface would enter the interrupt1 and then the shutdown phases twice.

lornat75:
At the moment the assumption is that configure() does not let stuff to be uninitialized if it fails.

In my understanding this PR changes this behaviors and let close() perform this uninitialization. However RFModule can be used also like this:

RFModule module;

if (module.configure(rf))
   module.runModule();

This would produce a behavior that is significantly different from:

module.runModule(rf);

because the latter would call close on failure.

To make the behavior consistent I propose we add close in the destructor.

Otherwise this PR would introduce an asymmetric behavior.

Another possibility could be to leave things as they are and just fix cases in which configure() fails leaving stuff uninitialized.

lornat75:
I realize I may be missing the reason why you submitted this PR -- maybe it is better if we talk about this.

barbalberto:
From what I see it seems that if we use the module like:

    RFModule module;
    module.runModule(rf);

then if the configure fails, the close is not called, while I'd expect it to be at a certain point.
I agree with @lornat75 that this introduce an asymmetric behaviour in the case he pointed out, and maybe that's not the better place to put the close() calls. That's why we opened the discussion.

My assumption is that whichever way the user choose to configure/run the module, at the end of the day the close should be called at least once (so having it in the destructor will assure this), but I know this assumption is arbitrary :-)

drdanz:
I think that the documentation here is not very clear about how a module should be stopped when the configure fails. I'm adding in the next comment all the relative docs.

The `close()` documentation clearly states
> Override this to perform memory cleanup or other activities.

so that's where I put all cleanup.

The configure phase calls several actions (open devices, attach and calibrate) in the order specified by the user, therefore if some action succeeds, and some later action fails, we must call the actions in the interrupt1 (park, etc) and then the shutdown (detach and close devices).

We started discussing about this because I did this change in the robotInterface, when the configure fails:

```diff
 if(/* startup phase error */) {
-    // A very long comment about the reason why we are calling interruptModule()
-    // and then close() here
-    interruptModule();
-    close();
+    stopModule();
 }
```

assuming that `stopModule()` would cause `close()` to be called and not just `interruptModule()`, but I was wrong. By looking at the code, I didn't notice that the `runModule()` called was `RFModule::runModule(yarp::os::ResourceFinder &rf)` and not `RFModule::runModule()`.

The second method implementation is something like this:
```c++
while (!isStopping() /* changed by stopModule */ ) {/* do stuff */}
close();
```

And I didn't notice that the `close()` is never reached, so I assumed that just calling `stopModule()` would reach it.

The easy (and probably right) thing to do would have been just adding a `close()` after the `stopModule()` call, but it looked weird, not clear and not well documented, so we started this discussion...

drdanz:
```c++
    /**
     * Calls updateModule() until that returns false.
     *
     * Make sure you first configure your module by calling the configure()
     * function. updateModule() is then called every getPeriod()
     * seconds.  Be aware that the the respond() command could be
     * asycnhronously at any time, if there is input from the
     * standard input or a port connected via attach().
     *
     * @return 0 on success
     *
     * \note attachTerminal() is no longer called automatically.
     */
    virtual int runModule();

    /**
     * Simple helper method to call configure() and then runModule().
     *
     * @param rf a previously initialized ResourceFinder
     * @return 0 upon success, non-zero upon failure
     */
    virtual int runModule(yarp::os::ResourceFinder &rf);

    /**
     * Configure the module, pass a ResourceFinder object to the module.
     *
     * @param rf a previously initialized ResourceFinder
     * @return true/false upon success/failure
     *
     * \note attachTerminal() is no longer called automatically. You
     * can call it in the configure function.
     */
    virtual bool configure(yarp::os::ResourceFinder &rf);

    /**
     * Close function.
     *
     * This is called automatically when the module closes, after the last call
     * to updateModule.
     * Override this to perform memory cleanup or other activities.
     *
     * @return true/false on success failure.
     */
    virtual bool close();

    /**
     * Try to halt any ongoing operations by threads managed by the module.
     *
     * This is called anynchronously just after a quit command is received.
     * By default it does nothing - you may want to override this.
     * If you have created any ports, and have any threads that are
     * might be blocked on reading data from those ports, this is a
     * good place to add calls to BufferedPort::interrupt() or
     * Port::interrupt().
     *
     * @return true if there was no catastrophic failure
     */
    virtual bool interruptModule();

    /**
     * Ask the module to stop.
     *
     * @param wait specifies if stop should block and wait termination. This is
     * not implemented yet.
     */
    void stopModule(bool wait = false);
```

pattacini:
@drdanz my main concern is related to situations as the one sketched below, which are likely spread around in the user code:

```cpp
bool configure() {
    Class *class=new Class();
    if (!isOk) {
        delete class;
        return false;
    }
}

bool close() {
    delete class;
    return true;
}
```

With the proposed change we'll have a seg-fault if `isOk==false` (it's a good rule to put `class=NULL;` after deletion, but who knows).

Other times, the use might rely on a `terminate()` helper which in turn gets called both in `configure()` and `close()` for object disposal, as for example done in [depth2kin](https://github.com/robotology/icub-main/blob/master/src/tools/depth2kin/src/module.cpp#L967).

The handling through the `closeOnConfigureFailure` option seems a quite nice solution.

drdanz:
After some discussion with @lornat75 I think that the best thing to do is to improve the documentation to specify what is called and where, and that when `configure()` fails, the `close()` method is not called. Then we will fix the code in robotInterface...
@barbalberto what do you think?

barbalberto:
So revert your changes to my last fix :-P
Ok, fine with me.

pattacini:
Closing and not merged as agreed

PR #483, opened 2015-04-23
IRemote calibrator

barbalberto:
This PR introduce a new interface to remotize the calibration procedure from user applications like yarpMotorGui. Both yarp tweak number and yarp_protocol_version were increased.
It introduces agnostic calls and yarp messages to a remote iCalibrator device, like:

    virtual bool isCalibratorDevicePresent(bool *isCalib);
    virtual bool calibrateSingleJoint(int j)=0;
    virtual bool calibrateWholePart()=0;
    virtual bool homingSingleJoint(int j)=0;
    virtual bool homingWholePart()=0;
    virtual bool parkSingleJoint(int j, bool _wait=true)=0;
    virtual bool parkWholePart()=0;
    virtual bool quitCalibrate()=0;
    virtual bool quitPark()=0;

The main difference between this interface and the remotized iControlCalibration is that in this case we are calling a device that knows how to do the calibration procedure (the xxxCalibrator device inside the robotInterface), so the user does not need to provide the calibration data anymore, avoiding to spread out of delicate parameter like calibration ones to user applications.

This interface has been implemented in remoteControlBoard and controlBoardwrapper2 devices and used in yarpmotorgui (both qt and gtk version)

In order for the wrapper to know which calibrator device has to use, its attach action has been extended to accept a special target with the 'Calibrator' keyword. This means that the wrapper config file has to add one line like in the example hereafter:

    < action phase="startup" level="5" type="attach">
      < paramlist name="networks">
          < elem name="FirstSetOfJoints">  left_upper_arm_mc < /elem>	
          < elem name="SecondSetOfJoints"> left_lower_arm_mc < /elem>
          < elem name="Calibrator"> left_arm_calibrator < /elem>
     < /paramlist>
    < /action>

Verify if it works in the same way with iCubInterface

The logic implemented in the yarpmotorguis, after a chat with @randaz81 is the following:

Calibration: Always use the remoteCalibrator (i.e. robootInterface), do not check for values in the .ini file. Should errors occourr, it'll check if the remoteCalibrator is missing (not instantiated or not configured) or if the calibration procedure failed and show appropriate error message.

No calibrator was found:
no calib device found

Error during calib procedure:
calilbration error
In case of calibration for whole part, the very same calibration procedure used at robotstart-up will be called. This means, for iCub robots, setSafePid + calib message + goToZero, keeping the correct joint order. For single joint calibration, only the calibration message will be sent.

Homing: if user defined values are found in the ini file, they'll be used. In this case the behaviour is as usual. It'll be quite, no pop up messages
homing old style

If no user defined values are found, then a request for homing will be sent to through the RemoteCalibrator interface to the calibrator. In this case an info message will be displayed to the user to notify a change in behaviour with respect to previous version.

remote homing

Info message have been set in this way with the idea of warning the user about a change, if we adopt the long-run perspective in which the remoteCalibrator is the default, maybe it make sense to swap the message policy so that using remoteCalibrator will always use valid values because they are took from robotInterface itself and it'll became the default way, so no reason to bother the user each time with a message, while if custom values are found in the ini file it is up to the user the responsability of theyr correctness and should be less frequent so maybe it make sense to have a warning message like "custom homing has been detected"

Before merging this PR it is better if also xml file for robots are updated, but is not such a big deal because with default config file the calibration does not work either because calib param are not provided while homing values are present, so the usability of the gui is the same, just shows a different error message :)
However I've to test it on the iCubInterface but comments are welcome in the meantime.


src/libYARP_dev/src/modules/ControlBoardWrapper/ControlBoardWrapper.cpp
```diff
@@ -722,12 +722,22 @@ bool ControlBoardWrapper::attachAll(const PolyDriverList &polylist)
...
-        // and attach
+        // look if we have to attach to a calibrator
+        std::string tmpKey=polylist[p]->key.c_str();
+        if(tmpKey == "Calibrator" || tmpKey == "calibrator")
+        {
+            // Set the IRemoteCalibrator interface, the wrapper must point to the calibrato rdevice
+            yarp::dev::IRemoteCalibrator *calibrator;
+            polylist[p]->poly->view(calibrator);
+
+            IRemoteCalibrator::setCalibratorDevice(calibrator);
```
lornat75: this branch is specific to "Calibrators"; if you change
IRemoteCalibrator::setCalibratorDevice(calibrator);

to IRemoteCalibrator::attach(PolyDriver &);

we could remove this special case?

src/libYARP_dev/src/RemoteCalibratorImpl.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * that knows the correct procedure and the correct parameters for the calibration,
+ * so that user doesn't have to worry about configuration parameters.
+ */
+
+IRemoteCalibrator::IRemoteCalibrator()
+{
+    _remoteCalibrator = NULL;
+}
+
+bool IRemoteCalibrator::setCalibratorDevice(yarp::dev::IRemoteCalibrator *dev)
```
lornat75: this can become a generic function
bool IRemoteCalibrator::attach(PolyDriver &p)
{
  yarp::dev::IRemoteCalibrator *calibrator;
+p->view(calibrator);

  if (!calibrator)
     return false;

  _remoteCalibrator=calibrator;
  return true;
}

randaz81:
can we resume the discussion on this pull request in order to merge asap? A lot of people are looking forward to this feature...

lornat75:
I had a personal discussion with @barbalberto, maybe he can summarize it here.

barbalberto:
I'm working on it testing some changes suggested by Lorenzo.
@lornat75 changing the function will not get rid of the if branch, because the 'attach' function has to be called on a different device, like:

```c++
if (!calibrator)
   subdevice[n].attach(target_device)

if(calibrator)
   wrapper.attach(target_device)
```

I don't see a way to have it working without an if branch, or some ugly custom workaround (but see below Alternative solution 3).

About the xml config file, after some tests, the configuration that can work are the following:  (I have access only at the information inside the <action> </action> piece of xml)

Current solution
```xml
      <paramlist name="networks">
             <elem name="dev1">         left_upper_arm_mc   </elem>
             <elem name="dev2">         left_lower_arm_mc   </elem>
             <elem name="calibrator">  left_arm_calibrator </param>
      </paramlist>
```
Not so good because it mixes motion control devices and calibrator under the keyword network, also 'dev1' and 'dev2' are custom user strings while 'calibrator' is a semantic keywork needed by the code in order to work. 

Alternative solution 1:
```xml
      <paramlist name="networks">
             <elem name="dev1">         left_upper_arm_mc   </elem>
             <elem name="dev2">         left_lower_arm_mc   </elem>
      </paramlist>

      <paramlist name="calibrator">  
             <elem name="pippo">  left_arm_calibrator </elem>
       </paramlist> 
```
This solution makes clear difference between networks and calibrators, elem name is not used by the code so it can be custom like 'dev1' and 'dev2'.
'calibrator' is a keyword exactly like 'network' is.
Since 'calibrator' is a list it may lead to think that more calibrators can be used for a single robot part, which is not true in the current implementation (does it make sense to have multiple calibrators for one part?)
This requires some small changes in the robotInterface and in the CBWrapper

Alternative solution 2:
```xml
      <paramlist name="networks">
             <elem name="dev1">         left_upper_arm_mc   </elem>
             <elem name="dev2">         left_lower_arm_mc   </elem>
      </paramlist>

      <param name="calibrator"> left_arm_calibrator </param>
```
Similar to the previous solution, but calibrator is a single value an not a list.

This last 2 solutions also allow to have the attach to the motionControl device and to the wrapper splitted in 2 different times, like:

```xml
      <paramlist name="networks">
         <elem name="a">  left_upper_arm_mc   </elem>
         <elem name="b">  left_lower_arm_mc   </elem>
      </paramlist>
  </action>   

    <action phase="startup" level="7" type="attach">
        <param name="calibrator">  left_arm_calibrator </param>
    </action>    
```
if some other action has to be performed in between.

Alternative solution 3:
A completely different solution is that this specific attach is automatically done when the action calibrate is called:

```xml
<action phase="startup" level="10" type="calibrate">
	<param name="target">left_arm_mc_wrapper</param>
</action>
```

Pro: 
- very small change only in the robotInterface when this action is called
- no need to change xml
- no 'if branch' in the CBWrapper.
Cons: it is a bit of black magic hidden from the user, no way of customize, if action calibrate is removed from the calibrator xml, then no calibration will be possible later (however if user wants to skip calibration a dedicated option is available, asaif to remove calib action was never needed).

randaz81:
Currently I can't see any practical advantage of solution 2 respect to 3, so my preference goes to solution 3. (Unless I'm missing something...)
About the cons you are mentioning: indeed removing the calib action was never needed, so this is not a drawback for me

lornat75:
I vote for 2 or 3. Thanks @barbalberto

barbalberto:
Ok, I''ll go for 3 and see implement it also in the iCubInterface

barbalberto:
After discussion with @randaz81 and @lornat75  we decided to drop support for iCubInterface for this particular feature. Drawback will be that if an external application will try to call the new interface after robot startup it will simply not work, which means yarpmotorgui calibrate button will not work.

I'm going to merge this PR

PR #490, opened 2015-05-15
Use ACE::abort instead of ACE::exit to kill RFModule

barbalberto:
Add yarp::os::abort to wrap ace::abort.
Use abort to kill RFmodule at 4th CRTL+C, because exit is not recommended for multithread process according to http://www.cplusplus.com/reference/cstdlib/exit and causes random segfaults when used.

### DIFF
--- a/src/libYARP_OS/src/RFModule.cpp
+++ b/src/libYARP_OS/src/RFModule.cpp
@@ -253,6 +253,16 @@ static void handler (int sig) {
         //printf("sent %s, got %s\n", cmd.toString().c_str(),
         //     reply.toString().c_str());
    // }
+
+#if defined(WIN32)
+	//on windows we need to reset the handler after beeing called, otherwise it will not be called anymore.
+	//see http://www.geeksforgeeks.org/write-a-c-program-that-doesnt-terminate-when-ctrlc-is-pressed/
+
+	//Additionally, from http://www.linuxprogrammingblog.com/all-about-linux-signals?page=show 
+	//The signal(2) function is the oldest and simplest way to install a signal handler but it's deprecated. 
+	// There are few reasons and most important is that the original Unix implementation would reset the signal handler to it's default value after signal is received.
+	ACE_OS::signal(SIGINT, (ACE_SignalHandler)handler);
+#endif
 }
 
 // Special case for windows. Do not return, wait for the the main thread to return.
--- a/src/libYARP_OS/src/RFModule.cpp
+++ b/src/libYARP_OS/src/RFModule.cpp
@@ -236,7 +236,7 @@ static void handler (int sig) {
     ct++;
     if (ct>3) {
         ACE_OS::printf("Aborting (calling exit())...\n");
-        yarp::os::exit(1);
+        yarp::os::abort();
     }
     ACE_OS::printf("[try %d of 3] Trying to shut down\n",
                    ct);
--- a/src/libYARP_OS/src/RFModule.cpp
+++ b/src/libYARP_OS/src/RFModule.cpp
@@ -235,7 +235,7 @@ static void handler (int sig) {
     static int ct = 0;
     ct++;
     if (ct>3) {
-        ACE_OS::printf("Aborting (calling exit())...\n");
+        ACE_OS::printf("Aborting (calling abort())...\n");
         yarp::os::abort();
     }
     ACE_OS::printf("[try %d of 3] Trying to shut down\n",
--- a/src/libYARP_OS/include/yarp/os/Os.h
+++ b/src/libYARP_OS/include/yarp/os/Os.h
@@ -39,7 +39,7 @@ namespace yarp {
         /**
         * Wrapper for ACE_OS::abort().
         */
-        YARP_OS_API void abort(void);
+        YARP_OS_API void abort(bool verbose=false);
 
         /**
         * Wrapper for ACE_OS::getenv().
--- a/src/libYARP_OS/include/yarp/os/Os.h
+++ b/src/libYARP_OS/include/yarp/os/Os.h
@@ -37,6 +37,11 @@ namespace yarp {
         YARP_OS_API void exit(int v);
 
         /**
+        * Wrapper for ACE_OS::abort().
+        */
+        YARP_OS_API void abort(void);
+
+        /**
         * Wrapper for ACE_OS::getenv().
         * @param v string that containt the environment variable name
         * @return the value corresponding to the envarionment variable v
--- a/src/libYARP_OS/src/Os.cpp
+++ b/src/libYARP_OS/src/Os.cpp
@@ -37,8 +37,13 @@ yarp::os::YarpSignalHandler yarp::os::signal(int signum, yarp::os::YarpSignalHan
 
 void yarp::os::exit(int v)
 {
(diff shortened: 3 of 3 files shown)


lornat75:
Calling exit is not a clean way to shutdown a process, but in the implementation of RFModule it was intended as a last resort in case the process is blocked and cannot return gracefully. Normally an RFModule process should be written in such a way that it never calls exit.

How is abort() better than exist() in this case?

barbalberto:
Yes, when you get there, it means there is no way (or no intention) to recover and you just want to kill the process  asap, so either exit or abort will do the job.

The only difference is that exit() tries to perform some clean up by calling the destructor for SOME of the objects. When you have multiple threads, then those threads will try to access data that no longer exists therefore you get a segfault, while abort is like a "kill -9", it does not try to clean things up.

So is that when you call exit(), the user will see a segfault and will point at it as a code bug in the cleanup while actually the segfault is caused by the exit itself. This was one of the segfaults Marco was asking to remove but it is simply not possible if we use the exit.

So the difference is simply in the printout the user will see:
Using exit:

Aborting (calling exit())...
segfault

using abort:

Aborting (calling exit())...
aborted

Since the result for exit and abort is the same, the message for the user will be misleading for the exit while more coherent with expectation for abort

pattacini:
That's true, but why in the code are we reaching the call to `exit() / abort()`?
There must be something hindering a graceful closure.

pattacini:
I mean, the call to `abort()` could be better than `exit()` but this somehow doesn't seem to solve the problem...

barbalberto:
yeah, but there is no problem to be solved. It reached there when you start the robot with the button pressed and don't want to wait for all the calibration timeouts, that's the main case ;)

pattacini:
Eh yes... because the code handling the timeout needs to be notified about the `ctrl+c` event and ... too much work I suppose :wink:

barbalberto:
When you press the ctrl+c the robotinterface starts the closing procedure which calls the abort function which stops the calibration / park procedure. But not immediatly, it has to perform some action in order to safely interrrupt and clean-up (this is hw dependent), so it is already done... still it looks it takes too much time for the user

pattacini:
ok, thanks.

randaz81:
abort() is the correct way to kill a multithreaded application, as documented by various C++ guidelines. Indeed the reason of our segfault is probably due to some static object, whose lifespan is not easily predictable (I remember a similar issue related to some ACE statics, fixed by Paul in the past) 

http://www.cplusplus.com/reference/cstdlib/exit/
http://www.cplusplus.com/reference/cstdlib/abort/
According to the documentation:
>"Data races:
Calling this function destroys all objects with static duration: A program with multiple threads running shall not call exit (see quick_exit for a similar function that does not affect static objects)."

The new c++ 11 implements:
http://www.cplusplus.com/reference/cstdlib/quick_exit/
According to the documentation: 
> "No additional cleanup tasks are performed: No object destructors are called. "

drdanz:
Just for reference, lintian warns when a shared library contains calls to 'exit()' or '_exit()' ([shlib-calls-exit](https://lintian.debian.org/tags/shlib-calls-exit.html)). I'm not sure why it does not mention abort(). Also I'm not sure if catching signals from a library is a good thing...

Anyway, I don't know if `abort()` is the solution... It will not cause the segfaults because it will just raise a SIGABRT signal that, if not caught in the main, will cause the program to be aborted, not calling any of destructors or the registered atexit functions, and therefore it might leave the system in an inconsistent state, depending on what is happening in the other threads. Also if in the main you install a signal handler for SIGABRT, the execution might not exit (I think).

In C++11 a new [quick_exit()](http://stackoverflow.com/questions/9758495/what-is-the-difference-between-stdquick-exit-and-stdabort-and-why-was-stdq) function was added, it does not call destructors, but still gives the opportunity to set exit functions by calling the functions registered using `at_quick_exit()`, I'd suggest we use this function if C++11 support is enabled, i.e. something like:

```c++
void yarp::os::quick_exit()
{
#ifdef YARP_HAS_CXX11
    std::quick_exit();
#else
    ACE_OS::abort(); // or perhaps ACE_OS::exit()
#endif
}
```

lornat75:
I think there is no good way to solve this. We are trying to terminate a process while it is running without giving it the chance to perform shutdown. Personally I would not consider the segfault an issue in this case.

barbalberto:
I agree with @lornat75. The point of this PR was not to solve the "good way to quit", which I think it is unfeasible, but just to have a nicer quit message, more inline with the user expectation.

drdanz:
I'm not very sure about what the user will expect at the fourth Ctrl+C :grin:

barbalberto:
not a segfault for sure.

lornat75:
If you are sure that abort behaves better than exit  think we can accept this PR. Have you tested this PR with RFModules in case the updateModule() is blocked on a semaphore or on a port?

barbalberto:
Test condition was a program sending trajectory to iCubSim:

I tried with the yarp::os::Semaphore by putting

```c++
 bool CYexample::updateModule()  {
    static long int keepAlive = 1;
    if(keepAlive%1000 == 0)
    {
        std::cout << " Called updateModule";
        std::cout << " Calling mutex.wait first time!" << std::endl;
        _mutex.wait();
        std::cout << " Calling mutex.wait second time!" << std::endl;
        _mutex.wait();
        printf("CYexample updateModule AFTER 2 mutex wait\n"); fflush(stdout);
    }

and it quits fine. 

Second test opening from terminal a read port
   yarp read /testPort 

Writing a message with expected reply from the updateModule
```c++
 bool CYexample::updateModule()  {
    static long int keepAlive = 1;
    if(keepAlive%1000 == 0)
    {
        Bottle cmd;
        Bottle response;
        cmd.addVocab(VOCAB_GET);
        cmd.addVocab(VOCAB_AXES);

        std::cout << "\n\nsending request " << cmd.toString() << std::endl;
        bool ok = test_port.write(cmd, response);
        std::cout << "Response is " << response.toString() << std::endl;

The program get stuck on the write as expected ("response is" is never printed), by pressing CTRL+C 4 times it exit correctly.

Output on the terminal is:
 Called updateModule Calling mutex.wait first time!
 Calling mutex.wait second time!
^C[try 1 of 3] Trying to shut down
^C[try 2 of 3] Trying to shut down
^C[try 3 of 3] Trying to shut down
^CAborting (calling exit())...
Aborted

Shall I put a quit test in the newly created test framework?

File: src/libYARP_OS/src/RFModule.cpp [hunk 1]
```diff
@@ -236,7 +236,7 @@ static void handler (int sig) {
     ct++;
     if (ct>3) {
         ACE_OS::printf("Aborting (calling exit())...\n");
```
pattacini: this message should be instead: `"calling abort()...\n"`

pattacini: or just "Aborting..." might be fine

barbalberto: It make sense! Did it in last commit

barbalberto:
If there is anything solid **against** this PR i think it can be safely merged

lornat75:
what about the behavior on Windows?

PR #495, opened 2015-05-27
Propagate return values alongside the data

barbalberto:
Each value broadcasted inside the stateExt port is paired with a boolean flag telling if the value is valid or not.
The flag will be set to true if the corresponding getXXX function in the remote device driver was succesful or false otherwise.
In this way if the user application tries a getMotorAcceleration and the robot/simulator does not implement this function (or it fails somehow) the return value in the user application will be false.
Right now, since all broadcasted values are always available the return value was always true but I think this is not correct.
This will re-establish the same behaviour of the former version of yarp without the stateExt.


lornat75:
This change looks ok to me.

randaz81:
now getEncoders() returns false and yarpmotorgui does not open anymore... please test before merge!  (and add a test case on the return values in the test suite...)

PR #497, opened 2015-06-01
Refreshed documentation

lornat75:
Still need to update the page "dependencies". We need to update the instructions as soon as we release the new binaries.

@drdanz


drdanz:
@lornat75 Thanks!

lornat75:
We are still missing the MacOS page :) any volunteer? @francesco-romano  :)

vtikha:
Why not do a similar documentation as the one I had done for the [wiki](http://wiki.icub.org/wiki/Homebrew_Mavericks)? 
It should be straightforward to transfer right?

kt10aan:
There is also this one in the wiki, that is up-to-date: [Homebrew](http://wiki.icub.org/wiki/Homebrew)

lornat75:
Of course the idea is to bring in the documentation from the iCub wiki related to YARP (this is what I have done myself). Which one of the two you suggest is the one more appropriate?

For consistency I also suggest to follow the same structure of the page I have uploaded.

kt10aan:
The two pages are similar, with the only difference being that [this one](http://wiki.icub.org/wiki/Homebrew) is more up to date.

In order to follow the same structure as the one you uploaded, the dependencies should be put on a different page?

lornat75:
uhm, no. In the Linux/Windows installation pages I put self-consistent installation instructions that when possible do not include compilation of dependencies (because they are available as precompiled packages). If for MacOS the situation is different feel free to add compilation instructions for dependencies. I hope I made myself clear.

vtikha:
@kt10aan Well not quite, the page you are referring to is missing some parts. Anyway, this is not a competition :-) We should take into account & merge both pages while creating the documentation.

kt10aan:
@vtikha I actually came across the wiki/Homebrew_Mavericks page a few days ago.. talking about double effort :D. Merging them is the best solution (Notably, the wiki/Homebrew page is missing the command line tools installation and the Qt5_DIR part.) Actually we should have merged them from the start

A couple of questions:
- Why you suggest to install wget? I never had to in order to install/compile anything in Homebrew (and YARP)
- GDK PIXBUF is for the GTK GUIs. Is it fixing something? (I haven't used the GTK guis on OS X for ages...). Or is it for that annoying bug on the iCubSIM?

Finally,  ```open -a TextEdit .bash_profile``` ...?? You are being mean to OS X users, I think they can use vim, emacs or at least nano :D:D

vtikha:
@kt10aan 
- wget was to give the opportunity to users to get sources (eg opencv) and compile it by hand, as one might need custom cmake flags.
- This could be removed as we are now fully on QT5. Hence important lack in Homebrew page. 
- 'open -a TextEdit .bash_profile' is an example and can be changed in order to use any editor. This is not a personal choice but since TextEdit is there by default it was chosen as per various beginner user requests. Plus you can notice that the step by step instructions are very detailed for this very reason.

kt10aan:
@vtikha The point you made about qt5 reminded me that the homebrew yarp formula has to be also updated upon the next yarp release. As it is now, it lists gtk as a hard dependency and doesn't mention qt5 at all.
My point about TextEdit was a joke :D (I honestly haven't opened TextEdit since 10.2!). 
And indeed your instructions were extremely detailed :)

francesco-romano:
@vtikha there is `curl` by default installed.

lornat75:
Added temporary link to http://wiki.icub.org/wiki/MacOSX:_installation

@vitkha: who can I assign to this task for MacOS so we move installation instructions from the wiki to the yarp documentation?

vtikha:
@lornat75, I will take care of it and commit it as soon as I can.

PR #500, opened 2015-06-07
added IRemoteVariables interface

randaz81:
the interface is composed by the following methods:
* virtual bool getRemoteVariable(yarp::os::ConstString key, yarp::os::Bottle *val) = 0;
* virtual bool setRemoteVariable(yarp::os::ConstString key, const yarp::os::Bottle val) = 0;
* virtual bool getRemoteVariablesList(yarp::os::Bottle* listOfKeys) = 0;
The advanced developer can use them to set/get internal variables which are not exposed by the remote device with a proper yarp interface.
This interface should be used for debug purposes only. To be NOT used in standard applications.


src/libYARP_dev/include/yarp/dev/IRemoteVariables.h
```diff
@@ -0,0 +1,68 @@
...
+{
+public:
+    /**
+     * Destructor.
+     */
+    virtual ~IRemoteVariables() {}
+
+    virtual bool getRemoteVariable(yarp::os::ConstString key, yarp::os::Bottle *val) = 0;
+
+    virtual bool setRemoteVariable(yarp::os::ConstString key, const yarp::os::Bottle val) = 0;
```
lornat75: const usually goes with a reference, did you mean const yarp::os::Bottle &val ?

src/libYARP_dev/include/yarp/dev/IRemoteVariables.h
```diff
@@ -0,0 +1,68 @@
...
+ */
+class YARP_dev_API yarp::dev::IRemoteVariables
+{
+public:
+    /**
+     * Destructor.
+     */
+    virtual ~IRemoteVariables() {}
+
+    virtual bool getRemoteVariable(yarp::os::ConstString key, yarp::os::Bottle *val) = 0;
```
lornat75: Maybe we can use c++ reference yarp::os::Bottle &val

src/libYARP_dev/src/modules/ControlBoardWrapper/RPCMessagesParser.cpp
```diff
@@ -2081,6 +2089,22 @@ bool RPCMessagesParser::respond(const yarp::os::Bottle& cmd, yarp::os::Bottle& r
                             }
                             break;
 
+                            case VOCAB_VARIABLE:
```
lornat75: I prefer if these conditions are checked inside a handleIRemoteVariable function to reduce the size of the big switch case; this is how it was done for new interfaces (i.e. torque control mode etc)

randaz81:
All features requests committed. Merging..

PR #505, opened 2015-06-12
removed some printfs, replaced with proper logging functions

randaz81:
tested with win32 msvc12, using Travis CI  to check for typos on different target platforms


randaz81:
tested with win32 msvc12, using Travis CI  to check for typos on different target platforms

PR #509, opened 2015-06-15
IBattery interface

randaz81:
Includes:
* IBattery interface
* BatteryWrapper 
* BatteryClient 
* yarpbatterygui-qt


src/libYARP_dev/src/modules/BatteryClient/BatteryClient.cpp
```diff
@@ -0,0 +1,298 @@
...
+    {
+        av=deltaT/count;
+    }
+    av=av*1000;
+    mutex.post();
+}
+
+bool yarp::dev::BatteryClient::open(yarp::os::Searchable &config)
+{   
+    ConstString carrier = config.check("carrier", Value("udp"), "default carrier for streaming robot state").asString().c_str();
```
lornat75: some leftover from previous code...

lornat75:
Are you sure you need three ports? maybe an rpc port may be enough in this case?

Next time I would use a thrift definition file, it will save you from the trouble of writing all the code and will be easier to maintain!

PR #517, opened 2015-06-22
Forced the refresh of the nameserver properties (host, port)

francesco-romano:
When `setNameServerContact` was called the properties were correctly saved to the configuration file, but it was not reloaded (until next run of yarp).
I forced the reload by calling `activate` with the `true` flag.
I don't know if this is the most correct way.

cc @lornat75 

PS:
I think that nobody (but me) uses this function, so I think I'll not break any compatibility
PPS:
@lornat75 the fact that all configurations passes through file writing causes some issues in some configuration, e.g. read-only filesystem, etc.. I think this is quite old yarp code, but maybe we can open a (low priority) issue on making optional the writing, and/or allowing yarp to be configured also in-memory. cc @alecive

### DIFF
--- a/src/libYARP_OS/src/Network.cpp
+++ b/src/libYARP_OS/src/Network.cpp
@@ -1279,7 +1279,9 @@ bool NetworkBase::setNameServerContact(Contact &nameServerContact)
         setNameServerName(nameServerContact.getName());
     nameConfig.fromFile();
     nameConfig.setAddress(nameServerContact);
-    return nameConfig.toFile();
+    bool result = nameConfig.toFile();
+    getNameSpace().activate(true);
+    return result;
 }
 
 


lornat75:
I though the nameserver was already by default writing to memory. Am I wrong?

lornat75:
I agree with your comment. However we need a way to cache nameserver parameters to avoid searching for it all the time.

francesco-romano:
What do you mean with "caching"?
In case the we use config files, then the first time they are loaded the configuration can stay in memory (or until a force reload)
In case we use a in-memory approach then.. there is only the cache.
Am I right?
I don't understand if you are referring to the PR or to my comment

alecive:
@lornat75 shall we merge this? I would still open an issue about the need of avoiding writing nameserver parameters to file, but this pull request in my opinion can be safely merged.

lornat75:
yes. I meant to comment on Francesco's comment.

PR #519, opened 2015-06-23
Change controlBoardWrapper2 input streaming port to BufferedPort, add…

pattacini:
… read Strict option

Add check if more then 20 messages are in queue to be read, warn the user of
a possible overflow (reader not able to handle all messages fast enough)


pattacini:
fix https://github.com/robotology/icub-main/issues/129

PR #523, opened 2015-07-01
Fix FfmpegGrabber to work with recent ffmpeg

Tobias-Fischer:
Hi,

I know most people use the OpenCVGrabber, but we needed to get the FfmpegGrabber working with a recent version of Ubuntu (14.04). This pull request contains the changes needed.

Best,
Tobias

### DIFF
--- a/src/modules/ffmpeg/FfmpegGrabber.cpp
+++ b/src/modules/ffmpeg/FfmpegGrabber.cpp
@@ -569,7 +569,7 @@ bool FfmpegGrabber::open(yarp::os::Searchable & config) {
 
 
     // Retrieve stream information
-    if(YARP_av_find_stream_info(pFormatCtx)<0) {
+    if(YARP_av_find_stream_info(pFormatCtx, NULL)<0) {
         printf("Could not find stream information in %s\n", fname.c_str());
         return false; // Couldn't find stream information
     }
@@ -579,7 +579,7 @@ bool FfmpegGrabber::open(yarp::os::Searchable & config) {
 
     if (pFormatCtx2!=NULL) {
         
-        if(YARP_av_find_stream_info(pFormatCtx2)<0) {
+        if(YARP_av_find_stream_info(pFormatCtx2, NULL)<0) {
             printf("Could not find stream information in %s\n", fname.c_str());
             return false; // Couldn't find stream information
         }
--- a/src/modules/ffmpeg/FfmpegWriter.cpp
+++ b/src/modules/ffmpeg/FfmpegWriter.cpp
@@ -91,7 +91,7 @@ static AVStream *add_audio_stream(AVFormatContext *oc, CodecID codec_id)
     AVCodecContext *c;
     AVStream *st;
 
-    st = av_new_stream(oc, 1);
+    st = avformat_new_stream(oc, NULL);
     if (!st) {
         fprintf(stderr, "Could not alloc stream\n");
         ::exit(1);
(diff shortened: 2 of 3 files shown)


pattacini:
I feel we can merge it, if no one has something to object ...

kt10aan:
Just a quick question, were there any checks against other systems, e.g., Debian Jessie. Thankfully, this ffmpeg mess will be over soon, as Debian (and therefore Ubuntu), abandoned libav in favour of the proper ffmpeg for their next release :)

drdanz:
I'm all in favour of removing deprecated stuff but we must ensure that this will not break compatibility with the other supported releases.
Also, do anyone know if we have ffmpeg packases on windows? If we do, we should check that as well...

pattacini:
I knew the fear of merging would have elicited a discussion :smirk:

Tobias-Fischer:
Hi,
I can understand your hesitations to merge the changes when it comes to the support of old platforms. I guess some old versions of ffmpeg will break when these changes are applied, but I haven't tested any.

Unfortunately, I don't have time to check against the various versions of ffmpeg which are around. I also guess that the function calls I had to replace were not all introduced in the same version of ffmpeg, which makes version checking even harder. The APIchanges (http://git.videolan.org/?p=ffmpeg.git;a=blob_plain;f=doc/APIchanges;hb=refs/heads/master) might provide a clue at what time which method got deprecated.

I'll leave this pull request around for others which can't compile the FfmpegGrabber in Ubuntu 14.04, but won't put more work into it.

Best, Tobias
PS: I guess for newer versions of Ubuntu even with these changes the grabber might not compile, as I still get dozens of deprecated warnings ..

kt10aan:
@Tobias-Fischer, its not about older platforms. It is about if it works with all platforms supported by YARP, that happen to have slightly different versions of ffmpeg (or libav) in their repositories.

See http://wiki.icub.org/wiki/YARP_Supported_Distributions

The proper (but time consuming...) way to address the issue is to have a bunch of conditionals in the code, for each operating system and its corresponding ffmpeg version.

Tobias-Fischer:
@kt10aan: Sorry, yes, that's what I meant to say :). I understand that conditionals are the way to go, however I don't have time to go ahead and implement them unfortunately ..

drdanz:
I just updated the wiki page, I'll do some testing with a few versions.

drdanz:
@Tobias-Fischer Which ffmpeg/libav version are you using? I wasn't able to compile your branch on debian testing. Tried both libav 11.4 and ffmpeg 2.7.1.

Also I noticed that you changed
```diff
- #define YARP_av_find_stream_info av_find_stream_info
+#define YARP_av_find_stream_info avformat_find_stream_info

 #ifdef USE_AVFORMAT_OPEN_INPUT
 #undef YARP_av_find_stream_info
 #define YARP_av_find_stream_info(a) avformat_find_stream_info(a,NULL)
```

and then
```diff
-if(YARP_av_find_stream_info(pFormatCtx)<0) {
+if(YARP_av_find_stream_info(pFormatCtx, NULL)<0) {
```

but that was supposed to work already, and (if it is wrong) only the definition of ``USE_AVFORMAT_OPEN_INPUT`` should be changed

drdanz:
by the way, I was able to build it using both libraries in the original version

Tobias-Fischer:
Hi @drdanz: I am using Version: 6:9.18-0ubuntu0.14.04.1 of libav. And yes, it seems like the definition of `USE_AVFORMAT_OPEN_INPUT` should be changed in this case. I didn't see the `ifdef`, sorry.

drdanz:
@Tobias-Fischer The fix was easier, just fixing the definition of ``USE_AVFORMAT_OPEN_INPUT`` was enough to build on Ubuntu Trusty, therefore I'm going to close this pull request. Can you try and confirm that it works for you as well?

Tobias-Fischer:
Hi @drdanz: Yes, that did the job. Thanks!

PR #525, opened 2015-07-02
[ServerInertial] Add config parameter for forcing strict write

drdanz:


lornat75:
Is there a good reason for this option? We normally discard "old" messages to reduce latency in control loop. This has been a good policy so far with the exception of controlboard client that started to had problems when we introduced mixed commands that can be sent in tight loop across the network.

pattacini:
I think @drdanz needed it while experimenting with oculus rift; he can be more precise.

drdanz:
Yes, exactly, I wanted to be sure that all packages are sent and received, to be able to receive the measurement closer to the timestamp of the camera image. It is not enabled by default, so the behaviour is unchanged unless this is explicitly enabled

Issue #510, opened 2015-06-16
Homebrew YARP formula


kt10aan:
I was very glad to see the updated OS X documentation page at http://wiki.icub.org/yarpdoc/install_yarp_mac.html

@vtikha, the instructions were very neat and tidy :)

I would like to ask about the fate of the YARP Homebrew formula. Since it is not mentioned in the updated docs, is it abandoned? Any plans of updating it? (some extensive updating of the formula is needed as dependencies have changed a lot since the last version). Also, a reminder, that the YARP formula was moved by Homebrew maintainers to the homebrew/x11 tap (https://github.com/Homebrew/homebrew-x11) as it was asking for a hard X11 dependency, something that is not the case now with Qt. Or we are waiting for the new release? :):)

lornat75:
I would like to keep the brew formula; I think @vtikha copied instructions for compiling yarp from sources, while yarp's brew formula installs from a release/snapshot (which is the equivalent of the "install from binaries" sections in the Linux and Windows instructions). @kt10aan: can you add this section yourself?

kt10aan:
I will add the section, but first a few issues to be decided:

I updated the yarp formula in my forked homebrew-x11 repository (as it was severely outdated). You can find it here: https://github.com/kt10aan/homebrew-x11/blob/master/yarp.rb

Changes include:
- Updated release version to 2.3.64 and corresponding sha1
- Removed the obsolete gtk dependencies and, as a result, the X11 dependency
- Changed some CMake flags in the old formula to reflect better the current YARP default CMake flags (e.g., no GTK, no GYarpmanager etc)
- Enabled by default the serial and serialport devices
- The formula now has three options: 
  - --with-qt5, that builds the Qt GUIS and installs qt5 as a dependency (no Qt5_DIR variable need to be set that way)
  - --with-yarprun-log, that enables yarprun_log
  - --with-opencv, that enables the opencv_grabber device and installs OpenCV as a dependency

I could add more options, like bindings for python and lua (and adding swig as optional dependency) but for now python bindings are broken in OS X (see #508).

Two questions:
- Do you think these are sane defaults and options or do you want something different (e.g., having the Qt5 guis enabled by default)?

- Since X11 is not a hard dependency anymore, the formula does not have to be in homebrew-x11. It can moved either to main homebrew or in homebrew-science. Do you agree/want that?

You can find the full diff of the changes here: https://gist.github.com/6309a80872ba33ee3a23

kt10aan:
[...]

PR #530, opened 2015-07-14
Update OS X YARP install docs

kt10aan:
Updated the OS X documentation to better reflect the updated YARP formula.

Specifically:
- Broadly divided in 3 major section: Installing prerequisites, installing YARP from Homebrew, installing YARP from source (and its dependencies from Homebrew)
- Removed references to installing X11 (and subsequently the PKG_CONFIG env variable. All gtk-related homebrew packages are X11-free now (and with a better Quartz theme!).

See also: #510 

@vtikha, could you please review the PR and change things that you think could be confusing to the user? I kept your writing style as much as I could, so that the instructions appear more uniform. Thanks!

All commits squashed and rebased :)

### DIFF
--- a/doc/installation/install_mac.dox
+++ b/doc/installation/install_mac.dox
@@ -5,11 +5,11 @@
  */
 
 /**
- * @page install_yarp_mac Installing YARP on MacOS
+ * @page install_yarp_mac Installing YARP on MacOS X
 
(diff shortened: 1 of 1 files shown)


vtikha:
@kt10aan Excellent, thank you very much for your contribution

PR #534, opened 2015-07-17
Update docs about bash completion on OS X

kt10aan:
Just a simple addition, regarding the instructions for enabling bash completion in OS X.

The commit was rebased :)


drdanz:
Thanks!

PR #535, opened 2015-07-18
Added option to skip binaries compilation

francesco-romano:
Added `YARP_BUILD_LIBRARIES_ONLY`, by default to NO.
If set to YES only libraries will be build: no `add_executable` cmake command will be called.

I could have missed something, so double check is welcome.
The rationale behind this is to allow the compilation only of yarp libraries

cc: @lornat75 @drdanz @vtikha @alecive


francesco-romano:
Now the option is hidden by default (marked as advanced)

lornat75:
Looks good to me!

francesco-romano:
@lornat75 good!
@drdanz As soon as we start working on task at VVV I need this PR merged. So, I think by tomorrow afternoon I'll merge if it is ok with you.

PR #546, opened 2015-07-20
[YARP_OS] Always build port authentication, but disable it if not explicitly enabled

### LINKED ISSUE #99: USE_PORT_AUTHENTICATION does not use resource finder to locate config file
If configured with -DUSE_PORT_AUTHENTICATION:BOOL=ON you will get messages like

```
  yarp: cannot read from <$HOME>/.yarp/conf/user.conf
```

because in `AuthHMAC.cpp` the user.conf file is still searched in this way:

``` C++
    String fname = String(NetworkBase::getEnvironment("HOME", &found).c_str()) + 
        "/.yarp/conf/user.conf";
```

### LINKED ISSUE #121: USE_PORT_AUTHENTICATION confuses users
In many many cases naive users compile yarp from sources, and enable USE_PORT_AUTHENTICATION (probably they think: why not?), with the result that they cannot establish connections. The error message is ambiguous.

We should either:
1) hide the flag so that only expert users use it
2) make the error message reported by the connect clearer to interpret. In any case fixing this requires re-compiling YARP which is an overhead so I would go for 1).

### LINKED ISSUE #135: USE_PORT_AUTHENTICATION should not mention the old config if it does not exist
Current log when both auth.conf and .yarp/conf/user.conf do not exist is:
```
Cannot find auth.conf using ResourceFinder, fallback to .yarp/conf/user.conf
Please move authentication to an auth.conf findable by 'yarp resource'
yarp: cannot read from /home/<user>/.yarp/conf/user.conf
No key for authentication found. Using default.
```
In order not to encourage the user to use the old file, the message should be just something like this
```
Cannot find auth.conf using ResourceFinder
No key for authentication found. Using default.
```
And eventually mention the user.conf file if this exists

drdanz:
CC-Issue #99
CC-Issue #121
Fixes #135


drdanz:
Oops, I just realized that #99 and #121 are already closed!

pattacini:
:unamused:

Issue #542, opened 2015-07-20
manager on windows opens a console when processes are executed on localhost


lornat75:

apaikan:
we can try this: http://www.codeproject.com/Articles/2537/Running-console-applications-silently

But i am not sure then the console apps can receive signals.

lornat75:
StartupInfo.dwFlags = STARTF_USESHOWWINDOW;
	StartupInfo.wShowWindow = SW_HIDE;

This works, however I think it also affects the guis :(

I would like to see if we can actually take advantage of this and use the consoles for input/output....

lornat75:
This fix caused shutdown problems. In windows when processes are created with their own console they no longer receive termination signals from parent (well, at least doing this is not obvious).

PR #549, opened 2015-07-23
Console with localbroker

lornat75:
This PR adds support to create process that start visible, invisible, or minimized. Implemented for Windows with localhost. Useful for console applications.

Examples:

\<display>--minimized\</display> 

\<display>--hidden\</display> 

...


lornat75:
Addresses #542

PR #552, opened 2015-07-25
implementing Property.fromCommand() for command line arguments given as a string

apaikan:
This modification adds Property.fromCommand(char* param, bool wipe) method to the Property class to parse the command line arguments given as a string.  For example:

```c++
yarp::os::Property prop;
prop.fromCommand("--verbose --show --from test.config");
cout<<"prop.toString()"<<endl;
...
```
Result: 
```
(show) (from test.config) (verbose)
```

### DIFF
--- a/src/libYARP_OS/include/yarp/os/Property.h
+++ b/src/libYARP_OS/include/yarp/os/Property.h
@@ -188,16 +188,16 @@ public:
     /**
      * Interprets a list of command arguments as a list of properties.
      * Keys are named by beginning with "--".  For example, with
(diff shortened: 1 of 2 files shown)


apaikan:
The `Property.fromCommand(char* param, bool wipe)` API has changed to `Property.fromArguments(char* argument, bool wipe)` for clarity.

```
yarp::os::Property prop;
prop.fromArguments("--verbose --show --from test.config");
cout<<"prop.toString()"<<endl;
...

PR #554, opened 2015-08-02
implements Networkbase::setConnectionQos() API

apaikan:
This pull-request implements a new API `Networkbase::setConnectionQos()` to adjust the QoS preferences of a connection. The QoS preferences are applied to the both side of a connection. That adjusts the communication's thread real-time properties (the thread which handles asynchronous read/write within a port object) and the IPV4/6 packets DSCP value. These features had already been implemented in YARP but now have been made available using the  `setConnectionQos()` API.  

Here is an example: 

```c++
yarp::os::QosStyle qos;
qos.threadPriority = 10;
qos.threadPolicy = 1;  // SCHED_FIFO on Unix
qos.packetPriority = yarp::os::QosStyle::HIGH; // DSCP 36
Network::setConnectionQos("/output", "/input", qos, qos);
```
The above example adjust the communication thread's (if available) real-time properties of the connection ("/output" -> "/input") to be scheduled using `SCHED_FIFO` with priority=10. The packets DSCP value is also set to 36 using  `yarp::os::QosStyle::HIGH`. The API offers 4 levels of predefined packet priorities as defined in `yarp/os/QosStyle.h` : 

```c++
    enum PacketPriorityType {
        LOW = VOCAB3('L','O','W'),   // DSCP = 10
        NORM = VOCAB4('N','O','R', 'M'),   // DSCP = 0
        HIGH = VOCAB4('H','I','G','H'),   // DSCP = 36
        CRIT = VOCAB4('C','R','I','T')   // DSCP = 44
    };
```

### DIFF
--- a/bindings/yarp.i
+++ b/bindings/yarp.i
@@ -378,6 +378,7 @@ namespace yarp {
 %include <yarp/os/RpcClient.h>
 %include <yarp/os/DummyConnector.h>
 %include <yarp/os/Things.h>
+%include <yarp/os/QosStyle.h>
 
 %define MAKE_COMMS(name)
 %feature("notabstract") yarp::os::BufferedPort<name>;
--- a/src/libYARP_OS/CMakeLists.txt
+++ b/src/libYARP_OS/CMakeLists.txt
@@ -117,7 +117,8 @@ set(YARP_OS_HDRS include/yarp/os/AbstractCarrier.h
                  include/yarp/os/YarpPluginSelector.h
                  include/yarp/os/YarpPluginSettings.h
                  include/yarp/os/SystemInfo.h
-                 include/yarp/os/SystemInfoSerializer.h)
+                 include/yarp/os/SystemInfoSerializer.h
+                 include/yarp/os/QosStyle.h)
 
 if(NOT YARP_NO_DEPRECATED)
   list(APPEND YARP_OS_HDRS include/yarp/os/begin_pack_for_net.h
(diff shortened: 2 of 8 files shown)


apaikan:
The `Networkbase::setConnectionQos()` does not have any affect on the platforms (e.g., Windows) which does not support real-time configuration or packet prioritization.

drdanz:
Thanks, it looks quite good, just a couple of comments:


* The core of `NetworkBase::setConnectionQos` looks like a lot of code duplication for source and destination ports, perhaps you could re-factor it to reduce code duplication
* I think it would be useful to have an overload that accepts only a QosStyle and sends the command only in both direction. Your example could then be re-written as
```c++
Network::setConnectionQos("/output", "/input", qos);
```
* It would be nice to have also a `NetworkBase::getConnectionQos`. Anyway I wonder what this should return, 2 QosStyles? Perhaps it would be better to have both sides in QosStyle and always pass only one to the setConnectionQos function?

Also nitpicks:
* quite -> quiet
* extra spaces added at line src/libYARP_OS/src/PortCore.cpp:1999

apaikan:
Thanks @drdanz 

- The core of `NetworkBase::setConnectionQos` looks ike a lot of code duplication... 
 Let keep it as it is because setting Qos preferences at source and destination of a connection is not done exactly in the same way. The reason is the destination (input) port does not support packet priority preferences. So separating the implementation in a third-party  function actually does not help for the code reuse and removing the dublication. 

- `NetworkBase::getConnectionQos` ... 
  There are still some issues which should be solved with this API. Adjusting Qos properties are done using the yarp administrative commands and using the "prop set/get". The original implementation of "prop set/get" in the admin port does not check for the validity of the properties. So basically whatever can be added to the admin property. Thus the get can return the currently set Qos properties but it does not ensure that the property were actually correctly applied. 

So let it be as it is until we come to better solution to check the validity of the commands. Considering this, the `NetworkBase::getConnectionQos` can be easily implemented without knowing the actual value of the applied Qos. 


The rest of the comments have been applied :)

drdanz:
@apaikan Merged, thanks!

PR #556, opened 2015-08-03
implementing QoS support in the remote_controlboard driver for real-time applications

apaikan:
This pull-requests implements the QoS support into the `remote_controlboard` driver. The driver can be opened using desired QoS preferences to prioritize the connections (e.g., `/icub/right_arm/state:o` to `/client/right_arm/state:i`). Thus, for example, a real-time controller can open a driver for a control board which has higher priority than other less-critical drivers/modules (e.g., yarpmotorgui). The QoS preferences can be individually configured for the local and remote ports of the driver. These preferences will be applied individually to each side of the connections if available. The QoS preferences are applied only for the `state`, `stateEx` and `Command` ports! 

 Here is a simple Lua script which demonstratess how the QoS properties can be configured for a driver:  

```lua
options = yarp.Property()
options:put("device", "remote_controlboard")
options:put("local", "/client/right_arm")
options:put("remote", "/icub/right_arm")

-- setting QoS preferences for the driver (local)
qos_local = options:addGroup("local_qos")
qos_local:put("thread_priority", 20)
qos_local:put("thread_policy", 1)
qos_local:put("packet_priority", yarp.Vocab_encode("HIGH"))  
-- C++ equivalent is qos_local.put("packet_priority", (int)QosStyle::HIGH)

-- setting QoS preferences for the driver (remote)
qos_remote = options:addGroup("remote_qos")
qos_remote:put("thread_priority", 30)
qos_remote:put("thread_policy", 1)
qos_remote:put("packet_priority", yarp.Vocab_encode("HIGH"))

-- open the driver
driver = yarp.PolyDriver(options)
...
```

- TODO: the similar implementaion should be done for the other drivers: `CartesianController`, `GazeController`, ...

### DIFF
--- a/src/libYARP_dev/src/modules/RemoteControlBoard/RemoteControlBoard.cpp
+++ b/src/libYARP_dev/src/modules/RemoteControlBoard/RemoteControlBoard.cpp
@@ -27,6 +27,7 @@
 #include <yarp/dev/ControlBoardInterfacesImpl.h>
 #include <yarp/dev/ControlBoardHelpers.h>
 #include <yarp/dev/PreciselyTimed.h>
+#include <yarp/os/QosStyle.h>
 
 #include <stateExtendedReader.hpp>
 
@@ -961,6 +962,29 @@ public:
         remote = config.find("remote").asString().c_str();
         local = config.find("local").asString().c_str();
 
+        // check the Qos perefernces if available (local and remote)
+        yarp::os::QosStyle localQos;
+        if (config.check("local_qos")) {
+            Bottle& qos = config.findGroup("local_qos");
+            if(qos.check("thread_priority"))
+                localQos.threadPriority = qos.find("thread_priority").asInt();
+            if(qos.check("thread_policy"))
+                localQos.threadPolicy = qos.find("thread_policy").asInt();
+            if(qos.check("packet_priority"))
+                localQos.packetPriority = (QosStyle::PacketPriorityType) qos.find("packet_priority").asInt();
+        }
+
+        yarp::os::QosStyle remoteQos;
+        if (config.check("remote_qos")) {
+            Bottle& qos = config.findGroup("remote_qos");
+            if(qos.check("thread_priority"))
+                remoteQos.threadPriority = qos.find("thread_priority").asInt();
+            if(qos.check("thread_policy"))
+                remoteQos.threadPolicy = qos.find("thread_policy").asInt();
+            if(qos.check("packet_priority"))
+                remoteQos.packetPriority = (QosStyle::PacketPriorityType) qos.find("packet_priority").asInt();
+        }
+
         bool writeStrict_isFound = config.check("writeStrict");
         if(writeStrict_isFound)
         {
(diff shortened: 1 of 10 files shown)


pattacini:
@apaikan remember to extend the documentation too, to keep track of such new functionalities.

pattacini:
The extension to the operational space drivers (i.e. `cartesiancontrollerclient`, `cartesiancontrollerserver`, `gazecontrollerclient`, `iKinGazeCtrl`) is not immediate in the sense that, aside from the standard motor interfaces opened up inside those drivers, we have also other ports and connections. Therefore, we should run an assessment somehow.

apaikan:
@pattacini  is it the right place to update the docs: http://wiki.icub.org/brain/icub_motor_control_tutorial.html

pattacini:
Uhm, that tutorial focuses on motor control, thus I'd avoid burdening it with details that are out of scope, in a sense, disorienting the readers.

Is there a page in yarp or elsewhere where `QoS` is described? That would be a good place. If not, we can hold on.

apaikan:
(+1) @pattacini  

That is good idea. Ill create a new doc page regarding the real-time and QoS features of Yarp.

src/libYARP_dev/src/modules/RemoteControlBoard/RemoteControlBoard.cpp
```diff
@@ -961,6 +962,29 @@ class yarp::dev::RemoteControlBoard :
...
         local = config.find("local").asString().c_str();
 
+        // check the Qos perefernces if available (local and remote)
+        yarp::os::QosStyle localQos;
+        if (config.check("local_qos")) {
+            Bottle& qos = config.findGroup("local_qos");
+            if(qos.check("thread_priority"))
+                localQos.threadPriority = qos.find("thread_priority").asInt();
+            if(qos.check("thread_policy"))
+                localQos.threadPolicy = qos.find("thread_policy").asInt();
```
barbalberto: I'd check here if the input is actually an integer or not. Unexperienced user may type things like "FIFO" as policy.
Do values have Min/Max ranges? Could worth check that too.

drdanz:
@apaikan +1 for the latest commit (0218ff9).
I cannot review the changes to the control board, because I don't know anything about that.

The only thing that does not look optimal is that if you want to set the qos only on one side of the connection, the command will be sent to both sides anyway.
Perhaps the default constructor of QosStyle should be invalid, so that you can just skip sending the commands?

apaikan:
@drdanz  and @barbalberto :  Thanks guys for the constructive comments. 

- I have modified the APIs of QosStyle class to adjust the packet priority in different manners: 

```c++
QosStyle style;
style.setPacketPriority(176); // directly using TOS bits
style.setPacketPriorityByDscp("AF42"); // using DSCP class type
style.setPacketPriorityByType(QosStyle::HIGH); // using our predefined priority levels 
```
The rest of the class attributes has their corresponding get/set methods (e.g.,  `setThreadPriority`, `getThreadPriority`, ...)

- the `setConnectionQos(...)` methods ignore sending admin commands to the relevant ports if the Qos properties given as the param left as default. This reduces the overhead of the redundant requests for the Qos to the remote admin ports.

drdanz:
@apaikan Merged, thanks

PR #568, opened 2015-08-25
Getting the port's owner process and thread ID using port admin commands  

apaikan:
This pull request add new functionality to the port admin commands (i.e., `prop get` ) to retrieve the process ID of the port's owner and the thread ID of the connection's owner (if exist).  The thread ID equal to -1, indicates that there is not any dedicated thread for the corresponding connections. 

Example: 
`/sender`  --->  `/receiver`

```
$ yarp admin rpc /sender
>> prop get /receiver 
Response: (sched ((priority 0) (pid 24192) (tid 24221) (policy 0))) (qos ((tos 0)))
```
If `/sender` does not have a dedicated thread for the corresponding connection then 
```
>> prop get /receiver
Response: (sched ((priority 0) (pid 24192) (tid -1) (policy 0))) (qos ((tos 0)))
```

### DIFF
--- a/src/libYARP_OS/include/yarp/os/SystemInfo.h
+++ b/src/libYARP_OS/include/yarp/os/SystemInfo.h
@@ -110,6 +110,15 @@ public:
     } NetworkInfo;
     */
 
+    /**
+     * @brief The ProcessInfo stuct provides the operating system proccess information.
+     */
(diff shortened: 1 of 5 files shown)


drdanz:
Looks good! Thanks

Issue #502, opened 2015-06-11
Neither interrupt nor close have effect on a blocking read if open was not called before on the port


lornat75:
Corner case: if you forget to open a port but call `read`, the calling thread get blocked forever (neither `interrupt` nor `close` unblock it). 

Probably the problem is not in the behavior of `interrupt` or `close`, rather the `read` should return with an error if called on a port that was not correctly opened.

apaikan:
closes on pull request #573.

PR #573, opened 2015-09-02
closes issue #502

apaikan:
@lornat75  wrote: Neither interrupt nor close have effect on a blocking read if open was not called before on the port!

This pull request close the corresponding issues by simply ignore the `Port:read()`, `BufferedPort::read()' and `RpcServer::read()` if the port is not open (active). 

Does it break the YARP-ROS interoperability?


drdanz:
Already fixed. closed.

PR #577, opened 2015-09-02
Vector of fix proposal2

lornat75:


drdanz:
I'm not sure how this builds on windows o_O

```diff
+template<class T> 
+  static int MapBottleTag() {
+    return tag;
+  }
```

`tag` is not declared anywhere, and obviously on linux it produces an `error: ‘tag’ was not declared in this scope`. Probably it should return `BOTTLE_TAG_DOUBLE`.

Lines 79 and 84 produce `error: explicit template specialization cannot have a storage class`, I think you just have to remove the ``static`` keyword from the function declaration (it is a function, not a class method, therefore the effect of the `static` keyword is that the function is available only in the current translation unit).

Also in my opinion the name ``MapBottleTag`` in a ``Vector`` class is a bit confusing. If I understand what you mean with that name, perhaps ``BottleTagMap`` would be a better name...

lornat75:
I have modified the PR to fix the problems you mention. 

Fixed use of static with inline.
Renamed the class (I did not like it either).
About the use of "tag"; I tried to add a default implementation that raises at least a runtime error. I think it is better like this because we want to make sure that the default implementation is never called. It would be better to make this error a compile time error in my opinion.

lornat75:
We really need to add a unit test to make sure we test the class VectorOf<> with various basic types (int, float etc). Templates are really tricky and errors depends a lot on how yarp is compiled. For example we should check if this works with yarp as DLL on windows, I suspect we need to add explicit instantiations in Vector.h (as it is done for VectorOf<double>

drdanz:
Still fails on linux...

```
src/libYARP_sig/include/yarp/sig/Vector.h:76:18: error: there are no arguments to ‘YARP_ASSERT’ that depend on a template parameter, so a declaration of ‘YARP_ASSERT’ must be available [-fpermissive]
```

PR #585, opened 2015-09-16
[tests] Improve EventTest

drdanz:
Test pull request to check coveralls integration.
Please do not merge.


drdanz:
Apparently there is still some weird effect in coverage checks, due to some randomness in the tests.
I think we should investigate this and find out which test(s) are causing this.

PR #620, opened 2015-10-23
Ros bayer (cleaned)

lornat75:
implemented automatic debayering when reading a bayer channel from ROS; when reading from a ImageOf<PixelMono> no debayering is performed leaving freedom to the user to do what he prefers. If the channel is read from a color image it performs debayering. Current implementations are very basic and do nearest neighbor interpolation. Implemented only debayer from 8 bit GRBG to RGB, RGBA, BGR, BGRA. Better functions could be used, for example by taking from the Bayer Carrier.


lornat75:
Looks like the compilation failures are not related to this PR, I would therefore merge the PR, @drdanz ?

drdanz:
@lornat75 I restarted the failed builds, let's wait a few more minutes just to be sure, but I'm ok with merging

PR #655, opened 2015-12-09
new yarp interface for laser scanner devices

randaz81:


randaz81:
wrong merge. will be deleted.

PR #662, opened 2016-01-05
MonitorObject.h is now installed

francesco-romano:
Fixed a documentation issue on MonitorObject.h
**Rewrote header guard macro**

I put in bold the reason why I opened a PR.
In particular, as this [link](http://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier) states, macro starting with a single or double underscore are reserved.
I modified only one for now, but I think we should change also the others.

@drdanz @lornat75 @traversaro

### DIFF
--- a/src/libYARP_OS/include/yarp/os/MonitorObject.h
+++ b/src/libYARP_OS/include/yarp/os/MonitorObject.h
@@ -56,7 +56,7 @@ public:
    
     /**
      * This will be called when one of the peer connections to the same import port receives data
-     * \Note this is available only if the portmonitor object attached to the input port
+     * @note this is available only if the portmonitor object attached to the input port
      */
     virtual void trig(void) { }
 
@@ -77,7 +77,7 @@ public:
      *
      * @param thing An instance of yarp::os::Thing object which can be used
      *        to typecast the data to the correct type. 
-     * @ return An instance of modified data in form of Thing
+     * @return An instance of modified data in form of Thing
      */
     virtual yarp::os::Things& update(yarp::os::Things& thing) { return thing; }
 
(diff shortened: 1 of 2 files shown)


drdanz:
Looks good, thanks.

Travis build is broken because I broke it :dizzy_face: , but should be fixed now... Can you please rebase over the latest master?

About the header guard, :+1: from me, even though I'd like to decide some nicer convention, ``YARP2_XXX`` looks something that could be defined for some other reason... perhaps something like ``yarp_os_MonitorObject_h``? @lornat75, what do you think?

Anyway, no need to block this pr for this reason... I'll merge as soon as it is rebased and travis build passes

lornat75:
@francesco-romano any chance you volunteer to change all guards in yarp :) ? I thin k it would be enough to remove the double underscores as you did for MonitorObject.h.

traversaro:
It seems to be a regular expression job... 
regular_expressions

francesco-romano:
I removed the change on the header guard because I address it in https://github.com/robotology/yarp/pull/663

@drdanz I prefer to define macro all capital. Anyway the change I did for now is simply to remove the underscores. If we want to change completely the style we have to do it manually (there are a lot of styles in yarp..)

src/libYARP_OS/include/yarp/os/MonitorObject.h
```diff
@@ -88,10 +88,10 @@ class YARP_OS_API yarp::os::MonitorObject
      *
      * @param thing An instance of yarp::os::Thing object which can be used
      *        to typecast the data to the correct type.
-     * @ return An instance of modified data in form of Thing
+     * @return An instance of modified data in form of Thing
      */
     virtual yarp::os::Things& updateReply(yarp::os::Things& thing) { return thing; }
 };
 
-#endif //_YARP2_MONITOROBJECT_
```
drdanz: This change no longer belongs to this patch...

drdanz:
Merged, thanks

PR #676, opened 2016-02-10
Fix cmake errors with CMake 3.5.0-rc1 on Windows

lornat75:
Apparently CMake is getting more strict raising errors in presence of empty parameters. This PR fixes some of these problems in an apparently harmless way. But please check it @drdanz.


drdanz:
About the regex replace fix, can you try reverting and quote `${_YI_DESTINATION}` instead? Regular expressions make sense on an empty string.

```diff
-    string(REGEX REPLACE "^${CMAKE_INSTALL_PREFIX}/" "" _YI_DESTINATION_RELATIVE ${_YI_DESTINATION})
+    string(REGEX REPLACE "^${CMAKE_INSTALL_PREFIX}/" "" _YI_DESTINATION_RELATIVE "${_YI_DESTINATION}"```)
```


The other commit should be
```diff
-if (NOT ${CMAKE_INSTALL_RPATH} STREQUAL "")
+if (NOT "${CMAKE_INSTALL_RPATH}" STREQUAL "")
```
or it might cause weird errors on some situations. Anyway there is something weird happening here, IIRC it should not be possible to get that far in AddInstallRPATHSupport on windows, since RPATH does not make sense on that platform. I'd like to investigate further this... I'm pretty sure we excluded this check since it should always be ensured that at lease one value is set in the list...

drdanz:
I made some tests and all these issues are caused by the same regression in CMake (see #677), so please don't merge (yet), I believe it will be fixed before the final 3.5 release

drdanz:
The patch fixing this issue already merged in CMake repository, it will be in CMake 3.5.0-rc3.
I'm Closing this, since it's now useless

lornat75:
ok, thanks for checking.

PR #687, opened 2016-02-17
Move robotInterface into YARP from icub-main

drdanz:
Non compatible changes compared to the original `robotInterface` in `icub-main`:

* `robotInterface` was renamed `yarprobotinterface` to follow the convention with yarp executables
* The RPC port opened is now /<robotname>/yarprobotinterface
* The default `.ini` file name is now `yarprobotinterface.ini`
* The DTD line for XML files is now
```xml
<!DOCTYPE [...] PUBLIC "-//YARP//DTD yarprobotinterface 1.0//EN" "http://www.yarp.it/DTD/yarprobotinterfaceV1.0.dtd">
```


gsaponaro:
Regarding this point:
>The RPC port opened is now //yarprobotinterface

do you mean ```/<robot>/yarprobotinterface```? It was changed to that style a few months ago: https://github.com/robotology/icub-main/commit/89d9bd26fd3ba9f561cbdabe9389953080c162b7, https://github.com/robotology/icub-main/commit/76cb3ad0d59f86c38edd578f64e1a4a167330025

Come to think of it, since you are renaming that RPC port anyway and we will re-do the XML scripts that interact with it, you could as well give the port an even clearer name, e.g., to ```/<robot>/yarprobotinterface/rpc``` or ```/<robot>/yarprobotinterface/rpc:i```, in line with other modules.

drdanz:
@gsaponaro yes, sorry, it was supposed to  be `/<robotname>/yarprobotinterface` but I did't quote it and`<robotname>` was probably interpreted as an html tag by github :smiley:
I updated the description above.

I'm not sure if there is a policy about the RPC port naming, @lornat75 @randaz81 what do you think?

randaz81:
afaik, rpc ports should end with `/rpc` (and not `/rpc:i`) because they are actually i/o ports

barbalberto:
Usually the suffix ```/rpc``` is added to a base name. The use of ```/rpc:i``` is not good since rpc is bi-directional, newer warpper/client are using this idea. ```:i``` and ```:o``` should be used for streaming port only.

drdanz:
RPC ports are actually bi-directional, but there are *server* and *client* ports, therefore I believe that `/rpc:i` is a tentative to represent a *server* port, while `/rpc:o` would represent a *client* port.

To be honest, now that I think about it, I really don't like the `/rpc` convention that adds an extra level to the port name and makes the name longer. I'd rather use something like

* `:i` for streaming input ports
* `:o` for streaming output ports
* `:s` for RPC server ports
* `:c` for RPC client ports

drdanz:
Merging, the builds failing on travis are failing also on master.

PR #688, opened 2016-02-17
build the portmonitor DLL plugin even if lua is not installed

apaikan:
@drdanz would you please check? :p

### DIFF
--- a/src/carriers/portmonitor_carrier/MonitorBinding.cpp
+++ b/src/carriers/portmonitor_carrier/MonitorBinding.cpp
@@ -8,10 +8,14 @@
  */
 
 #include "MonitorBinding.h"
-#include "MonitorLua.h"
 #include "MonitorSharedLib.h"
 
+#ifdef ENABLED_PORTMONITOR_LUA
+    #include "MonitorLua.h"
+#endif
+
 #include <yarp/os/ConstString.h>
+#include <yarp/os/LogStream.h>
 
 using namespace yarp::os;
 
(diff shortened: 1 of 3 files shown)


src/carriers/portmonitor_carrier/MonitorBinding.cpp
```diff
@@ -25,8 +29,14 @@ MonitorBinding::~MonitorBinding()
 
 MonitorBinding* MonitorBinding::create(const char* script_type)
 {
-    if(ConstString(script_type) == "lua")
+    if(ConstString(script_type) == "lua") {
+#ifdef ENABLED_PORTMONITOR_LUA
         return new MonitorLua();
+#else
+        yError()<<"Lua portmonitor plugin is not enabled!";
```
drdanz: Perhaps this error should be a little more specific, as "Lua portmonitor plugin is not enabled!" is a bit cryptic... something like "Cannot load LUA plugin. YARP portmonitor was built without LUA support".

src/carriers/portmonitor_carrier/lua/MonitorLua.cpp
```diff
@@ -15,7 +15,6 @@
 using namespace yarp::os;
 using namespace std;
 
-
```
drdanz: This change is useless ;)
Since this is the only change to this file I suggest to remove it, to keep the patch clean.

drdanz:
Besides these 2 comments, it looks good to me :+1:

drdanz:
Merged, thanks.

Issue #683, opened 2016-02-15
openni2 module not taking noMirror parameter into account 


jgvictores:
* Not working: modules/openni2 -> noMirror parameter -> mirrorON variable -X does not set the mirror parameter on the device throughout all https://github.com/robotology/yarp/blob/d7f67c8bfc52c48a545420e02ffdd8357b685069/src/modules/openni2/OpenNI2SkeletonTracker.cpp
* Solution (tentative): it seems that the OpenNI2 API allows calling setMirroringEnabled() for the RGB and the depth channels.

* Example of working code on a similar module: modules/kinect/OpenNI -> noMirror parameter -> mirrorON variable -> finally sets the mirror parameter of the device using the OpenNI API: https://github.com/robotology/yarp/blob/d7f67c8bfc52c48a545420e02ffdd8357b685069/src/modules/kinect/OpenNI/KinectYarpDeviceServerLib/KinectSkeletonTracker.cpp#L107
* Note: the motivation for this parameter was described for modules/kinect/OpenNI; e.g. robot teleoperation vs PbD.

kt10aan:
Thanks for the catch!!

I fixed it in my YARP fork (care had to to be taken for mirroring to be set on/off only when playback is off):
https://github.com/kt10aan/yarp/commit/c6a66357fd732b0cff71caa3cdee4336dfe32852

I will merge it with ``robotology/yarp`` after I have finished testing it today :)

kt10aan:
@barbalberto, you commented on a commit in ``kt10aan/yarp`` that was later squashed. 
In any case, @barbalberto commented that setting mirroring on is not guarantee that mirroring will work and he brought as an example that mirroring for RGB mode will be always on for the Asus Xtion.

Actually, setting the mirroring On/Off for the RGB mode is a bit pointless, since its a, well, RGB camera! The point of the mirroring option is for the depth stream to be mirrored.

As for if the sensors support mirroring for the depthStream, Kinect does support it and the Asus Xtion series does support it either. So I do not think there is any concern about handling the case where mirroring is not supported. 

With the other options, if the user sets a not supported hardware option, the driver proceeds to open as usual, but prints out an informative message about the requested feature being turned off.

Come to think of it, I should add a message for turning mirroring on too!

barbalberto:
I do believe it matters because the device will stream both rgb and depth images but they will be specular and therefore they cannot be used together for example to generate a point cloud from them or to match the color for each pixel.

[...]

PR #689, opened 2016-02-21
[openni2] setMirroringEnabled on depth and rgb; yError if status not ok

jgvictores:
[modules/openni2] setMirroringEnabled on depth and rgb; yError if status not ok.
For more information, read/watch issue #683 .


kt10aan:
Closed. See discussion on issue #683

PR #699, opened 2016-02-29
[Test] libYARP_OS

claudiofantacci:
Added a test for the _SystemInfo_ class (has 0% coverage according to coveralls.io) of libYARP_OS.
Corrected a typo in the SystemInfo.h doc.

### DIFF
--- a/src/libYARP_OS/include/yarp/os/SystemInfo.h
+++ b/src/libYARP_OS/include/yarp/os/SystemInfo.h
@@ -64,7 +64,7 @@ public:
 
 
     /**
-     * @brief The LoadInfo stuct holda the current cpu load information
+     * @brief The LoadInfo stuct holds the current cpu load information
      */
     typedef struct LoadInfo {
         double cpuLoad1;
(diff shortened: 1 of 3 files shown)


drdanz:
@claudiofantacci "Merge branch 'master' of https://github.com/robotology/yarp" is a commit that is generated automatically when you `pull` without rebasing and master received some updates. This kind of commits makes the history very messy.

Please rebase over the latest master and squash the two "good" commits into one. If you don't know how to do it just come and ask ;)

alecive:
@drdanz I don't know if this is unsafe in some cases (I don't think so), but @claudiofantacci you might find useful to configure your git client to automatically rebase when pulling. On linux, you have to run the following command:

```
git config --global pull.rebase true
```

drdanz:
@alecive I don't like "magic", I like to know what I'm doing... anyway it might be useful for someone.

@claudiofantacci Thanks, I'll merge as soon as the tests finish running.

claudiofantacci:
@alecive thanks for your suggestion! @drdanz told me that it would be safe to do that, but I prefer not to "force" rebase for every pull and instead to clearly understand when to use it not :-)

PR #700, opened 2016-03-02
attaching portmonitor plug-ins directly to the port using Yarp admin commands

apaikan:
**This pull-request allows attaching directly a pmonitor plug-in to a Yarp port.**

Motivation
-------------
 Up to now, a portmonitor plug-in could be attach to a specific connection either in the sender or receiver side. This could be useful if the user is interested in filtering/modifying data only within a specific connection. However, if the result of a plug-in is needed to be used by two different modules, the plug-in must be replicated twice for each connection. Thus it would be useful to have the plug-in attached directly to a port which process the outgoing data from the port before passing it to the connection's carriers. 

Usage
--------
Plug-ins (written in C++ or Lua) can be directly attached to a port using the Yarp admin port's commands. See the example below: 

|          sender         |      receiver 1      |      receiver 2       |
| ---------------------- | ----------------------| ----------------------|
| `$ yarp write /w`   |  `$ yarp read /r1` |  `$ yarp read /r2` |

```
$ yarp connect /w /r1
$ yarp connect /w /r2
```
```
$ yarp admin rpc /w
>> atch out "(type lua) (context monitors) (file bot_modifier.lua)"
Response: [ok]
>>
```
Now if you write something in the sender terminal, the corresponding message is modified by the `bot_modifier.lua` plugin and is handed to bother receivers. 

To remove the plug-in, in the yarp admin terminal simply type: 
```
$ yarp admin rpc /w
>> dtch out
```
Available commands: 
```
[atch] [out] $prop      # attach a portmonitor plug-in to the port's output
[atch] [in]  $prop      # attach a portmonitor plug-in to the port's input
[dtch] [out]            # detach portmonitor plug-in from the port's output
[dtch] [in]             # detach portmonitor plug-in from the port's input
```
Q&A
------

- What are the dependencies and what if the portmonitor carrier is not enabled in YARP? 
 Well. The implementation leverages `yarp::os::Carriers::chooseCarrier` method to look for the portmonitor plug-in at the run time. Nothing for the portmonitor's implementation is directly included in the Yarp PortCore. If the portmonitor carrier is not found, an informative message will be returned. 

- Can I attach a plug-in to the input port? 
 Yes.  Since Yarp ports are bidirectional (can be simultaneously read and write), you need to specify that in which  direction the plug-in should modify the data. To attach a plug-in to an input port, you need to send the `atch in <parameter>` admin command to the corresponding port. 

- What happen if i have a plug-in at the port and one on the connection? 
 Nothing strange! the plug-ins are chained in this case. The data will be processed by both of them. The order is as bellow: 

```
Port.write() -> [port out plug-in] -> [carrier sender plug-in]    ------->
Port.read() <- [port in plug-in]  <-  [carrier receiver plug-in]  <-------                                                                                                      
```

TODO
--------
- Adding a command to the yarp companions. e.g., `yarp attach <param>` and `yarp detach <param>`. We need to agree on the syntax!

- Adding portmonitor to the import port is disabled in this commit since it needs further reliability tests.   

- This commit should be eventually merged  with the `portmonitor_plugins` branch which allows loading the portmonitor C++  (dll) plug-ins using YARP plug-in manager.

### DIFF
--- a/src/carriers/portmonitor_carrier/PortMonitor.cpp
+++ b/src/carriers/portmonitor_carrier/PortMonitor.cpp
@@ -30,7 +30,17 @@ bool PortMonitor::configure(yarp::os::ConnectionState& proto)
 
     Property options;
     options.fromString(proto.getSenderSpecifier().c_str());
+    options.put("source", sourceName);
+    options.put("destination", portName);
+    options.put("sender_side",
+             (proto.getContactable()->getName() == sourceName) ? 1 : 0);
+    options.put("receiver_side",
+             (proto.getContactable()->getName() == portName) ? 1 : 0);
+    options.put("carrier", proto.getRoute().getCarrierName());
+    return configureFromProperty(options);
+}
 
+bool PortMonitor::configureFromProperty(yarp::os::Property& options) {
     if(binder) delete binder;
     binder = NULL;
 
@@ -68,18 +78,17 @@ bool PortMonitor::configure(yarp::os::ConnectionState& proto)
     info.clear();
     info.put("filename", strFile);
     info.put("type", script);
-    info.put("sender_side",
-             (proto.getContactable()->getName() == sourceName));
-    info.put("receiver_side",
-             (proto.getContactable()->getName() == portName));
-    info.put("source", sourceName);
-    info.put("destination", portName);
-    info.put("carrier", proto.getRoute().getCarrierName());
+    info.put("source", options.find("source").asString());
+    info.put("destination", options.find("destination").asString());
+    info.put("sender_side",  options.find("sender_side").asInt());
+    info.put("receiver_side",options.find("receiver_side").asInt());
+    info.put("carrier", options.find("carrier").asString());
 
     PortMonitor::lock();
     bReady =  binder->load(info);
     PortMonitor::unlock();
     return bReady;
+    return false;
 }
 
 void PortMonitor::setCarrierParams(const yarp::os::Property& params)
@@ -152,10 +161,11 @@ bool PortMonitor::acceptIncomingData(yarp::os::ConnectionReader& reader)
         }
     }
 
-    getPeers().lock();
-    yAssert(group);
-    result = group->acceptIncomingData(this);
-    getPeers().unlock();
+    if(group!=NULL) {
+        getPeers().lock();
+        result = group->acceptIncomingData(this);
+        getPeers().unlock();
+    }
     return result;
 }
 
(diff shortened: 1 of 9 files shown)


pattacini:
Great PR :+1:

drdanz:
It's a great addition and the code looks great, but I have a couple of questions, perhaps pedantic, but I just want to be sure of the behaviour before this is merged.

1. What happens in your example if I connect another port after attaching the port monitor? Does the new port get the modifier?

2. Can you attach the portmonitor to a disconnected port and connect a port later?

3. What happens when you disconnect all the ports? What happens if you reconnect the ports later?

3. What does getOutputCount() return when a portmonitor is attached and no ports are connected (because they were never connected, or because they were disconnected later)?

4. Can you attach *any* `ModifierCarrier` or just the portmonitor? The "core" part does not contain any reference to portmonitor, just the selection part, so perhaps it wouldn't be hard to extend to any `ModifierCarrier`?

5. (Actually this is a comment) The portmonitor is becoming an important part of YARP, therefore, unless we change it to work with any `ModifierCarrier`, perhaps it should not be treated as an extra carrier, and loaded as a plugin but merged in YARP_OS like the `tcp`, `udp`, etc.

6. (Actually another comment) It would be **really** nice to have **unit tests** for this, testing the behaviours of the previous questions.

apaikan:
Thanks @drdanz . Here are the clarifications: 

1- *What happens in your example if I connect another port after attaching the port monitor? Does the new port get the modifier?*
 Yes! Since the plugin is connected to the port instead of a connection, all the connection from the port got the modifier

2- *Can you attach the portmonitor to a disconnected port and connect a port later?* 
 Yes! The plugin is attached to the port itself! 

3- *What happens when you disconnect all the ports? What happens if you reconnect the ports later?* 
Basically nothing. The plugin remain attached to the port independent from the connections and their states. 

4- *What does getOutputCount() return when a portmonitor is attached and no ports are connected (because they were never connected, or because they were disconnected later)?*
 Attaching a plugin to a port does not create any `PortCoreUnit`! Thus it does not affect the `getOutputCount()`. The plugin applied before the data from `Port.write()` passed to any output unit (connection). 

5- *Can you attach any ModifierCarrier or just the portmonitor? The "core" part does not contain any reference to portmonitor, just the selection part, so perhaps it wouldn't be hard to extend to any ModifierCarrier?*
Correct! It can be easily extended for any other modifier! 

6- *(Actually this is a comment) The portmonitor is becoming an important part of YARP, therefore, unless we change it to work with any ModifierCarrier, perhaps it should not be treated as an extra carrier, and loaded as a plugin but merged in YARP_OS like the tcp, udp, etc*
Agreed! :) 

7- *(Actually another comment) It would be really nice to have unit tests for this, testing the behaviours of the previous questions.*

Again agreed! The reason that the unit test for portmonitor is not implemented yet is that the portmonitor plugin marked as optional (not enabled by default). Anyway the unit test could be made also optional.

PR #706, opened 2016-03-12
[bindings] added yarp::sig::Matrix in bindings

arocchi:
yarp::sig::Matrix was not included in bindings.
Now all bindings include also Matrix, and in particular the python bindings allow to access elements with the same semantic as cpp, e.g.:
```
import yarp

m = yarp.Matrix(3,3)
m[2,2] = 1.0
```


drdanz:
Thanks!

Issue #690, opened 2016-02-23
OpenNI2 device and possible migration to RGBDWrapper


kt10aan:
Here is @barbalberto's post from issue #683 :

Ok, I see there is there is good will for improvement and I like it :+1: so here is the 'long answer', I hope you guys can read it all and provide feedback.

My interest in the OpenNI device came from issue #643 I'm in charge to work on.

I examined all drivers related to kinect-like devices (in IIT we need one in a new robot) and I recently implemented the same in the Gazebo simulator (https://github.com/robotology/gazebo-yarp-plugins/issues/100) so we can work on simulator and then go on the real robot.

What I found out is that there is no real 'standard' in yarp for this kind of devices, and the situation is even worse for pointclouds. So I'd like to make some order in this.

I found out there are 4 kinect-like yarp drivers but all of them lack of some key features or have some dis-advantage.
Please think of the following points as ways to improve the yarp-ish software related to this awesome depth cameras devices.
My observations are:
**A)** Driver in yarp are meant to be splitted into 2 different part (modules/class/object):
- Proper driver: this part reads the data from hardware (no yarp port!)
- Wrapper: this part uses an interface to read data from the former one and it send the data on the network through yarp ports.
  The reason is to re-use the same wrapper with different hw-devices, i.e. we will have the same ports and same data, same rpc calls regardless of using kinect, xtion or other devices (ideally also a streovision sistem with 2 cameras and disparity) through abstraction layer. Obviously abstraction layer has to be comprehensive to cover small differencies that can arise between different hw producers.

This is what is done for motioncontrol, so we use the same controlBoardWrapper2 over CAN devices, ETH devices, simulators (iCub, gazebo, robotran), ethercat device and so on. This allows code to work on different robots by abstracting hw level.  
All kinect-like yarp drivers I saw, do the 2 things (driver and wrappper) together in a single module and cannot be separated. This means, for example, there was no 'pure wrapper' I could use with the simulators, therefore I had to write a yet another one.

**B)** As follow from lack of an unified wrapper, each device opens it own ports and send data using is own format: 
- OpenNI2 uses one port for `rgb` and one port for `imageOf<PixelMono16>`  which is unsigned int
- robotology/kinect-wrapper uses one port for `rgb` and one port for `imageOf<PixelMono16>`  which is unsigned
- yarp/modules/kinect/OpenNi(1) uses  `ImageOf<PixelInt>` which is `int32` with _sign_
[...]

PR #710, opened 2016-03-16
Open ni2 device clean wip

barbalberto:
This PR changes the OpenNI2 device CMake file in such a way that if NiTE2 library is not found, then the compilation of the skeleton extraction is skipped.


drdanz:
LGTM, but perhaps @kt10aan or @jgvictores want to have at look at it before merging?

kt10aan:
@drdanz, I have tested  the PR. See https://github.com/robotology/yarp/issues/690

drdanz:
@kt10aan cool thanks! I see in the same issue that also @jgvictores tested it, therefore I'm merging this, if there is some further issue we can fix it later.

drdanz:
@barbalberto merged, thanks!

PR #716, opened 2016-03-30
Cleanup of tutorials

gabrielenava:
- in yarp/example/os/image_process_module.cpp the input of module.runModule() has been modified according to what is required by the RFmodule.cpp (which substitutes the deprecated code Module.cpp) 

- in yarp/doc/yarp_cmake_hello.dox it has been added the suggestion to create a build directory inside the source folder and launch cmake from it


pattacini:
Seems fine, merging.
Thanks @gabrielenava :+1:

PR #866, opened 2016-08-12
Fixes #621: Add check for NULL to fix potential segfault

### LINKED ISSUE #621: OS::PortTest failing in CI travis?
I noticed that in some PR travis CI fail with the follow error:

29/63 Test #29: OS::PortTest .......................***Exception: SegFault  3.77 sec
test 30
      Start 30: OS::PortReaderBufferTest

This problem seems to be unrelated with the code changed in PR and is sporadic, maybe there is a race condition.

@drdanz

Tobias-Fischer:
This fixes #621. The addition of `op!=NULL` is the actual fix; moving `if(done) { ... }` up in the other protected area is just clean-up (and does not change the logic). 

I've written a small comment in the code, as just moving the `if(replied ...)` up in the other protected area is not enough. The call to `closeBasic` might set `op=NULL`, and thus there is another check required whether `op!=NULL`.

I had this running for ~16 hours without crash.

Would be great if you could merge @drdanz.
<a href='#crh-start'></a><a href='#crh-data-%7B%7D'></a>
<a href='https://www.codereviewhub.com/'></a>
<a href='https://www.codereviewhub.com/robotology/yarp/pull/866?approve=1'></a>&nbsp;<a href='https://github.com/robotology/yarp/pull/866'></a>
<a href='#crh-end'></a>

### DIFF
--- a/src/libYARP_OS/src/PortCoreOutputUnit.cpp
+++ b/src/libYARP_OS/src/PortCoreOutputUnit.cpp
@@ -322,15 +322,16 @@ bool PortCoreOutputUnit::sendHelper() {
         if (buf.dropRequested()) {
             done = true;
         }
-    }
-    if (done) {
-        closeBasic();
-        finished = true;
-        closing = true;
-        setDoomed();
+        if (done) {
+            closeBasic();
+            closing = true;
+            finished = true;
+            setDoomed();
+        }
     }
 
-    if(replied && op->getSender().modifiesReply()) {
+    // Another check for op!=NULL is required as closeBasic() might set op=NULL
+    if(op!=NULL && replied && op->getSender().modifiesReply()) {
             cachedReader = &op->getSender().modifyReply(*cachedReader);
     }
 


drdanz:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/robotology/yarp/pull/866'>Refresh</a>.<a href='#crh-update-none'></a>

drdanz:
Awesome! Merged, thanks!

Tobias-Fischer:
:+1:

drdanz:
> moving if(done) { ... } up in the other protected area is just clean-up (and does not change the logic).

@Tobias-Fischer are you sure that this change does not change the logic?
If I understand it correctly, now `closeBasic()` and `setDoomed()` are no longer called when `op==NULL`, but I don't know if this is supposed to happen. What do you think?

Tobias-Fischer:
@drdanz I'm pretty sure the logic doesn't change. `done` is a local variable defaulting to `false`, and is only ever set to `true` if `op!=NULL`. Thus moving the `closeBasic(); setDoomed();` block up in the `op!=NULL` block doesn't change the logic (I think).

drdanz:
@Tobias-Fischer you are right, sorry! :+1:

drdanz:
@Tobias-Fischer For readability, I think I will move the `done` definition inside the `if` as well...

Tobias-Fischer:
@drdanz Yes this makes sense

PR #894, opened 2016-09-01
Feat/run module threaded

claudiofantacci:
RFModule updateModule can now be run on a separate thread without blocking the thread calling runModule().

In particular, the following functions have been **_added_**:
 - `runModuleThreaded()`: it is the non-blocking equivalent of runModule(). A new thread within RFMoudle is created which itself run the runModule() method of the RFModule. The calling thread is not blocked within RFModule and return as soon as the handling thread is created and started.
 - `runModuleThreaded(ResourceFinder &rf)`: equivalent to runModuleThreaded() but also invoke configure(ResourceFinder &rf).
 - `joinModule(double seconds = -1)`: this new method is equivalent to the standard join(double secnods) method of Thread. It is used to check whether the handling thread in charge of calling runModule() has properly been terminated or not. This function is particularly useful to wait until possible deallocations are completed within the RFModule close() method.

The following functions have been **_modified_**:
 - `stopModule(bool wait = false)`: the original behavior is not changed, but the unimplemented wait feature has been used to add thread join capability to the method. If runModuleThreaded() is invoked, stopModule(true) will jointly ask the RFModule to stop and wait for the thread to return using a join() call. On the other hand, if stopModule(false) is used, the user will have to use joinModule(double seconds) to check proper deallocation of the RFModule.

**_Minor modification_** to existing functions:
 - `runModule()` and `runModuleThreaded()` will also check whether a runModuleThreaded() was invoked before or not to guarantee a singleton invocation of the runModule().

Hunk of example code:
```c
RFModuleDerivedClass ex;
ResourceFinder rf;
// Do something with rf

ex.runModuleThreaded(rf);

while (!ex.isStopping()) {
    // Do something
}

ex.joinModule();
```

### DIFF
--- a/src/libYARP_OS/include/yarp/os/RFModule.h
+++ b/src/libYARP_OS/include/yarp/os/RFModule.h
@@ -90,6 +90,40 @@ public:
     virtual int runModule(yarp::os::ResourceFinder &rf);
 
     /**
+     * Calls updateModule() on a separate thread until that returns false.
+     *
+     * Make sure you first configure your module by calling the configure()
+     * function. updateModule() is then called every getPeriod()
+     * seconds.  During execution of updateModule() the following methods may be
+     * executed asynchronously:
+     *  - respond(): this is called if there is input from the standard input or
+     *               a message from an input port connected to the module via
+     *               attach().
+     *  - interruptModule(): this method is called by the handlers of  the
+     *               following signals: SIGINT, SIGTERM and SIGBREAK (WIN32).
+     *               Interrupt is a good place to execute code that unblocks
+     *               pending reads (i.e. blocking reads on port).
+     *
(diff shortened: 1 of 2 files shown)


drdanz:
@claudiofantacci The patch looks good to me, thanks! It would be really useful if you could add some test on the `ModuleTest`, i.e. at least one using `stopModule(true)` and one using `stopModule(false)`+`joinModule()`

drdanz:
@claudiofantacci I'll merge it anyway since you are not around, but I'll bug you when you come back for the unit tests :smiling_imp:

claudiofantacci:
I was just about to reply.
I will start working on the ModuleTest tomorrow (hopefully) and of course will resume after coming back form US.
No worries "bugging" me for it 😇

PR #917, opened 2016-09-28
getEnvelope/setEnvelope documentation improvement

gsaponaro:
After discussion with @drdanz
See also http://wiki.icub.org/wiki/Timestamps_in_YARP (2006)


drdanz:
@gsaponaro Merged, thanks!

PR #926, opened 2016-10-07
Fix SIGPIPE errors when running tests with gdb

drdanz:
* **OS/PortCore: do not interrupt input unit in closeMain**

  When an input port is deleted, all the ports connected to this port should be disconnected first. Therefore a message is sent to these ports, and these perform the actual disconnection. The real disconnection is executed after sending a final message on the socket.

  The interrupt() call removed by this patch closes the socket internally, therefore when the final message is sent by the output port, the output socket is broken, hence the SIGPIPE.

  Removing the interrupt here does not seem to have side effects, since the socket is closed later, when the message arrives.

* **NetworkTest: Interrupt and close the port when the SlowResponder thread is stopped**

  This fixes the SIGPIPE crash with gdb

### DIFF
--- a/doc/release/v2_3_66_1.md
+++ b/doc/release/v2_3_66_1.md
@@ -48,6 +48,7 @@ Bug Fixes
   other platforms (#633)
 * `yarp plugin` command now works with `SKIP_ACE` enabled.
 * Fixed UDP and MCAST on macOS (#637)
+* Fixed SIGPIPE when closing input port before output ports.
 
 
 ### YARP_DEV
--- a/src/libYARP_OS/src/PortCore.cpp
+++ b/src/libYARP_OS/src/PortCore.cpp
@@ -373,7 +373,6 @@ void PortCore::closeMain() {
             if (unit!=NULL) {
                 if (unit->isInput()) {
                     if (!unit->isDoomed()) {
-                        unit->interrupt();
                         Route r = unit->getRoute();
                         String s = r.getFromName();
                         if (s.length()>=1) {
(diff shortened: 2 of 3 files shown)


drdanz:
After a better look I'm no longer 100% sure whether the socket is closed for real or not, after a better look, I don't see any call to `stream.close_reader()` and `stream.close_writer()` besides the ones in `interrupt()`, and I don't see `interrupt()` called anywhere else. Please do not merge yet

drdanz:
`stream.close_reader()` and `stream.close_writer()` are definitely not called. On the other hand, `stream.close()` is called, and internally it calls  [`close_writer()`](https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/ace/SOCK_Stream.cpp#L33), therefore I assume that this will do everything that is required to close the socket. Therefore I believe that tcp/udp is fine.

Nonetheless I'm not sure what removing the `interrupt()` will cause to other protocols. Anyone has any clue?

drdanz:
I found these interesting lines in [ThreadImpl.cpp:55](https://github.com/robotology/yarp/blob/0e5a04386c2468a2bfeb2d17387f2a97ad31ef0a/src/libYARP_OS/src/ThreadImpl.cpp#L55) and [Network.cpp:611](https://github.com/robotology/yarp/blob/0e5a04386c2468a2bfeb2d17387f2a97ad31ef0a/src/libYARP_OS/src/Network.cpp#L611)

```c++
ACE_OS::signal(SIGPIPE, SIG_IGN);
```

According to this, `SIGPIPE` is ignored by all the threads. Commenting out these 2 lines causes the errors also without gdb. And I confirm that all the tests pass with this patch and these 2 lines commented.

drdanz:
Tested on iCub + red ball demo, works.

Issue #929, opened 2016-10-17
Ros topic interoperation instability when using UDP protocol


aerydna:
Disconnecting and re-connecting a yarp port that is publishing in a ros topic can cause the program to `exit()`.
I presume this to happens if the used protocol is the UDP because of the following:
If you publish on some topic, with Rviz you can subscribe to it, disconnect, re-connect and switch to UDP protocol by checking the "Unreliable" checkbox.
If you disconnect and re-connect while the checkbox is on somehow the yarp application will execute `carriers/xmlrpc_carrier/xmlrpc/XmlRpcValue::binaryFromXml()` that calls an `exit()` right after logging `"binaryFromXml disabled until license of base64.h determined\n"`.

drdanz:
@aerydna Closing, since I just merged #930 that *should* fix this issue. I cannot say for sure though, can you please test it again and reopen this issue if it is not fixed?

aerydna:
good! i'll test it as soon as possible..

PR #930, opened 2016-10-18
XmlRpc++: Restore methods using base64

drdanz:
base64.h was removed long time ago from the XmlRpc++ library used by the xmlrpc carrier, because the original license of the file was unknown at the time. The methods of the library using these base64 methods were commented out.

ROS probably in some cases send some data encoded with base64, therefore issue #929.

Since the license of the file is still unknown, a different version is imported here and used inside the XmlRpc++ library. This version is taken from [here](http://www.adp-gmbh.ch/cpp/common/base64.html) and the license is explicit (zlib license) and it is compatible with LGPL.

Also the XmlRpc++ library was moved to the extern folder and each patch was extracted from the history, so that it is clear where the code comes from and what happened to it.


extern/xmlrpcpp/xmlrpcpp/src/base64.h
```diff
@@ -1,4 +1,37 @@
...
+      claim that you wrote the original source code. If you use this source code
+      in a product, an acknowledgment in the product documentation would be
+      appreciated but is not required.
+
+   2. Altered source versions must be plainly marked as such, and must not be
+      misrepresented as being the original source code.
+
+   3. This notice may not be removed or altered from any source distribution.
+
+   Ren? Nyffenegger rene.nyffenegger@adp-gmbh.ch
```
gsaponaro: minor thing: the '?' should be an 'é' here => René

drdanz: Thanks! Fixed.

drdanz:
`devel` is currently not building due to a stupid error that appeared a few days ago on macOS with c++11 disabled. Merging since it builds on every other platform

PR #933, opened 2016-10-19
Fix the bundle folder to be named "<target>.app.app"

drdanz:
After removing the GTK guis, also the code setting the `OUTPUT_NAME` property for gui targets was removed.
This code was setting the output name for the target `yarpfoo-qt` to `yarpfoo`.
This property was then copied to the bundle copy of the target, i.e. `yarpfoo.app` target had `OUTPUT_NAME` set to `yarpfoo`.
If `OUTPUT_NAME` is not set, it uses the name of the target as the output name, i.e. `yarpfoo.app`

CMake adds automatically `.app` to this name, therefore if `OUTPUT_NAME` is not set, `yarpfoo` copy target gerates a bundle `yarpfoo.app.app`

In this patch, if `OUTPUT_NAME` is not set, it is forced to the name of the original target.

Also the `TARGET_DEST` and `TARGET_ORIG` options are removed and replaced by the `TARGET` option

### DIFF
--- a/cmake/YarpOSXUtilities.cmake
+++ b/cmake/YarpOSXUtilities.cmake
@@ -99,6 +99,12 @@ function(YARP_OSX_DUPLICATE_AND_ADD_BUNDLE)
     # Enable bundle creation
     set_target_properties(${_target_dest} PROPERTIES MACOSX_BUNDLE ON)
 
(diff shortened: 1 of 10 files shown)


drdanz:
@francesco-romano Can you please test it?

PR #942, opened 2016-11-03
Fix segfault while closing yarp modules

barbalberto:
This issue fix the destruction of logForwarder causing segfault in closure of any yarp application.


barbalberto:
Rebased on current devel

barbalberto:
The error in travis is a fail on FrameTransformClientTest which is not related to this PR.
I'd merge.

PR #943, opened 2016-11-03
Fix RFModuleTest

claudiofantacci:
Fixes ModuleTest name in RFModuleTest and add new tests for the new threaded-mode functionalities.

PR #946, opened 2016-11-05
Fix thread return value when C++11 compilation is activated

claudiofantacci:
This is a dedicated PR that follows PR #943.

The 3 commits address 2 issues:

- Fix/Align the join() return value of YARP threads when C++11 compilation is activated. In particular, the return value is aligned with the one used by ACE and pthread threads (even though, as pointed out by @drdanz, `EINVAL` should be used as return value in place of -1, but the requirement is a value different from 0 and -1 is the same used by ACE.)
- Improve tests for YARP thread. The return values of join() calls are now tested.

### DIFF
--- a/doc/release/v2_3_66_2.md
+++ b/doc/release/v2_3_66_2.md
@@ -11,6 +11,7 @@ Important Changes
 
 Bug Fixes
 ---------
+* Fixed join() return values of threads when C++11 compilation is activated (macro PLATFORM_THREAD_JOIN).
 
 ### GUIs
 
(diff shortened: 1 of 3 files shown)


drdanz:
@claudiofantacci Thanks! One of the tests failed for some unrelated race condition, I just restarted it... I will merge this as soon as it finishes.

claudiofantacci:
@drdanz I noticed that as well, but I could not explain myself what happened under the hood.
Good to hear that, let's see what happen 😄

PR #958, opened 2016-11-15
Fix quaternion handling functions

drdanz:
See #910


randaz81:
This pull request has been deprecated by #959

PR #967, opened 2016-11-18
Setting modules' environment variables within application XML file

apaikan:
This pull request  (as requested by  :notes: @randaz81) allows to set the multiple environment variables for each single module within the application xml file. example below: 

```xml
    <module>
        <name>mymodule</name>
        <node>localhost</node>
        <prefix>/test</prefix>
        <environment>VAR1=value1;VAR2=value2</environment>
    </module>
```
the `mymodule` will receive the flowing variables in its local environment-variables set: 

```ini
...
YARP_PORT_PREFIX=/test
VAR1=value1
VAR2=value2
```

### DIFF
--- a/src/libYARP_OS/src/Run.cpp
+++ b/src/libYARP_OS/src/Run.cpp
@@ -9,6 +9,7 @@
 #include <yarp/os/Time.h>
 #include <yarp/os/impl/Logger.h>
 #include <yarp/os/impl/RunProcManager.h>
+#include <yarp/os/impl/SplitString.h>
 #include <yarp/os/SystemInfo.h>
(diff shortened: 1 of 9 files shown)


drdanz:
@apaikan It's not building on windows, I think that you might have to include `ACE_INCLUDE_DIRS` somewhere, `SplitString.h` seems to require it.

drdanz:
@apaikan FYI I just committed some changes to `SplitString`, it should now build even without `ACE_INCLUDE_DIRS`.

drdanz:
@apaikan A `YARP_OS_impl_API` was missing from `SplitString`, I just fixed it and rebased your branch (also I removed last commit that should be useless now). Let's see if it still fails...

PR #968, opened 2016-11-18
Remote cb fix get data

randaz81:
These very important commits from @barbalberto fix an insidious race condition causing segfault of a remoteControlBoard client. Currently under testing.

### DIFF
--- a/src/libYARP_dev/src/devices/RemoteControlBoard/RemoteControlBoard.cpp
+++ b/src/libYARP_dev/src/devices/RemoteControlBoard/RemoteControlBoard.cpp
@@ -1628,13 +1628,8 @@ public:
         else
         {
             extendedPortMutex.wait();
-            ret = extendedIntputStatePort.getLast(j, last_singleJoint, lastStamp, localArrivalTime);
+            ret = extendedIntputStatePort.getLastSingle(j, VOCAB_ENCODER, v, lastStamp, localArrivalTime);
             extendedPortMutex.post();
-
-            if(ret && last_singleJoint.jointPosition_isValid)
-                *v = last_singleJoint.jointPosition[0];
-            else
-                ret = false;
         }
         if (ret && Time::now()-localArrivalTime>TIMEOUT)
             ret=false;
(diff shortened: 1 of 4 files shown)


src/libYARP_dev/src/devices/RemoteControlBoard/RemoteControlBoard.cpp
```diff
@@ -2067,29 +2013,22 @@ class yarp::dev::RemoteControlBoard :
      * @param v pointer to storage for the return value
      * @return true/false, upon success/failure (you knew it, uh?)
      */
-    virtual bool getMotorEncoderTimed(int j, double *v, double *t) {
+    virtual bool getMotorEncoderTimed(int j, double *v, double *t)
+    {
         double localArrivalTime = 0.0;
-        bool ret = false;
+        bool ret(false);
```
drdanz: FYI This change has no effect, since copy initialization and direct initialization do exactly the same for builtin types.

src/libYARP_dev/src/devices/RemoteControlBoard/RemoteControlBoard.cpp
```diff
@@ -2105,13 +2044,13 @@ class yarp::dev::RemoteControlBoard :
      * @return true/false on success/failure. Failure means encoders have not been received
      * from the server or that they are not being streamed with the expected rate.
      */
-    virtual bool getMotorEncoders(double *encs) {
+    virtual bool getMotorEncoders(double *encs)
+    {
```
drdanz: Ok, having a better look at the file, there are already different styles, so feel free to disregard the other comments except for line 1619

drdanz:
@randaz81 @barbalberto Are you planning to merge this before the release?

barbalberto:
Yes, this is an important bugfix. I just want to run some more tests.
(about the alignement you are right, probably the IDE take initiative while saving the file, sorry. )

barbalberto:
Fixed a small issue with a value not initialized ans tested with iCub Sim.
It looks ok, I'd go merging in devel

PR #969, opened 2016-11-22
fixes: yarpview on windows was crashing on prop get command from admin rpc port

apaikan:
This pull request also fixes crashing yarpview on windows while using yarpviz!


drdanz:
@apaikan merged, thanks.
Please use white spaces instead of tabs next time (I will fix this).

PR #971, opened 2016-11-23
fix bug in setRemoteVariable

randaz81:
This fix allows to set remote variables which do not belong to the first set of joints. e.g. (0 1 2) (0 1 HERE 3 4)


randaz81:
Working. Tested on iCubGenova04.

PR #977, opened 2016-11-29
RGBDbugfix

aerydna:
-
-camInfo params for ros interoperation
-typo correction
-yarp::os::Node add "/" at the beginning of the name if not present.

### DIFF
--- a/src/devices/depthCamera/depthCameraDriver.cpp
+++ b/src/devices/depthCamera/depthCameraDriver.cpp
@@ -754,6 +754,15 @@ bool depthCameraDriver::setIntrinsic(Property& intrinsic, const intrinsicParams&
     }
 
     intrinsic.put("retificationMatrix", Value(mat.toString()));
+
+    intrinsic.put("distortionModel", "plumb_bob");
+    intrinsic.put("k1", values.distortionModel.k1);
+    intrinsic.put("k2", values.distortionModel.k2);
+    intrinsic.put("t1", values.distortionModel.t1);
+    intrinsic.put("t2", values.distortionModel.t2);
+    intrinsic.put("k3", values.distortionModel.k3);
+
+    intrinsic.put("stamp", yarp::os::Time::now());
     return true;
 }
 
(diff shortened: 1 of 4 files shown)


drdanz:
@aerydna please rebase the branch over the latest `master` (if it is a bugfix) or over the latest `devel` if it contains new features, there are several "Merge branch" commits and the "Prepare for next feature release (2.3.70)" commit that should not be there

aerydna:
ok Sorry

src/libYARP_OS/src/Node.cpp
```diff
@@ -549,13 +549,22 @@ Node::Node() :
 Node::Node(const ConstString& name) :
         mPriv(new Helper)
 {
+    ConstString _name;
+    if(name.at(0) != '/')
+    {
+        _name == "/"+name;
```
aerydna: you're right, small oversight..

barbalberto: Ok as a bugfux.
As a personal taste, I don't like so much when the the code changes user params on its will, and I think a constructor should never attempt operation which may fail, like opening port or check config params.

I propose to remove this option and substitute with (if not yet existing) an empty constructor and a init/config method which accepts the node name. This way the method can clearly fail and a better error handling can be achieved.

Can we chat a bit about this in the next IFI day?

src/libYARP_OS/src/Node.cpp
```diff
@@ -549,13 +549,22 @@ Node::Node() :
...
 {
+    ConstString _name;
+    if(name.at(0) != '/')
+    {
+        _name = "/"+name;
+    }
+    else
+    {
+        _name = name;
+    }
```
drdanz: Please, make a separate PR for this change, even if it is very small it is a change of behaviour in YARP_OS, and I don't want this to be hidden in a "RGBDbugfix" pull request...

drdanz:
Merged, thanks.

PR #984, opened 2016-12-01
no more hardcoded number of parts

aerydna:
no more hardcoded limit on the number of parts. only the check has been removed and everything seems to work.
however the code is messy at the moment and his behaviour is not obvious. @barbalberto had informed me that this gui has been converted to Qt from a previous gtk version so i presume that some things are survived phantoms of the previous version (and the main suspect is that limiting the number of part was one of them).


drdanz:
@randaz81 Can you please have a look at this?

randaz81:
@aerydna can you also remove the reprehensible ENA variable?

aerydna:
ENA and lots of other useless variable removed.
added close on CTRL+C

drdanz:
@randaz81 When you are satisfied with the changes, please approve it.

PR #985, opened 2016-12-01
Fix fake motion control open - issue #983

barbalberto:
Fix issue #983. The fakemotioncontrol was not opening correctly if 'Encoder' parameter was used in the config file


drdanz:
Rebased and merged. Thanks

PR #989, opened 2016-12-01
Fix for issue #975 - Adding warning when plugin has a wrong library parameter

bhigy:


src/libYARP_dev/src/Drivers.cpp
```diff
@@ -87,7 +87,10 @@ class DriversHelper : public YarpPluginSelector {
             settings.readFromSearchable(prop,name);
             settings.open(lib);
             ConstString location = lib.getName().c_str();
-            if (location=="") continue;
+            if (location=="") {
```
drdanz: Is location set empty after the library failed loading? Or does this work only for empty location? Do you mind adding a comment here to explain?

bhigy: I tried by putting a wrong library name and it enters the if. So yes, wrong library name gives an empty location. I'll add a comment.

drdanz: Perfect! Merging this then!

PR #990, opened 2016-12-01
Added FindOpenNI2 

giuliavezzani:
Trying to fix [#110 ycm](https://github.com/robotology/ycm/issues/110).


drdanz:
Thanks @giuliavezzani 
Unfortunately probably this won't work for windows, the part inside the `if(WIN32)` should be kept, porting it to the new variables defined.

Also the right place for this PR is YCM (devel branch), we can import it in YARP after it is merged in YCM.
I'm closing this, please fix in in YCM.

PR #995, opened 2016-12-02
Fix for issue #832 and further related improvements

massimoregoli:
I made the following changes:

- With respect to the issue #832, now a warning is printed if parenthesis are not balanced.
- I noticed that, in a parameter's value, when parenthesis are not balanced (or in general when the syntax of the value is not respected), also other parameters may be affected (that is they might not be stored in the `Property` object). This is due to the fact that before creating the `Property `object an intermediate string representation is used. I got rid of this (apparently useless) intermediate step, so that when the syntax of a value is not respected, only the related parameter will be corrupted.
-  In `PolyDriver.cpp` I added a warning when in a `Property` object there are two 'device' parameters nested one in the other (one is part of the value of the other). This situation was currently managed applying a workaround, that is now ignored when `YARP_NO_DEPRECATE` is enabled.
- In `Param.cpp` the `toProperty()` now parses the value before putting it in the `Property `object.

### DIFF
--- a/src/libYARP_dev/src/PolyDriver.cpp
+++ b/src/libYARP_dev/src/PolyDriver.cpp
@@ -214,13 +214,17 @@ bool PolyDriver::coreOpen(yarp::os::Searchable& prop) {
     if (prop.check("device",part)) {
         str = part->toString().c_str();
     }
+
+#ifndef YARP_NO_DEPRECATE // since yarp 2.3.70
     Bottle bot(str.c_str());
     if (bot.size()>1) {
-        // this wasn't a device name, but some codes -- rearrange
+		yWarning("Passing 'device' parameter with a list of parameters as value is deprecated. This might be an internal bug. If you didn't do it please report an issue at https://github.com/robotology/yarp");
+		// this wasn't a device name, but some codes -- rearrange
         p.fromString(str.c_str());
         str = p.find("device").asString().c_str();
         config = &p;
     }
+#endif
 
     DeviceDriver *driver = NULL;
 
--- a/src/libYARP_dev/src/PolyDriver.cpp
+++ b/src/libYARP_dev/src/PolyDriver.cpp
@@ -215,7 +215,7 @@ bool PolyDriver::coreOpen(yarp::os::Searchable& prop) {
         str = part->toString().c_str();
     }
 
-#ifndef YARP_NO_DEPRECATE // since yarp 2.3.70
+#ifndef YARP_NO_DEPRECATED // since YARP 2.3.70
     Bottle bot(str.c_str());
     if (bot.size()>1) {
         // this wasn't a device name, but some codes -- rearrange
--- a/src/libYARP_dev/src/PolyDriver.cpp
+++ b/src/libYARP_dev/src/PolyDriver.cpp
@@ -218,8 +218,8 @@ bool PolyDriver::coreOpen(yarp::os::Searchable& prop) {
 #ifndef YARP_NO_DEPRECATE // since yarp 2.3.70
     Bottle bot(str.c_str());
     if (bot.size()>1) {
-		yWarning("Passing 'device' parameter with a list of parameters as value is deprecated. This might be an internal bug. If you didn't do it please report an issue at https://github.com/robotology/yarp");
-		// this wasn't a device name, but some codes -- rearrange
+        // this wasn't a device name, but some codes -- rearrange
+        yWarning("Passing 'device' parameter with a list of parameters as value is deprecated. This might be an internal bug. If you didn't do it please report an issue at https://github.com/robotology/yarp");
         p.fromString(str.c_str());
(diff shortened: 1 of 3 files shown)


src/libYARP_dev/src/PolyDriver.cpp
```diff
@@ -214,13 +214,17 @@ bool PolyDriver::coreOpen(yarp::os::Searchable& prop) {
     if (prop.check("device",part)) {
         str = part->toString().c_str();
     }
+
+#ifndef YARP_NO_DEPRECATE // since yarp 2.3.70
     Bottle bot(str.c_str());
     if (bot.size()>1) {
-        // this wasn't a device name, but some codes -- rearrange
+		yWarning("Passing 'device' parameter with a list of parameters as value is deprecated. This might be an internal bug. If you didn't do it please report an issue at https://github.com/robotology/yarp");
```
drdanz: Please do not use tabs here and in the next line

src/yarprobotinterface/Device.cpp
```diff
@@ -82,7 +82,7 @@ class RobotInterface::Device::Private
     inline yarp::os::Semaphore* lst_sem() const { return &driver->threadListSemaphore; }
 
     inline bool isValid() const { return drv()->isValid(); }
-    inline bool open() { return drv()->open(paramsAsProperty().toString()); }
+	inline bool open() { return drv()->open(paramsAsProperty()); }
```
drdanz: Also here

src/yarprobotinterface/Device.cpp
```diff
@@ -129,13 +129,33 @@ class RobotInterface::Device::Private
...
+			if (counter != 0){
+				yWarning() << "Parentheses not balanced for param " << param.name();
+			}
+
+			yarp::os::Value value;
+			value.fromString(param.value().c_str());
+			prop.put(param.name(),value);
+		}
+
+		return prop;
```
drdanz: And in this whole block

src/yarprobotinterface/Param.cpp
```diff
@@ -105,6 +105,8 @@ bool RobotInterface::Param::isGroup() const
 yarp::os::Property RobotInterface::Param::toProperty() const
 {
     yarp::os::Property p;
-    p.put(mPriv->name.c_str(), mPriv->value.c_str());
+	yarp::os::Value v;
+	v.fromString(mPriv->value.c_str());
```
drdanz: and here

src/libYARP_dev/src/PolyDriver.cpp
```diff
@@ -214,13 +214,17 @@ bool PolyDriver::coreOpen(yarp::os::Searchable& prop) {
     if (prop.check("device",part)) {
         str = part->toString().c_str();
     }
+
+#ifndef YARP_NO_DEPRECATE // since yarp 2.3.70
```
drdanz: Should be `YARP_NO_DEPRECATED` (a `D` is missing)
Since you should already change this line, please make also "YARP" uppercase in the comment...

barbalberto:
Tests fails on Linux, with the error:
```
[ 65%] Building CXX object src/yarprobotinterface/CMakeFiles/yarprobotinterface.dir/Device.cpp.o
/home/travis/build/robotology/yarp/src/yarprobotinterface/Device.cpp: In member function ‘bool RobotInterface::Device::Private::open()’:
/home/travis/build/robotology/yarp/src/yarprobotinterface/Device.cpp:85:63: error: no matching function for call to ‘yarp::dev::PolyDriver::open(yarp::os::Property)’
     inline bool open() { return drv()->open(paramsAsProperty()); }
```

@massimoregoli @drdanz  can you please have a look?

drdanz:
Rebased and fixed the error on linux and a -Wsign-compare warning

drdanz:
@massimoregoli Merged, thanks!

pattacini:
I'm having big problems launching the `cartesiancontrollerserver` components along with the simulator.

If I type:
```
$ yarprobotinterface --context simCartesianControl --config no_legs.xml
```

I get:
```
[ERROR]SolverNameToConnect option is missing
```

The option is there.

@drdanz has this error something to do with this patch?

drdanz:
@pattacini can you post the whole log?

pattacini:
Here it is: [error.txt](https://github.com/robotology/yarp/files/649741/error.txt).

pattacini:
Ok, with the help of @drdanz, I think we got it.

The **groups in the conf file** are passed on to the device driver's method `open()` as **big string**, which is in turn a mistake. Here's the new [log.txt](https://github.com/robotology/yarp/files/649888/log.txt).

**Could you please fix this asap**?

drdanz:
@pattacini Should be fixed now

pattacini:
Great, it works!
Thanks for the swift response.

PR #996, opened 2016-12-02
Cleanup

drdanz:
Just some cleanup using `clang-tidy` as helper.


src/libYARP_OS/src/BottleImpl.cpp
```diff
@@ -835,7 +835,7 @@ void StoreDouble::fromString(const ConstString& src)
     // YARP Bug 2526259: Locale settings influence YARP behavior
     // Need to deal with alternate versions of the decimal point.
     ConstString tmp = src;
-    size_t offset = tmp.find(".");
+    size_t offset = tmp.find('.');
```
lornat75: I guess this was tested, in principle it is not equivalent (like many other changes in this PR).

lornat75: Fine!

PR #997, opened 2016-12-05
Depth camera fix config file and add missing documentation

barbalberto:
Fix an error in the config file parsing
Add documentation

Other minor fix and cleanup


barbalberto:
Test passed, can we merge?

PR #999, opened 2016-12-05
Yarp view autosize

randaz81:
added --autosize option
added menu item which allows the main yarpview window to automatically resize to the new frame size when a video stream width/height is time-variant.  

tested with:
yarpdev --device test_grabber -mode size


drdanz:
@randaz81 merged, thanks

PR #1006, opened 2016-12-10
Yarpmanager to load sub-folders by default

pattacini:
I found myself more comfortable having `yarpmanager` to load applications also from sub-folders **by default**.

This way, the usual path `~/.local/share/yarp/applications` can have a tree of sub-folders without impacting `yarpmanager`.

As of now, in fact, to make `yarpmanager` deal with the above configuration, we have to create a context containing the proper option.

cc @Tobias-Fischer @clement-moulin-frier

### DIFF
--- a/src/yarpmanager/src-manager/main.cpp
+++ b/src/yarpmanager/src-manager/main.cpp
@@ -148,7 +148,7 @@ int main(int argc, char *argv[])
     }
 
     if(!config.check("load_subfolders")){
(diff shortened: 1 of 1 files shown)


drdanz:
@apaikan Can you please review this change?

pattacini:
Ping ⏰ 😉

pattacini:
Thanks @apaikan ❤️

PR #1009, opened 2016-12-12
DepthCameraDriver: segfault on device shutdown fixed

aerydna:


pattacini:
@aerydna again, you should comment more on the PR you open.
For example, which device you mean to fix? I can see it's the RGBD, but only figuring out from your branch.

It's definitely better to detail this when opening PR's.

drdanz:
Merged! Thanks.
I agree with @pattacini though...

drdanz:
Also next time please add a line or two in the relative release notes file (in this case doc/release/v2_3_68_1.md).

PR #1017, opened 2016-12-13
CMake: Fixes to yarp_add_idl()

drdanz:
* When generating ROS idls for paths containing ".." the files are no
  longer generated every time make is executed.
* When path contains something like ".." the temporary files are no
  longer generated outside current build directory.
* Mark the generated files as generated.
* Force cmake to re-run when the file is modified, even if the previous
  version had no output files.


randaz81:
For example, this fixes a problem with "robotology/navigation" repository which was rebuilt from scratch every time a "make" command was issued, even if no files were changed.

drdanz:
Merged since the build was successful on my fork.

PR #1021, opened 2016-12-13
laserFromDepth

randaz81:
This new device driver allows to quickly convert depth data (from a kinect-like device) to laser data.
More specifically, this device reads an horizontal line from the depth image published by a IRGBDSensor capable device and publishes a laser stream like a IRangefinder2D device.


barbalberto:
Looks good to me

barbalberto:
Should not be a PR into devel??

PR #1031, opened 2016-12-19
Yarpmanager impro

### LINKED ISSUE #885: Introducing user defined (environment) variables in application description files
It would be very useful to add more flexibility to the way we can parametrise an application (defined by a **yarpmanager** application description file), by introducing user defined environment variables whose scope would be limited to the application, as it is done in the `icub-tests` xml scripts, where such parameters are wrapped by the tag `<environment>`.

Currently, the closest we have to that is the `<prefix>` mechanism, which allows setting a prefix to all the ports opened by the application. For instance :

```
<!-- Embedded application -->
   <application>
      <name>Application_Name</name>
      <prefix>icubGazeboSim</prefix>
   </application>
   ...
```

In some cases this is not enough. Let's assume we write an application which defines connections between ports that have different prefixes (using the `external` parameter doesn't fully solve the problem), or opens a port for which the variable part in the name is not a prefix...

Even if other solutions exist, using a user defined variable here would be much simpler and efficient, like:

```
<!-- Embedded application -->
   <application>
      <name>Application_Name</name>
      <environment>--robotname icubGazeboSim --context leftLeg</environment >
   </application>
   ...
```

aerydna:
fixes #885 
waiting an @apaikan check to quickly extend to other xmlparsing classes (module resources etc).

`$ENV{foo}` will access to the environment variable "foo"

`<var name="bar">baz</var>` will create a baz-valued, bar variable
then
`${bar}` will access to the "bar" variable and so will give you the value "baz"


apaikan:
@aerydna great! 
just a comment: how  is the error handling in `parseText(const char* element)`?  for example in case of undefined variable , we need to throw a warning message or an error message should be prompted for wrong format (e.g., `${MyVar` or `$ENVV{test}, ..`
examples: 
```xml
 <parameters> ${MyVar <parameters>
 <parameters> $ENNV{MyVar} <parameters>
 <parameters> $FOO{MyVar} <parameters>
...
```xml

**Comment by aerydna** [comment]
for undefined variable i would follow the linux behaviour, where `set MYVAR=` followed by `echo $MYVAR` prints an empty line. for generic error checking maybe the best thing is to checking the use of `$`  and print some warning (to allow the use of $ for personal uses).

**Comment by drdanz** [comment]
@apaikan @aerydna why was this pr closed without merging it?

**Comment by apaikan** [comment]
by mistake!

**Comment by apaikan** [comment]
@aerydna  Undefined variables are fine! However, 
- typos such a `${MyVar`  should be handled.
- the invalid keywords such a `$ENVV{test}` should be handled too.  
- usually XML is case insensitive. But we should decide if `${FOO}` and `${foo}` are same for us or not!  

Linux bash interprets differently `'${FOO}'` and `"${FOO}"`! For example: 

```
$ export FOO=bar
$ echo "$FOO"
bar

$ echo '$FOO'
$FOO
```

aerydna:
less WIP and more definitive at the moment.. but some test have to be done.. i'll continue next year. happy new year to whom are reading!

PR #1033, opened 2016-12-19
CMakeLists.txt ported to new OpenNI2 variables

giuliavezzani:
This change is necessary for fixing [ycm/#111](https://github.com/robotology/ycm/pull/111).


barbalberto:
:+1:

PR #1036, opened 2016-12-20
New broken test

aerydna:
adds a broken test regarding wrong effect misleading port name after ros node creation. see issue #1037

### DIFF
--- a/tests/libYARP_OS/NodeTest.cpp
+++ b/tests/libYARP_OS/NodeTest.cpp
@@ -52,6 +52,14 @@ public:
 
 private:
 
+    void wrongNamePort()
+    {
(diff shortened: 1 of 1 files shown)


tests/libYARP_OS/NodeTest.cpp
```diff
@@ -257,6 +265,11 @@ void NodeTest::typePropTest() {
 
 void NodeTest::runTests() {
     NetworkBase::setLocalMode(true);
+
+#ifdef BROKEN_TEST
```
randaz81: Fine for me, but what is the purpose of this BROKEN_TEST macro? i.e. why this test is not always executed?

PR #1043, opened 2016-12-21
Deprecate yarp::os::Value::operator==(const char *alt)

### LINKED ISSUE #1028: Value val == 0 gives SEGFAULT
```
    Value val = 1000;
    if (val == 0)
```
The == operator for values implicitly assumes that the right operand is a string, thus this kind of code gives segmentation fault.

C.f.r. line 268 of `yarp::os::Value::operator==`
```
 // comparisons with strings worked "accidentally", users depend on them
    bool operator==(const char *alt) const {
        return asString() == alt;
    }
```
This behaviour emerged using `bottle.get(n)` without `.asInt()`

aerydna:
fixes #1028.
`<yarp::os::Value> == <literal integer>` calls the `const char*` comparison overload (often resulting in a segfault). new `int` comparison overload has been added with relative test.

### DIFF
--- a/src/libYARP_OS/include/yarp/os/Value.h
+++ b/src/libYARP_OS/include/yarp/os/Value.h
@@ -276,6 +276,16 @@ public:
         return asString() != alt;
     }
 
+    bool operator==(const int alt) const
+    {
+        return asInt() == alt;
+    }
+
+    bool operator!=(const int alt) const
+    {
+        return asInt() != alt;
+    }
+
     /**
      * Set value to correspond to a textual representation.
      * For example if str="10" then the value will be an integer,
--- a/tests/libYARP_OS/ValueTest.cpp
+++ b/tests/libYARP_OS/ValueTest.cpp
@@ -235,6 +235,11 @@ public:
             checkTrue((v1!=v2), "(blob) operator!= ok");
             checkTrue((v1!=*(reinterpret_cast<const int*>(v2.asBlob()))), "(blob) value ok");
         }
+
+        {
+            Value v(0);
+            checkTrue(v == 0, "equivalence with integer literal ok");
+        }
     }
 
     virtual void runTests() {


src/libYARP_OS/include/yarp/os/Value.h
```diff
@@ -276,6 +276,16 @@ class YARP_OS_API yarp::os::Value : public Portable, public Searchable {
         return asString() != alt;
     }
 
+    bool operator==(const int alt) const
```
randaz81: why are we not removing the bool operator==(const char *alt) mess?

randaz81: Have you evaluated the impact of removing the == operator (simply counting the number of uses (i.e. errors) by commenting out the operator?

randaz81: Additional considerations for multipurpose data-type like value: operator == always generates problems and should be generally avoided. What happens for example if the value internally stored was a double? With this commit you will allow users to compare it with an int. The comparison is ok (more or less), but I fear it may generate other misusage problems in the future...

tests/libYARP_OS/ValueTest.cpp
```diff
@@ -235,6 +235,11 @@ class ValueTest : public UnitTest {
             checkTrue((v1!=v2), "(blob) operator!= ok");
             checkTrue((v1!=*(reinterpret_cast<const int*>(v2.asBlob()))), "(blob) value ok");
         }
+
+        {
+            Value v(0);
+            checkTrue(v == 0, "equivalence with integer literal ok");
```
randaz81: Unclear to me: what we want to test here?
According to the lately introduced `YARP_DEPRECATED_MSG`, shouldn't this be `v.asInt() == 0`?

randaz81:
Rebase requested

tests/libYARP_OS/ValueTest.cpp
aerydna: i've simply forgot to delete the test

randaz81:
Rebase requested

drdanz:
What happened here? You removed the deprecation stuff and restored the first version of the patch...

drdanz:
FYI we deleted the branch and force pushed the merge with the correct version of the patch.

PR #1052, opened 2017-01-04
fixes narrowing conversion error

randaz81:
Fixes the following error on Jetson TK1 (arm64 architecture):

`/usr/local/src/robot/yarp/src/libYARP_OS/src/AbstractCarrier.cpp:222:36: error: narrowing conversion of ‘-1’ from ‘int’ to ‘char’ inside { } [-Wnarrowing]`


randaz81:
Merged into `devel `since I think that this error deals with c++11 stuff, so the error should be not present into `master `branch

PR #1053, opened 2017-01-05
[yarpmanager] Unify viewModule and viewApplication, fix #1007

### LINKED ISSUE #1007: Opening single application multiple times shouldn't be allowed in yarpmanager
Hey @apaikan 

@Tobias-Fischer saw that it's possible to open the same application multiple times within the `yarpmanager`.

It shouldn't be an intended behavior, isn't it?

Tobias-Fischer:
Hi,
This fixes #1007. Previously it didn't work as `a->getAppName()` returns `""` (which should be fixed as well I guess).
Now, the same logic as in `viewModule` and `viewResource` is being used to correctly only being able to open an application once.

Best, Tobi

### DIFF
--- a/src/yarpmanager/src-manager/mainwindow.cpp
+++ b/src/yarpmanager/src-manager/mainwindow.cpp
@@ -470,14 +470,7 @@ void MainWindow::viewModule(yarp::manager::Module *module)
 void MainWindow::viewApplication(yarp::manager::Application *app,bool editingMode)
 {
     for(int i=0;i<ui->mainTabs->count();i++){
-        GenericViewWidget *w = (GenericViewWidget *)ui->mainTabs->widget(i);
-        if(w->getType() ==  yarp::manager::APPLICATION){
-            ApplicationViewWidget *a = (ApplicationViewWidget *)w;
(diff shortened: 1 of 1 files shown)


drdanz:
@apaikan Can you please review this change?

apaikan:
@Tobias-Fischer nice catch! :

Nicogene:
Hi @Tobias-Fischer,

in #1206 on `master` I reverted (not intentionally) the changes of this PR because I didn't know that this bug was already  fixed in `devel`(usually bugs are fixed on `master` branch).

I used a different logic to check if the the application is already open, not based on `appName`(that could be not unique), but based on `fileName`(it has to be unique).

Can you please check if my changes are good to you?

I tested and it works, but maybe you intended a different thing.
Thanks

Tobias-Fischer:
Hi @Nicogene,
As long as it works, it's fine with me :+1:. Did you try whether the merge from `master` into `devel` went fine, i.e. whether #1206 also works fine in `devel`?
Best, Tobi

Nicogene:
It works, but honestly I prefer your solution, less file changed, more clean, in fact we will restore your changes in #1221

PR #1065, opened 2017-01-15
fakeLaser with map option

randaz81:
added option `--test map_test` to load a map inside `FakeLaser `and simulate obstacle detection.


src/libYARP_OS/include/yarp/os/impl/SemaphoreImpl.h
```diff
@@ -20,7 +20,8 @@ namespace yarp {
 //Decide which implementation to use
 #include <yarp/conf/system.h>
 
-#ifdef YARP_HAS_CXX11
+#if 0
```
drdanz: Please remove this change as it is not related to this patch

src/devices/fakeLaser/CMakeLists.txt
```diff
@@ -20,7 +20,8 @@ if(NOT SKIP_fakeLaser)
                             fakeLaser.cpp)
   target_link_libraries(fakeLaser YARP_OS
                                   YARP_sig
-                                  YARP_dev)
+                                  YARP_dev
+                                  YARP_math)
```
drdanz: You need to add a `get_property` a few lines before to find YARP_math headers

randaz81: done

PR #1067, opened 2017-01-16
ServerFrameGrabber enhancement

Nicogene:
These commits introduce:

- `ServerFrameGrabber`: moving implementation from .h to .cpp
- `ServerFrameGrabber`: addition of `yarp::dev::Implement_RgbVisualParams_Parser` and `yarp::dev::IRgbVisualParams`as private members, for parsing rgb camera parameters.
- `TestFrameGrabber`: moving implementation from .h to .cpp
- `TestFrameGrabber`: moved in folder src/devices/test_grabber and turn it in plugin.
- `TestFrameGrabber`: added the inheritance from `yarp::dev::IRgbVisualParams` and implemented the relative virtual methods.
- Added a test for checking the functionality of `ServerFrameGrabber` and `TestFrameGrabber` after these changes.

All these changes have been moreover tested with `FrameGrabberGui2` and `usbCamera` for checking backward compatibility

Please review code.


tests/CMakeLists.txt
```diff
@@ -118,6 +118,11 @@ if(YARP_COMPILE_TESTS)
     add_definitions(-DYARP_ANALOGWRAPPER_TESTS)
   endif()
 
+  # Compile TestFrameGrabberTest only if fakeAnalog is enabled
```
barbalberto: Comment: only if test_grabber is enabled ;-)

Nicogene:
@barbalberto I forgot to change PolyDriverTest, now test_grabber is an optional plugin, and this test try to open it also if it has not been enabled :sweat_smile:

tests/CMakeLists.txt
Nicogene: :sweat_smile:

drdanz:
Looks good to me, but I'll leave the last word to @barbalberto :wink:

src/libYARP_dev/CMakeLists.txt
```diff
@@ -84,7 +84,6 @@ set(YARP_dev_HDRS include/yarp/dev/all.h
                   include/yarp/dev/ServerSerial.h
                   include/yarp/dev/ServerSoundGrabber.h
                   include/yarp/dev/ServiceInterfaces.h
-                  include/yarp/dev/TestFrameGrabber.h
```
drdanz: One small issue...
`TestFrameGrabber.h` is a public API in master. Should we deprecate it instead of removing it?
@randaz81 @pattacini @barbalberto @traversaro

barbalberto: We are not REmoving it, just moving it  :-) 
It has been transformed into a proper plugin. It is still used for testing purpose.

Nicogene:
@barbalberto commit imported, you're welcome :wink:

drdanz:
As discussed with @Nicogene, we are delaying this until #1069 is merged in devel

drdanz:
Merged. Thanks @Nicogene

PR #1073, opened 2017-01-24
YarpPluginFixture:Add the check for the available portmonitors

Nicogene:
Please review code.

### DIFF
--- a/src/rtf-plugins/fixture-managers/yarpplugin/YarpPluginFixture.cpp
+++ b/src/rtf-plugins/fixture-managers/yarpplugin/YarpPluginFixture.cpp
@@ -24,14 +24,14 @@ using namespace yarp::dev;
 
 PREPARE_FIXTURE_PLUGIN(YarpPluginFixture)
 
-bool YarpPluginFixture::scanPlugins(ConstString name){
+bool YarpPluginFixture::scanPortmonitor(ConstString name){
     YarpPluginSelector selector;
     selector.scan();
     Bottle lst=selector.getSelectedPlugins();
     bool res=false;
     for (int i=0; i<lst.size(); i++) {
         Value& options = lst.get(i);
-        if(name == options.check("name",Value("untitled")).asString())
+        if(name == options.check("name",Value("untitled")).asString() && "portmonitor"== options.check("type",Value("untitled")).asString())
             res=true;
     }
     return res;
@@ -42,7 +42,7 @@ bool YarpPluginFixture::setup(int argc, char** argv) {
     if(argc<=1)
     {
         //RTF::RTF_FIXTURE_REPORT();
-        RTF_ASSERT_ERROR("YarpPluginFixture: Please specify --devices and/or --plugins and/or --carriers parameters");
+        RTF_ASSERT_ERROR("YarpPluginFixture: Please specify --devices and/or --portmonitor and/or --carriers parameters");
         return false;
     }
     RTF_FIXTURE_REPORT("YarpPluginFixture: setupping fixture...");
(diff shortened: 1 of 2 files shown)


src/rtf-plugins/fixture-managers/yarpplugin/YarpPluginFixture.cpp
```diff
@@ -24,15 +24,21 @@ using namespace yarp::dev;
...
+bool YarpPluginFixture::scanPlugin(yarp::os::ConstString name, yarp::os::ConstString type){
     YarpPluginSelector selector;
     selector.scan();
     Bottle lst=selector.getSelectedPlugins();
     bool res=false;
     for (int i=0; i<lst.size(); i++) {
         Value& options = lst.get(i);
-        if(name == options.check("name",Value("untitled")).asString())
-            res=true;
+        if(type=="portmonitor"){
```
drdanz: I think that in order to make it easier to extend later, this could be something like this:

```c++
if(!type.empty()) {
    //...
} else {
    // ...
}
```

Nicogene: You are right, done :smile:

src/rtf-plugins/fixture-managers/yarpplugin/YarpPluginFixture.cpp
```diff
@@ -42,11 +52,11 @@ bool YarpPluginFixture::setup(int argc, char** argv) {
     if(argc<=1)
     {
         //RTF::RTF_FIXTURE_REPORT();
-        RTF_ASSERT_ERROR("YarpPluginFixture: Please specify --devices and/or --plugins and/or --carriers parameters");
+        RTF_ASSERT_ERROR("YarpPluginFixture: Please specify --devices and/or --portmonitor and/or --carriers parameters");
```
drdanz: `--plugins` is missing here.

Also `--devices`, `--carriers` and `--plugins` are plural, i think this should be plural as well...

src/rtf-plugins/fixture-managers/yarpplugin/YarpPluginFixture.h
```diff
@@ -29,10 +29,11 @@ class YarpPluginFixture : public RTF::FixtureManager {
     virtual bool setup(int argc, char** argv);
     virtual bool check();
     virtual void tearDown();
-    bool scanPlugins(yarp::os::ConstString name);
+    bool scanPlugin(yarp::os::ConstString name, yarp::os::ConstString type="");
```
drdanz: Unrelated to the patch, but since you changed the name, perhaps this method should be private? Or protected? I don't think it is useful to have it public...

By the way, just out of curiosity, why did you change `scanPlugins` to `scanPlugin`?

src/rtf-plugins/fixture-managers/yarpplugin/YarpPluginFixture.h
```diff
@@ -29,10 +29,11 @@ class YarpPluginFixture : public RTF::FixtureManager {
...
 private:
     yarp::os::Bottle devices;
     yarp::os::Bottle plugins;
+    yarp::os::Bottle portmonitor;
```
drdanz: Maybe this variable name should be plural as well?

src/rtf-plugins/fixture-managers/yarpplugin/YarpPluginFixture.h
Nicogene: Nothing, I simply did a mess with plural and singular, it's better if I put all plural, this function included :sweat_smile:

drdanz:
Merged, thanks!

PR #1079, opened 2017-01-31
WIP fix of fishy motorgui bug

aerydna:
bug founded following some warnings. it seems that in some unknown cases the motorgui, instead of deleting the "/" character at the beginning of the robot/part name was deleting the first character (regardless of the character).

@randaz81 what do you think?


aerydna:
solved by @randaz81 instead

PR #1086, opened 2017-02-06
FindACE: actually preserve the state of CMAKE_TRY_COMPILE_CONFIGURATION

PeterBowman:
Presumably a copy-paste overlook introduced in 54892aa, see previous `if` clause and the *Reset variables...* section further down in which this variable will not be restored if `CMAKE_REQUIRED_INCLUDES` happens to *not* have been defined.


drdanz:
Merged, thanks

PR #1108, opened 2017-02-22
IControlLimits2Raw inherit publicly from IControlLimitsRaw

jgvictores:
`IControlLimits2Raw` inherit publicly from `IControlLimitsRaw`, similar to other interfaces; allows b/w compatibility viewing `IControlLimitsRaw`.

Default `private`inheritance looked more like a bug/typo than something really intentional.


barbalberto:
:+1:

drdanz:
Merged and cherry-picked in `master`. Even if it is an ABI change this look like a bug to me.

PR #1109, opened 2017-02-22
[yarprobotinterface] closeDevices reversed order

francesco-romano:
It is usually a good practice to free resources in the reverse order of the one used to acquired them.
Probably this code won't affect any devices, but I think it may avoid future issues.

cc @drdanz @randaz81 @traversaro


drdanz:
Cherry-picked also in master

PR #1138, opened 2017-03-22
Added conversion from/to rotation matrix <-> yaw-pitch-roll angles.

ale-git:
Added conversion from/to rotation matrix <-> yaw-pitch-roll angles (Rx*Ry*Rz order), to be used with R1 Gazebo equivalent model.


drdanz:
Please, next time add a line to the release notes

PR #1147, opened 2017-03-27
Oculus touch IJoypadController implementation

aerydna:
Implementation of the IJoypadController interface for teleoperation purpose

### DIFF
--- a/src/devices/ovrheadset/OVRHeadset.cpp
+++ b/src/devices/ovrheadset/OVRHeadset.cpp
@@ -4,10 +4,6 @@
  * CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
  */
 
-#define AXIS_COUNT 8
-#define STICK_COUNT 2
-#define BUTTON_COUNT 13
-#define INPUTERRORCHECK if (inputStateError) return false
 #include "OVRHeadset.h"
 #include "InputCallback.h"
 #include "TextureBuffer.h"
@@ -34,6 +30,10 @@
 #include <dxgi.h> // for GetDefaultAdapterLuid
 #pragma comment(lib, "dxgi.lib")
 #endif
+
+static constexpr unsigned int AXIS_COUNT   = 8;
+static constexpr unsigned int STICK_COUNT  = 2;
+static constexpr unsigned int BUTTON_COUNT = 13;
 
 #if defined(_WIN32)
  #define GLFW_EXPOSE_NATIVE_WIN32
@@ -159,7 +159,6 @@ yarp::dev::OVRHeadset::OVRHeadset() :
         inputStateError(false)
 {
     yTrace();
-
 }
 yarp::dev::OVRHeadset::~OVRHeadset()
 {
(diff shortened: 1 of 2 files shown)


aerydna:
@drdanz  the first commit appear to be committed by two user (me and you).. i didn't even know it was possible.. what can i do?

src/devices/ovrheadset/OVRHeadset.cpp
```diff
@@ -852,7 +885,9 @@ void yarp::dev::OVRHeadset::run()
     yarp::os::Stamp predicted_stamp(distortionFrameIndex, predicted_ts.HeadPose.TimeInSeconds);
 
     // Get Input State
-    ovr_GetInputState(session, ovrControllerType_Active, &inputState);
+    inputStateMutex.lock();
+    result = ovr_GetInputState(session, ovrControllerType_Active, &inputState);
```
randaz81: I wonder if putting this blocking call on a separated thread may improve performance of this display thread?

src/devices/ovrheadset/OVRHeadset.cpp
```diff
@@ -4,7 +4,10 @@
  * CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
  */
 
-
+#define AXIS_COUNT 8
+#define STICK_COUNT 2
+#define BUTTON_COUNT 13
```
drdanz: Please do not use `#define`, you can use  `static constexpr size_t AXIS_COUNT = 8;` (or `int`, or whatever it is expected) that works pretty much in the same way, but it is type safe.

Also please move these below the `#include`s

src/devices/ovrheadset/OVRHeadset.cpp
```diff
@@ -151,19 +155,167 @@ yarp::dev::OVRHeadset::OVRHeadset() :
...
+    //    predicted = port.second.find("predicted") != std::string::npos;
+    //    prefix = predicted ? "/oculus/predicted" : "/oculus";
+    //    name = prefix + "/headpose/" + port.second + ":o";
+    //    if (!port.first->open(name)) {
+    //        yError() << "Cannot open" << port.second << "port";
+    //        this->close();
+    //        return false;
+    //    }
+    //    port.first->setWriteOnly();
+    //}
```
drdanz: Please, leave this comment out of the PR, since it is not related...
I'm partially ok with this refactoring (I'd use `std::map` instead of `std::vector<std::pair>` and `emplace` instead of `push_back`), but we can discuss this on a separate PR

src/devices/ovrheadset/OVRHeadset.cpp
```diff
@@ -151,19 +155,167 @@ yarp::dev::OVRHeadset::OVRHeadset() :
...
         logoEnabled(true),
         crosshairsEnabled(true),
-        batteryEnabled(true)
+        batteryEnabled(true),
+        inputStateError(false)
 {
     yTrace();
-}
 
+}
```
drdanz: This change is not useful, please put the bracket back in its place. Also the other spaces removed later are not necessary, but I won't block the PR for this 🙈

src/devices/ovrheadset/OVRHeadset.cpp
```diff
@@ -1334,3 +1497,126 @@ void yarp::dev::OVRHeadset::DebugHmd(ovrHmdDesc hmdDesc)
...
+        value.push_back(atan2(inputState.Thumbstick[stick_id].y, inputState.Thumbstick[stick_id].x));
+    }
+    value.push_back(inputState.Thumbstick[stick_id].x);
+    value.push_back(inputState.Thumbstick[stick_id].y);
+    return true;
+}
+bool yarp::dev::OVRHeadset::getTouch(unsigned int touch_id, yarp::sig::Vector& value) 
+{
+    return false;
+}
```
drdanz: Missing newline at the end of the file

src/devices/ovrheadset/OVRHeadset.cpp
```diff
@@ -4,7 +4,10 @@
  * CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
  */
 
-
+#define AXIS_COUNT 8
+#define STICK_COUNT 2
+#define BUTTON_COUNT 13
+#define INPUTERRORCHECK if (inputStateError) return false
```
drdanz: Please remove this macro and just use the `if (inputStateError) return false`... When I read a function I want to know if and when it is returning...

src/devices/ovrheadset/OVRHeadset.h
```diff
@@ -17,6 +17,9 @@
 #include <GL/glew.h>
 #include <OVR_CAPI.h>
 #include <OVR_CAPI_GL.h>
+#include <yarp/dev/IJoypadController.h>
```
drdanz: This include should be with the previous block

src/devices/ovrheadset/OVRHeadset.h
```diff
@@ -64,6 +82,11 @@ class OVRHeadset : public yarp::dev::DeviceDriver,
     static void glfwErrorCallback(int error, const char* description);
     static void ovrDebugCallback(uintptr_t userData, int level, const char* message);
     static void DebugHmd(ovrHmdDesc hmdDesc);
+    inline void errorManager(ovrResult error);
+    inline void fillAxisStorage();
+    inline void fillErrorStorage();
+    inline void fillButtonStorage();
+    inline void fillHatStorage();
```
drdanz: This should be fine in this specific case since the methods are private, but it is dangerous in the general case... when you declare a function `inline` you should have the implementation in the same file, or you risk to get an "unresolved external" error from the linker.

To be honest, I don't think that the compiler here will `inline` these functions, since their definition is not that short, so I'd just remove the inline.

drdanz:
Merged, thanks

PR #1150, opened 2017-03-29
JoypadControlServer: Fix -Wmismatched-tags warnings

drdanz:
@aerydna I'm not sure whether you prefer to keep them as `struct`s and therefore change the forward declaration like I did in this path, or if you'd prefer to change them to `class`...


aerydna:
for me it's fine to keep them as structs

drdanz:
Merged, thanks for the feedback.

PR #1151, opened 2017-03-29
TcpRosCarrier fix

randaz81:
This commit fixes a critical issue in TcpRosCarrier causing a severe slowdown of yarp::os::publisher when multiple ros clients are connected to the same topic.


drdanz:
We should add a line to the release notes...

PR #1157, opened 2017-03-30
[WIP] IFrameProvider Interface 

aerydna:
Interface representing a provider of frames (and possibly their velocity and acceleration). 
30/03/2017 [WIP] is for a discussion before definitively putting it into yarp.
it is supposed to replace the previous one (currently use for the codyco demo) in https://github.com/robotology-playground/xsens-mvn/blob/master/yarp/IFrameProvider.cpp 
@randaz81 @DanielePucci @traversaro @francesco-romano @barbalberto

### DIFF
--- a/src/libYARP_dev/CMakeLists.txt
+++ b/src/libYARP_dev/CMakeLists.txt
@@ -88,7 +88,8 @@ set(YARP_dev_HDRS include/yarp/dev/all.h
                   include/yarp/dev/TestMotor.h
                   include/yarp/dev/Wrapper.h
                   include/yarp/dev/ControlBoardInterfacesImpl.inl
-                  include/yarp/dev/IRobotDescription.h)
+                  include/yarp/dev/IRobotDescription.h
+                  include/yarp/dev/IFrameProvider.h)
 
 set(YARP_dev_IMPL_HDRS )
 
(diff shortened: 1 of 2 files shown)


aerydna:
disclaimer: There might be some oversight.. i've opened the pr immediatly to begin the discussion.

randaz81:
Similar functionalities are already provided in yarp by:
- https://github.com/robotology/yarp/blob/master/src/libYARP_dev/include/yarp/dev/IFrameTransform.h
- https://github.com/robotology/yarp/tree/master/src/libYARP_dev/src/devices/FrameTransformClient
- https://github.com/robotology/yarp/tree/master/src/libYARP_dev/src/devices/FrameTransformServer

I cannot find any good reason to add "yet another interface", instead of extending the exiting ones, implementing the missing functionalities (vel and accs in this case?). Additionally the philosophy behind http://wiki.ros.org/tf2 is by far superior to the interface concept. I fear that here we are really losing time in reinventing the wheel... :-(

barbalberto:
This requirements emerged from a Dynamic Control group (Nori &C.) because they need to broadcast also velocity and acceleration along with the position/orientation. True that ros t2 has a lot of stuff, we can use those messages for the ros integration.
Yarp does not support them yet since out TF handles only position and orientation, so I think we have to add them on way or another.

I agree in extending the current one, but we have to make up our mind how to do it. I guess this was the purpose of this PR. 
One way is to add more stuff in an existing interface (with usual back compatibility issues), the other one is to have a new interface _deriving_ from the previous and thus extending it.
This commit does neither of the two so I think it is not a good starting point.

Also the type of data used here is **not** what we decided in the meeting and it has to be changed.

_**I suggest to move this discussion/work proposal to robotology-playground**_ since this stuff is probably gonna change a lot before it gets stable and we can play with it more freely.

aerydna:
i think we need a live discussion on this topic because the matter here is not so simple.

PR #1158, opened 2017-03-31
[WIP] Frame Transform type Expansion (after new release)

aerydna:
added linear + angular velocity and acceleration

### DIFF
--- a/src/libYARP_math/include/yarp/math/FrameTransform.h
+++ b/src/libYARP_math/include/yarp/math/FrameTransform.h
@@ -15,119 +15,55 @@
 
 namespace yarp
 {
-    namespace math
+namespace math
+{
+
+
+
+class YARP_math_API FrameTransform
+{
+public:
+    std::string src_frame_id;
+    std::string dst_frame_id;
+    double      timestamp;
+
+    struct Translation_t
     {
-        class YARP_math_API FrameTransform
+        double tX;
+        double tY;
+        double tZ;
+
+        void set(double x, double y, double z)
         {
-            public:
-            std::string src_frame_id;
-            std::string dst_frame_id;
-            double      timestamp;
+            tX = x;
+            tY = y;
+            tZ = z;
+        }
+    } translation;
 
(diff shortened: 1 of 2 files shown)


src/libYARP_math/include/yarp/math/FrameTransform.h
```diff
@@ -13,121 +13,70 @@
 #include <yarp/math/Math.h>
 #include <yarp/math/Quaternion.h>
 
+#define VALID_POSITION         1  // 0b00000001
+#define VALID_ROTATION         2  // 0b00000010
+#define VALID_LIN_VELOCITY     4  // 0b00000100
```
barbalberto: Ususally instead of write directly the numbers `2`, `4` and so on, those values are written like `1 << 1`, `1 << 2` in such a way it is more clear they are used for bitwise operation and not as a value.

src/libYARP_math/include/yarp/math/FrameTransform.h
```diff
@@ -13,121 +13,70 @@
...
+#define VALID_LIN_ACCELERATION 8  // 0b00001000
+#define VALID_ANG_VELOCITY     16 // 0b00010000
+#define VALID_ANG_ACCELERATION 32 // 0b00100000
+#define VALID_FULL VALID_POSITION | VALID_ROTATION | VALID_LIN_VELOCITY | VALID_LIN_ACCELERATION | VALID_ANG_VELOCITY | VALID_ANG_ACCELERATION
+
```
barbalberto: Are those deines intended to be used by the user application or only for internal usage?
In the latter case, we shall move them in the `.cpp`, in the former can they be enum or at least can the name be more specific, like `TF_VALID_FULL`? I'm a bit worried about possible name conflict with other interfaces or similar.

francesco-romano: I agree with @barbalberto on moving them to an enum. I would avoid preprocessor macros for these use cases.

PR #1093, opened 2017-02-15
Transform client update

aerydna:
adds timed transforms on frameTransform suite

PR #1164, opened 2017-04-04
Oculus touch tracking

aerydna:
position and rotation tracking via tf.

### DIFF
--- a/src/devices/ovrheadset/OVRHeadset.cpp
+++ b/src/devices/ovrheadset/OVRHeadset.cpp
@@ -21,9 +21,11 @@
 #include <yarp/os/Time.h>
 #include <yarp/sig/Image.h>
 #include <yarp/os/LockGuard.h>
+#include <yarp/os/Property.h>
+#include <yarp/math/FrameTransform.h>
 
-#include <cmath>
-
+#include <math.h>
+#include <unordered_map>
 #include <OVR_CAPI_Util.h>
 
 #if defined(_WIN32)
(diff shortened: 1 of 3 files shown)


aerydna:
waiting #1093 to be reviewed and merged. added [wip] only for this reason

src/devices/ovrheadset/OVRHeadset.cpp
```diff
@@ -61,6 +63,8 @@ static constexpr unsigned int BUTTON_COUNT = 13;
 #undef check
 #endif
```
barbalberto: Is this `#ifdef(win) ...` stuff imported from the OVR external library? I think the best place for this stuff is the CMake file, what do you think @drdanz ?stuff

randaz81: @barbalberto can you please tell us if this comment is still valid or if it refers to an outdated commit?

drdanz:
Merged thanks

src/devices/ovrheadset/OVRHeadset.cpp
```diff
@@ -21,9 +21,11 @@
 #include <yarp/os/Time.h>
 #include <yarp/sig/Image.h>
 #include <yarp/os/LockGuard.h>
+#include <yarp/os/Property.h>
+#include <yarp/math/FrameTransform.h>
 
-#include <cmath>
-
+#include <math.h>
```
drdanz: :-1:

PR #1186, opened 2017-04-23
README.md: macOS: tap and install in a single command

vitorgalvao:
No need to tap and install separately. This command will do both.

Related, that repo has a confusing name. Opened https://github.com/robotology/homebrew-cask/issues/7 to discuss that.


drdanz:
Sorry for the delay. Merged, thanks!

PR #1206, opened 2017-05-08
yarpmanager: fix SaveAs button, before it doesn't do anything

### LINKED ISSUE #755: yarpmanager save as... button does not seem to work
Hi, 
the Save as... menu option of yarpmanager++ seems to not do anything, independently of whether the open application has been modified, or not and you just want to change the name. 
launching yarpmanager++ with sudo does not help either (actually, a different version of it is opened, which might be bug too).

Nicogene:
The changes introduced by this PR fix #755:

TODO before merging
- [x] fix `actionSave_As` 
- [x] Add warning in case of a new file created with the same name
- [x] Fix of addition to the tree of multiple application pointing to the same xml file
- [x] Fix double tab open of the same application
- [x] add shortcuts for the actions
- [x] tests
- [x] update yarp changelog

Ready to be merged :+1:

### DIFF
--- a/src/libYARP_manager/src/manager.cpp
+++ b/src/libYARP_manager/src/manager.cpp
@@ -86,6 +86,10 @@ Manager::~Manager()
 
 bool Manager::addApplication(const char* szFileName, char* szAppName_, int len)
 {
+    if(find(listOfXml.begin(), listOfXml.end(),szFileName) == listOfXml.end())
+        listOfXml.push_back(szFileName);
+    else
+        return false;
     XmlAppLoader appload(szFileName);
     if(!appload.init())
         return false;
--- a/src/libYARP_manager/src/manager.cpp
+++ b/src/libYARP_manager/src/manager.cpp
@@ -84,7 +84,7 @@ Manager::~Manager()
 }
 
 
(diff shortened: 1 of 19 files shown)


src/libYARP_manager/src/manager.cpp
```diff
@@ -396,6 +404,12 @@ bool Manager::updateConnection(unsigned int id, const char* from,
     return true;
 }
 
+Node* Manager::getNode(string appName)
+{
+    Node* node=YARP_NULLPTR;
+    node = knowledge.getNode(appName);
```
drdanz: :+1:

src/yarpmanager/src-manager/mainwindow.cpp
```diff
@@ -1058,6 +1086,69 @@ void MainWindow::onSave()
...
+            return;
+        }
+        reportErrors();
+        onSave();
+        ww->setAppName(oldAppName);
+        ww->setFileName(oldFileName);
+        onTabClose(ui->mainTabs->currentIndex());
+        syncApplicationList();
+        Application* newApp= YARP_NULLPTR;
+        newApp = (Application*) lazyManager.getNode(currentAppName.toStdString());
```
drdanz: :+1:

PR #1224, opened 2017-05-22
Transform client with ports

randaz81:
Added the following command to FrameTransformClient
* list_frames: print all the available refence frames
* list_ports: print all the opened ports for tranform broadcasting
* publish_transform <src> <dst> <portname> <format>: opens a port to publish transform from src to dst
* unpublish_transform <portname>: closes a previously opened port to publish a transform
* unpublish_all <portname>: closes a all previously opened ports to publish a transform


src/libYARP_dev/src/devices/FrameTransformClient/FrameTransformClient.cpp
```diff
@@ -736,6 +844,54 @@ bool yarp::dev::FrameTransformClient::waitForTransform(const std::string &target
     return true;
 }
 
+FrameTransformClient::FrameTransformClient() : RateThread(0.010)
```
randaz81: fixed

PR #1226, opened 2017-05-22
The ResourceFinder Class (basic) tutorial is misleading

aikolina:
While reading [The ResourceFinder Class (basic)](http://www.yarp.it/master/yarp_resource_finder_basic.html) tutorial I noticed that there is NO "random_motion" in the examples directory, BUT the correct file is "rf_basic". So I modified the tutorial accordingly.


drdanz:
Merged, thanks

PR #1231, opened 2017-05-24
Fix Mutex behavior in yarp::dev::BoschIMU

diegoferigo:
Referring to https://github.com/robotology/yarp/issues/1229, these commits should fix the sharing of  data between the `run()` and `read()` methods.

I still have to test this branch on the robot, in the meantime please let me know if you have any feedback.

Particularly, I changed the behavior when a sensor fails to read a measure. Before the missing measure was set to zero, I think it has more sense setting its value to the previous measurement.

cc @barbalberto @traversaro @francesco-romano

### DIFF
--- a/src/devices/imuBosch_BNO055/imuBosch_BNO055.cpp
+++ b/src/devices/imuBosch_BNO055/imuBosch_BNO055.cpp
@@ -17,6 +17,7 @@
 #include <yarp/os/LogStream.h>
 #include <yarp/os/Log.h>
 #include <yarp/math/Math.h>
+#include <yarp/os/LockGuard.h>
 
 #include "imuBosch_BNO055.h"
 
@@ -435,8 +436,6 @@ void BoschIMU::run()
 //     void *tmp = (void*) &response[2];
 //     raw_data = static_cast<int16_t *> (tmp);
 
-    // TODO: how to optimally protect only code filling the data vector?
-
     // In order to avoid zeros when a single read from a sensor is missing,
     // initialize the new measure to be equal to the previous one
     data_tmp = data;
@@ -553,15 +552,13 @@ void BoschIMU::run()
     }
 
     // Protect only this section in order to avoid slow race conditions when gathering this data
-    mutex.lock();
-
-    data       = data_tmp;
-    quaternion = quaternion_tmp;
-
-    mutex.unlock();
-
-    if(timeStamp > timeLastReport + TIME_REPORT_INTERVAL)
     {
+        LockGuard guard(mutex);
+        data       = data_tmp;
+        quaternion = quaternion_tmp;
+    }
+
+    if (timeStamp > timeLastReport + TIME_REPORT_INTERVAL) {
         // if almost 1 errors occourred in last interval, then print report
         if(errs.acceError + errs.gyroError + errs.magnError + errs.quatError != 0)
         {
@@ -582,7 +579,7 @@ void BoschIMU::run()
 
 bool BoschIMU::read(yarp::sig::Vector &out)
 {
-    mutex.lock();
+    LockGuard guard(mutex);
     out.resize(nChannels);
     out.zero();
 
@@ -595,7 +592,6 @@ bool BoschIMU::read(yarp::sig::Vector &out)
         out[15] = quaternion.z();
     }
 
-    mutex.unlock();
     return true;
 };
 
--- a/src/devices/imuBosch_BNO055/imuBosch_BNO055.cpp
+++ b/src/devices/imuBosch_BNO055/imuBosch_BNO055.cpp
@@ -435,12 +435,9 @@ void BoschIMU::run()
 
     // TODO: how to optimally protect only code filling the data vector?
 
-    yarp::sig::Vector data_tmp;
-    data_tmp.resize(nChannels);
-    data_tmp.zero();
-
-    // Save the last errors count required later
-    errCounter errs_old = errs;
+    // In order to avoid zeros when a single read from a sensor is missing,
+    // initialize the new measure to be equal to the previous one
+    yarp::sig::Vector data_tmp = data;
 
     ///////////////////////////////////////////
     //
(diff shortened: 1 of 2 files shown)


src/devices/imuBosch_BNO055/imuBosch_BNO055.cpp
```diff
@@ -436,9 +435,9 @@ void BoschIMU::run()
 
     // TODO: how to optimally protect only code filling the data vector?
 
-    mutex.wait();
-
-    data.zero();
+    // In order to avoid zeros when a single read from a sensor is missing,
+    // initialize the new measure to be equal to the previous one
+    yarp::sig::Vector data_tmp = data;
```
traversaro: Personal Nitpick : the creation of a `yarp::sig::Vector` involves dynamical allocation of memory, that is tipically a costly a non-realtime safe operation. For vector of this sizes the dynamic the performance hit is probably negligible also due to the optimization in the memory allocator. Anyhow for cleanness I typically allocate buffers only once (i.e. as class members), but feel free to due what you seem more fit.

francesco-romano: Given than we are cleaning this part, you should declare the variable as a class instance.

barbalberto:
What to do when the the device fails to read is always a difficult point. The most important thing is that the read methods returns false and the user checks it. 
Then data will not be valid in any case, so whether it is set to zero, to previous value or to -infinite actually should not matter since they shall not be used anyway.
I think that it is more easy to spot the sensor mulfunction if all the values are set to zero but, on the other hand, I know it may be dangerous if the code trusts the values without doing any check on the actual data.

I'll let the user of the code (mainly wholebody guys) decide what they prefer.
In case we choose for keeping the latest value, a kind of watchdog mechanism warning the user that values are not changing from a long time would be good.

src/devices/imuBosch_BNO055/imuBosch_BNO055.cpp
```diff
@@ -463,93 +462,101 @@ void BoschIMU::run()
...
 
     // If 100ms have passed since the last received message
     if (timeStamp+0.1 < yarp::os::Time::now())
     {
 //         status=TIMEOUT;
     }
-    mutex.post();
+
+    // Protect only this section in order to avoid slow race conditions when gathering this data
+    mutex.lock();
```
francesco-romano: I would use a lock guard. It is exception-safe.

src/devices/imuBosch_BNO055/imuBosch_BNO055.cpp
barbalberto: I agree, it is better to allocate buffers once and for all as a class member.
It'll not impact on performaces drammatically, but devices usually run on pc104 or low power computer, so any improvement easy to implement is welcome

src/devices/imuBosch_BNO055/imuBosch_BNO055.cpp
barbalberto: I think lock_guard will actually lock all the function, not only a small portion, but I may be very wrong.

> The class lock_guard is a mutex wrapper that provides a convenient RAII-style mechanism for owning a mutex _**for the duration of a scoped block.**_

From http://en.cppreference.com/w/cpp/thread/lock_guard

francesco-romano:
@traversaro 
> Just for completeness, it is worth mentioning that an alternative strategy was to claim the mutex just after every sendReadCommand . This would involve locking the mutex three times for each sensor reading loop, but it would ensure fresher data, at least for a subset of the sensors.

I would prefer a single critical section

diegoferigo:
> Just for completeness, it is worth mentioning that an alternative strategy was to claim the mutex just after every sendReadCommand . This would involve locking the mutex three times for each sensor reading loop, but it would ensure fresher data, at least for a subset of the sensors. @traversaro 

I thought that as well, but it could happen that the `read()` acquires half data from the current call, and half from the previous. The solution I adopted looked cleaner, but I'm open to discuss it.

> What to do when the the device fails to read is always a difficult point. The most important thing is that the read methods returns false and the user checks it. @barbalberto 

Good point. FYI @gabrielenava @S-Dafarra @nunoguedelha. What's your opinion about the zero or previous value?

src/devices/imuBosch_BNO055/imuBosch_BNO055.cpp
francesco-romano: Yes, but you can limit the scope:
```
{
// critical region
}
```

src/devices/imuBosch_BNO055/imuBosch_BNO055.h
```diff
@@ -132,15 +132,17 @@ class yarp::dev::BoschIMU:   public yarp::dev::DeviceDriver,
 {
 protected:
 
-    yarp::os::Semaphore         mutex;
     bool                        verbose;
     short                       status;
     int                         nChannels;
     yarp::sig::Vector           data;
+    yarp::sig::Vector           data_tmp;
```
francesco-romano: You should also call a `resize` on `data_tmp` in the `threadInit` method

src/devices/imuBosch_BNO055/imuBosch_BNO055.cpp
barbalberto: You are right, I just fear some compiler optimization on some OS may remove the scope since there is no real reason for it ...
It'd be very weird behaviour for a compiler, but I'll sleep better with the old fashioned lock()  :smile: 
Anyway it is fine if you are sure it does what we want it to

francesco-romano: Why should a compiler remove a scope?
A part from old versions of VS I don't remember compilers that does not respect the scope. It is something defined and enforced by the language, so the compiler should comply with it

src/devices/imuBosch_BNO055/imuBosch_BNO055.h
diegoferigo: Missed that, thanks!

src/devices/imuBosch_BNO055/imuBosch_BNO055.cpp
```diff
@@ -436,9 +438,9 @@ void BoschIMU::run()
 
     // TODO: how to optimally protect only code filling the data vector?
```
barbalberto: Now we can remove the TODO comment :smile_cat: :laughing: :laughing:

diegoferigo:
Tested on the robot
screen

so_good

drdanz:
@diegoferigo When this PR is ready to be merged, please remove the WIP from the title, thanks!

diegoferigo:
@drdanz It is ready, I forgot the WIP :wink:

drdanz:
Merged, thanks!

PR #1247, opened 2017-05-31
Fix ROS only option in ControlBoardWrapper

francesco-romano:
Still to be tested on the robot


randaz81:
Fine for me, but I'm pretty sure that this was already discussed/fixed in the past months @barbalberto

francesco-romano:
After we find out if this is a regression or not, I would like to test it on the robot before merging.

francesco-romano:
Tested on iCubGenova04

PR #1249, opened 2017-06-01
NameClient: fix yarp name /port accepts

### LINKED ISSUE #1197: 'yarp name get /port accepts' does NOT return anything
Hello, 

I've just started to follow the tutorial on YARP ([Getting started with YARP ports](http://www.yarp.it/note_ports.html)) and I had a little problem while testing the command `yarp name get /port accepts` on Ubuntu 16.04.

According to the tutorial, the result of the command should be the following one:
```
[you type] yarp name get /summer accepts
[response] port /summer property accepts = tcp text text_ack udp mcast shmem
           *** end of message
```
On the contrary, in my case I get nothing:
```
adinale@iitrbcslap125:~$ yarp name get /summer accepts
port /summer property accepts =
*** end of message
```
But apparently, other people who has tested this command got at least a partial results like:
```
port /summer property accepts = udp mcast
*** end of message
```
So I have few questions regarding this issue:
- Is this a common misbehavior?
- Do I need some specific setup to make this command work?
- Is there an updated version of this command that I should use?

Thanks!

**NOTE:** I would like to underline that the ports that I created before reaching this critical point of the tutorial work perfectly!

Nicogene:
This PR Fix #1197.
Now it actually check the available carriers in the machine.

Please review code.

### DIFF
--- a/doc/release/v2_3_68_1.md
+++ b/doc/release/v2_3_68_1.md
@@ -82,6 +82,7 @@ Bug Fixes
   functions.
 * Maximum string size in yInfo(), yDebug(), etc. functions was increased to 1024
   characters.
+* Fixed commands `yarp name get /port accepts` and `yarp name get /port offers`. 
 
 #### YARP_sig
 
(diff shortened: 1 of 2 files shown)


drdanz:
I'm not sure if this is actually used internally for creating the connection, but it is a step towards that goal...

PR #1250, opened 2017-06-01
JoypadControlSuite pre-release patch

aerydna:
added sticks management in the SDLJoypad (including axis reverse). added script execution as partial implementation of the IJoypadController interface for it to be activated in the desired device. added various utility e fix

### DIFF
--- a/src/devices/SDLJoypad/SDLJoypad.cpp
+++ b/src/devices/SDLJoypad/SDLJoypad.cpp
@@ -7,10 +7,17 @@
 #include "SDLJoypad.h"
 #include <cstdio>
 #include <yarp/os/LogStream.h>
+#include <cmath>
+#include <yarp/os/Time.h>
 
 
 using namespace yarp::dev;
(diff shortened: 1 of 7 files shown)


src/devices/SDLJoypad/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ yarp_prepare_plugin(SDLJoypad
 if(ENABLE_SDLJoypad)
 
   set(CMAKE_INCLUDE_CURRENT_DIR ON)
+  set (CMAKE_CXX_STANDARD 11)
```
drdanz: This is not required, all YARP is built with this flag

aerydna: Actually it seems to Be required by some ide to autocomplete standard c++11 specific method (e.g.. std::toString()).. or maybe there are some trick I don't know.. in case, let me know

drdanz:
Rebased and merged. Thanks

PR #1255, opened 2017-06-05
transformServer new feature: Static transform from config file

aerydna:
transformServer can now read tf from his configuration file and publish them as static tf

### DIFF
--- a/src/libYARP_dev/src/devices/FrameTransformServer/FrameTransformServer.cpp
+++ b/src/libYARP_dev/src/devices/FrameTransformServer/FrameTransformServer.cpp
@@ -432,7 +432,56 @@ bool FrameTransformServer::threadInit()
     return true;
 }
 
+bool FrameTransformServer::parseStartingTf(yarp::os::Searchable &config)
+{
+    if (config.find("starting_tf").isInt())
+    {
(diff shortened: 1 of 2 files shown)


src/libYARP_dev/src/devices/FrameTransformServer/FrameTransformServer.cpp
```diff
@@ -432,7 +432,56 @@ bool FrameTransformServer::threadInit()
     return true;
 }
 
+bool FrameTransformServer::parseStartingTf(yarp::os::Searchable &config)
+{
+    if (config.find("starting_tf").isInt())
```
randaz81: This is better:
yarp::os::Searchable& tf_group = config.findGroup("STARTING_TF");

BTW: I'd suggest different group name, such as USER_TF, STATIC_TF etc..

aerydna: @randaz changed!

randaz81:
Tests failed due to an issue in devel now fixed (https://github.com/robotology/yarp/commit/cd5995c9fd013e4e88798b84e67e264879d09b74). 
Branch rebased.

PR #1256, opened 2017-06-06
BufferedPort: add check if the port is closed before calling write()

### LINKED ISSUE #1236: Deadlock on device closure
Me, @aerydna and @barbalberto noticed a deadlock when you close a device(that inherits from `RateThread`) if in `run()`  `prepare()` of a `BufferedPort` is called before that the port is open.
In #1235 I added a `BROKEN_TEST` to reproduce this bug.

Nicogene:
This PR fix #1236.

Please review code.


drdanz:
Cherry-picked in master (2 commits) and rebased

drdanz:
Merged, thanks

PR #1263, opened 2017-06-09
Return empty list as getBusInfo's reply if no connections available

PeterBowman:
Apparently, `rosnode info` (ROS Python utility) always expects a `XMLRPCLegalValue` (see Slave API, output parameter `busInfo`) even if it's empty. Not doing so results in a `ValueError` caused by attempting to unpack three variables from a tuple of two values.

* http://wiki.ros.org/ROS/Slave_API#line-33
* https://github.com/ros/ros_comm/blob/ac9f45f/tools/rosnode/src/rosnode/__init__.py#L528
* https://github.com/ros/ros_comm/blob/ac9f45f/tools/rosnode/src/rosnode/__init__.py#L79

This bug was encountered when a CBW2 configured with `useROS true` (or `only`) was queried by `rosnode info /nameOfNode`. Moreover, `rostopic echo /nameOfTopic` hung the terminal (no output). Tested on ROS Kinetic, Ubuntu 16.04. Please review, should this entail any further implications.

### DIFF
--- a/doc/release/v2_3_68_1.md
+++ b/doc/release/v2_3_68_1.md
@@ -84,6 +84,8 @@ Bug Fixes
   characters.
 * Fixed commands `yarp name get /port accepts` and `yarp name get /port offers`.
 * Fixed deadlock during a device closure after calling a prepare of a closed port.
+* Fixed reply to `yarp::os::Node::Helper::getBusInfo` if no connections available,
+  now it returns an empty list.
 
 #### YARP_sig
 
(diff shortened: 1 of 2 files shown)


drdanz:
@PeterBowman Looks good to me, thanks!
Can you please rebase over latest master and add a line to the `doc/release/v2_3_68_1.md` file?

PeterBowman:
> Can you please rebase over latest master and add a line to the doc/release/v2_3_68_1.md file?

@drdanz done. The first commit's message was fixed, too.

drdanz:
Merged. Thanks a lot!

PR #1264, opened 2017-06-12
yarpview: add check yarpserver is running

barbalberto:
Add check in yarpview to verify yarpserver is running at startup, it was missing.


drdanz:
Merged, thanks

PR #1272, opened 2017-06-16
YarpManager: Fix "Computer" resource check

Nicogene:
This PR introduces:

- add the property "yarprun" to the ports opened throgh `yarp run` or `yarprun`
- add the check for "Computers" if the corresponding port has been opened through `yarprun`

TODO:

- [x] Test on `iCubGenova01` setup

Please review code.


drdanz:
Please rebase it onto devel, the change to yarprun cannot be considered a bug fix, and it should be mentioned in the changelog

drdanz:
Merged, thanks

PR #1274, opened 2017-06-20
joypadControlClient: rpc connection message verbosed

aerydna:
rpc handshake now prints the name of the remote port


drdanz:
Should be on master

drdanz:
Merged, thanks!

PR #1275, opened 2017-06-20
Fix yarp::os::Route

Nicogene:
This PR fixes the `yarp::os::Route::swapNames()` segfault in Windows, and MacOS (@francesco-romano ?).

Please review code.


francesco-romano:
Yes.
I already fixed it, but I did not push it probably.

Still, I don't know why `std::swap` causes a segfault

drdanz:
Merged, thanks

PR #1278, opened 2017-06-23
Ovr device updates

aerydna:
OVRDevice Cleanupped and updated


src/devices/ovrheadset/OVRHeadset.cpp
```diff
@@ -27,15 +27,16 @@
...
 #include <unordered_map>
 #include <OVR_CAPI_Util.h>
+#include <OVR_Math.h>
 
 #if defined(_WIN32)
 #include <dxgi.h> // for GetDefaultAdapterLuid
 #pragma comment(lib, "dxgi.lib")
 #endif
 
-static constexpr unsigned int AXIS_COUNT   = 8;
```
drdanz: This change is not necessary, `constexpr` is c++11 and this is a .cpp file. `YARP_CONSTEXPR` and similar are useful only for public header files where we want to keep the compatibility with c++98

PR #1283, opened 2017-06-27
Event driven joypad

aerydna:
The Joypad EventDriven mode can be activated when using the `IJoypadController` interface by providing a valid pointer to a `IJoypadEvent` object. A thread will start and will call the `IJoypadEvent::action` method containing the detected events at every `RateThread::run`.

### DIFF
--- a/src/libYARP_dev/include/yarp/dev/IJoypadController.h
+++ b/src/libYARP_dev/include/yarp/dev/IJoypadController.h
@@ -46,6 +46,13 @@ public:
     */
     virtual ~IJoypadController(){}
 
+    /**
+      Activate event Driven mode
+     * @brief eventDriven
+     * @param enable a bool to turn on or off the eventDriven mode
+     * * @param event a pointer to a valid yarp::dev::IJoypadEvent object whom action() method will be called on event detection
+     * @return true if succeded. false otherwise
+     */
     virtual bool eventDriven(bool enable, yarp::dev::IJoypadEvent* event = YARP_NULLPTR){return false;}
     virtual bool isEventDriven(){return false;}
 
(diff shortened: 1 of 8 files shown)


src/libYARP_dev/include/yarp/dev/IJoypadController.h
```diff
@@ -155,6 +171,97 @@ class  YARP_dev_API yarp::dev::IJoypadController
...
+    virtual bool getRawAxis(unsigned int axis_id, double& value) = 0;
+    virtual bool getRawStick(unsigned int stick_id, yarp::sig::Vector& value, JoypadCtrl_coordinateMode coordinate_mode) = 0;
+    virtual bool getRawTouch(unsigned int touch_id, yarp::sig::Vector& value) = 0;
+    using IJoypadController::m_actions;
+    using IJoypadController::executeAction;
+    using IJoypadController::parseActions;
+
+public:
+
+    virtual bool getAxisCount(unsigned int& axis_count) YARP_OVERRIDE YARP_FINAL;
```
drdanz: I'm not 100% sure, but I think that `override final` is redundant, `final` implies `override`

aerydna: I thought so too.. but it seems not. compiler gives you the Wsuggest-override warning (with only final) and you can put final in a virtual method also in the first declaration (the one in the top class).. it doesn't make any sense to me but is possible..
so you have to specify final and override..

aerydna: https://stackoverflow.com/questions/29412412/does-final-imply-override

drdanz:
Merged thanks.

PR #1291, opened 2017-07-11
Actually export absolute paths from YARP_CONFIGURE_EXTERNAL_INSTALLATION

PeterBowman:
Instantaneous regression caused by 3e4a85c, suppressing 3dc9e91 just a few minutes after it was committed (both were bundled within #400).

### DIFF
--- a/cmake/YarpInstallationHelpers.cmake
+++ b/cmake/YarpInstallationHelpers.cmake
@@ -200,12 +200,12 @@ function(YARP_CONFIGURE_EXTERNAL_INSTALLATION _name)
                ROBOTS
                STATIC_PLUGINS
                DYNAMIC_PLUGINS)
-    set(${_NAME}_${_dir}_INSTALL_DIR_FULL "${CMAKE_INSTALL_PREFIX}/${${_NAME}_${_dir}_INSTALL_DIR}" CACHE INTERNAL "${_dir} installation directory for ${_name} (absolute)")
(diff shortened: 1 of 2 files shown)


cmake/YarpInstallationHelpers.cmake
```diff
@@ -200,12 +200,12 @@ function(YARP_CONFIGURE_EXTERNAL_INSTALLATION _name)
                ROBOTS
                STATIC_PLUGINS
                DYNAMIC_PLUGINS)
-    set(${_NAME}_${_dir}_INSTALL_DIR_FULL "${CMAKE_INSTALL_PREFIX}/${${_NAME}_${_dir}_INSTALL_DIR}")
+    set(${_NAME}_${_dir}_INSTALL_DIR_FULL "${CMAKE_INSTALL_PREFIX}/${${_NAME}_${_dir}_INSTALL_DIR}" CACHE INTERNAL "${_dir} installation directory for ${_name} (absolute)")
```
drdanz: *_FULL* variables are not supposed to be cached variables, they are generated depending on ${CMAKE_INSTALL_PREFIX} and the relative version, see also how *FULL* and *non-FULL* variables are handled in GNUInstallDirs.
The problem here might be that this is a function and therefore they should be set in the `PARENT_SCOPE` to be available outside of the function.

PeterBowman: @drdanz makes sense, thanks. Fixed at latest commit.

drdanz:
Merged, thanks

PR #1294, opened 2017-07-12
Remove using namespace directive from headers 

### LINKED ISSUE #1285: "using namespace std" directive in several headers
i found the directive `using namespace std;` in 21 headers (libYARP_manager and yarpmanager plus one in the mainwindow.h of yarpdataplayer).

this can cause several problem.. for instance by including `vector` or `map` in ResurceFinder.h arise some ambiguities between std and libYARP_OS.

Nicogene:
This PR remove `using namespace` directive from several headers.
Fixes #1285
Please review code.


drdanz:
Merged, thanks

PR #1301, opened 2017-07-19
Joypad control watchdog

aerydna:
added a network watchdog to the JoypadControlClient to prevent wrong input in case of network issues


drdanz:
Merged, thanks

PR #1304, opened 2017-07-28
Add name server property

Nicogene:
This PR add the property *nameserver* to the port of the nameserver. 
It will be useful to check if exist a `yarpserver` launched with a certain namespace.

Please review code


drdanz:
Merged, thanks

PR #1306, opened 2017-07-31
Add iCubCluster functionality to yarpmanager

### LINKED ISSUE #730: icub-cluster in yarpmanager
It would be really nice to have all the features provided by icub-cluster in yarpmanager++.

Nicogene:
This PR import iCubCluster inside yarpmanager that looks like: 
image

It has been successfully tested on `iCubGenova01` setup and it has (for now) the same functionalities of `icub-cluster.py`.

on `WIN32` the iCubCluster panel is for now disabled (sorry @randaz81 , @pattacini  and @lornat75 :sweat_smile: ).

@traversaro you owe me a beer :stuck_out_tongue:.

We should discuss about the cross dependency with icub-main( it needs only the `cluster-config.xml`) and about the deployer for windows.
We could keep it as it is for now and try to figure out later how to launch commands from and to windows machines.

@claudiofantacci @barbalberto @vtikha 

Fixes #730.

TODO before merging:

- [x] Implement the `More...` button, maybe in a smarter way...
- [x] Make ui nicer  
- [x] Fix the cross dipendence with iCub..
- [x] Fix codacy

### DIFF
--- a/src/libYARP_manager/include/yarp/manager/xmlclusterloader.h
+++ b/src/libYARP_manager/include/yarp/manager/xmlclusterloader.h
@@ -5,8 +5,8 @@
  */
 
 
-#ifndef YARP_MANAGER_XMLCLUSTERLOADER
-#define YARP_MANAGER_XMLCLUSTERLOADER
+#ifndef YARP_MANAGER_XMLCLUSTERLOADER_H
+#define YARP_MANAGER_XMLCLUSTERLOADER_H
 
 #include <yarp/manager/ymm-types.h>
 #include <yarp/manager/manifestloader.h>
@@ -16,7 +16,7 @@
 namespace yarp {
 namespace manager {
 
-struct ClusNode
+struct ClusterNode
 {
     std::string name = "";
     bool display = false;
@@ -34,7 +34,7 @@ struct Cluster
     std::string nameSpace = "";
     std::string nsNode = "";
     std::string ssh_options = "";
-    std::vector<ClusNode> nodes;
+    std::vector<ClusterNode> nodes;
 };
 
 /**
@@ -43,7 +43,7 @@ struct Cluster
 class XmlClusterLoader {
 
 public:
-    XmlClusterLoader(std::string szFileName);
+    XmlClusterLoader(const std::string& szFileName);
     virtual ~XmlClusterLoader();
     bool parseXmlFile(Cluster& _cluster);
 
@@ -54,8 +54,8 @@ private:
     Cluster     cluster;
 };
 
-} // namespace yarp
 } // namespace manager
+} // namespace yarp
 
 
-#endif // __YARP_MANAGER_XMLCLUSTERLOADER_
+#endif // YARP_MANAGER_XMLCLUSTERLOADER_H
--- a/src/libYARP_manager/include/yarp/manager/xmlclusterloader.h
+++ b/src/libYARP_manager/include/yarp/manager/xmlclusterloader.h
@@ -1,0 +1,61 @@
+/*
+ * Copyright (C) 2017 iCub Facility
+ * Authors: Nicolo' Genesio <nicolo.genesio@iit.it>
+ * CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
+ */
+
+
+#ifndef YARP_MANAGER_XMLCLUSTERLOADER
+#define YARP_MANAGER_XMLCLUSTERLOADER
+
+#include <yarp/manager/ymm-types.h>
+#include <yarp/manager/manifestloader.h>
+
+
+
+namespace yarp {
+namespace manager {
+
+struct ClusNode
+{
+    std::string name = "";
+    bool display = false;
+    std::string displayValue = "none";
+    std::string user = "";
+    std::string ssh_options = "";
+    bool onOff = false;
+    bool log = true;
+};
+
+struct Cluster
+{
+    std::string name = "";
+    std::string user = "";
+    std::string nameSpace = "";
+    std::string nsNode = "";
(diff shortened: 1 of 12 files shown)


barbalberto:
Scusate ma 37 email di codacy su una singola PR non sono un tantino eccessive???

claudiofantacci:
@barbalberto, let's look the other way around, there "should be no code pushes" with so many inaccuracies.

Nicogene:
Sorry guys for spam, I tried to make codacy happy but every push it finds something new

diegoferigo:
I've been busier that usual recently and the work slowed down a bit, but most of codacy warnings can be addressed automatically by setting up any [clang-format](https://github.com/robotology-playground/clang-format/).

barbalberto:
It is fine if there are a lot of inaccuracy in the code. The tool is there to catch them all for you.
My point was simply: can we have a single mail with all the messages? 
Or even better: a single mail with "codacy found some issue, go on github and look check them out"
Too many mail are just spam.

claudiofantacci:
@barbalberto that is a valid point! There should be somewhere in settings a way to stop emails. Let me check!

claudiofantacci:
Ok, according to Codacy doc, this kind of option is on GitHub side, under **Webhook and Services**. I can't do anything more as I'm not enabled to change settings on robotology.
@drdanz should look into this.

Anyway, now let's go back on track with the PR :-)

src/yarpmanager/src-manager/clusterWidget.cpp
```diff
@@ -0,0 +1,477 @@
...
+        it->setIcon(0, QIcon(":/computer-available22.svg"));
+    }
+    else
+    {
+        it->setIcon(0, QIcon(":/computer-unavailable22.svg"));
+    }
+
+}
+
+std::string ClusterWidget::getSSHCmd(std::string user, std::string host, std::string ssh_options)
```
drdanz: This kind of warnings is good... instead of `std::string host` you should pass a `const std::string& host`. This saves an expansive copy of the string. Same for the other similar warnings.

drdanz:
I wonder why it says "drdanz deleted a comment from codacy-bot", I didn't delete anything 😮

src/libYARP_manager/include/yarp/manager/xmlclusterloader.h
```diff
@@ -0,0 +1,61 @@
...
+
+protected:
+
+private:
+    std::string confFile;
+    Cluster     cluster;
+};
+
+} // namespace yarp
+} // namespace manager
```
drdanz: This is actually the opposite, i.e.
```c++
} // namespace manager
} // namespace yarp
```

src/libYARP_manager/include/yarp/manager/xmlclusterloader.h
```diff
@@ -0,0 +1,61 @@
...
+
+
+#endif // __YARP_MANAGER_XMLCLUSTERLOADER_
```
drdanz: The comment here is wrong

src/libYARP_manager/src/xmlclusterloader.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+
+#include <yarp/manager/xmlclusterloader.h>
+#include <yarp/manager/utility.h>
+#include <yarp/manager/ymm-dir.h>
+#include <tinyxml.h>
+#include <yarp/os/Value.h>
+#ifdef WITH_GEOMETRY
+#include <yarp/os/Property.h> // for parsing geometry information
+#endif
```
drdanz: what is this ``WITH_GEOMETRY`` used for? I cannot find it anywhere else.

src/libYARP_manager/include/yarp/manager/xmlclusterloader.h
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (C) 2017 iCub Facility
+ * Authors: Nicolo' Genesio <nicolo.genesio@iit.it>
+ * CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
+ */
+
+
+#ifndef YARP_MANAGER_XMLCLUSTERLOADER
```
drdanz: According to our policies this should be `YARP_MANAGER_XMLCLUSTERLOADER_H`

src/libYARP_manager/include/yarp/manager/xmlclusterloader.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include <yarp/manager/ymm-types.h>
+#include <yarp/manager/manifestloader.h>
+
+
+
+namespace yarp {
+namespace manager {
+
+struct ClusNode
```
drdanz: I think that saving the `ter` from cluster here is not really worth, I'd rename it `ClusterNode`

drdanz:
I changed the settings, codacy should now just send one message with a summary.
I'll try it for a while, but to be honest I don't see a good reason for not having the comments on the pull request, so if I don't like the new setting I'll switch back! 💪

randaz81:
"Type the command you wish to execute:" ...
super! what is the syntax? can I execute the command on a generic machine (one of the machines listed above?) It could be useful to start roscore, if needed.

About the --ros option: Can you add an additional green tick to show if the nameserver has been started with --ros option? (just checking if a /ros port is active, I think..)

Nicogene:
"Type the command you wish to execute:"  was already implemented also on the "old" cluster, in this one you can execute on the nodes you selected a command via `ssh -f`.
If the `-- ros` is checked when you check if the nameserver is up it check also that the port `/ros` is up.

drdanz:
Merged, thanks!

PR #1322, opened 2017-08-18
Updated RPATH macro

francesco-romano:
Added a new variable INSTALL_NAME_DIR to specify the absolute installation path of the libraries.
This variable will be used if RPATH is disabled either for the full project or for the installation

### DIFF
--- a/cmake/ycm-0.4.0/modules/AddInstallRPATHSupport.cmake
+++ b/cmake/ycm-0.4.0/modules/AddInstallRPATHSupport.cmake
@@ -6,6 +6,7 @@
 #
 #   add_install_rpath_support([BIN_DIRS dir [dir]]
 #                             [LIB_DIRS dir [dir]]
+#                             [INSTALL_NAME_DIR] [dir]]
 #                             [DEPENDS condition [condition]]
 #                             [USE_LINK_PATH])
 #
@@ -61,8 +62,15 @@
 #  - ``LIB_DIRS`` list of directories to be added to the RPATH. These
 #    directories will be added "relative" w.r.t. the ``BIN_DIRS`` and
 #    ``LIB_DIRS``.
-#  - ``DEPENDS`` list of conditions that should be TRUE to enable
+#  - ``INSTALL_NAME_DIR`` directory where the libraries will be installed.
+#    This variable will be used only if ``CMAKE_SKIP_RPATH`` or
+#    ``CMAKE_SKIP_INSTALL_RPATH`` is set to ``TRUE`` as it will set the 
+#    ``INSTALL_NAME_DIR`` on all targets
+#  - ``DEPENDS`` list of conditions that should be ``TRUE`` to enable
 #    RPATH, for example ``FOO; NOT BAR``.
+#
+# Note: see https://gitlab.kitware.com/cmake/cmake/issues/16589 for further
+# details.
 
 #=======================================================================
 # Copyright 2014 Istituto Italiano di Tecnologia (IIT)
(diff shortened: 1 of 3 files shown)


cmake/ycm-0.4.0/modules/AddInstallRPATHSupport.cmake
```diff
@@ -100,6 +99,19 @@ function(ADD_INSTALL_RPATH_SUPPORT)
                              "${_multiValueArgs}"
                              "${ARGN}")
 
+  # if either RPATH is disabled or INSTALL_RPATH is disable
```
claudiofantacci: Should be *disable**d** at the end of the comment, or rephrased as:
"if either RPATH or INSTALL_RPATH is disabled"

cmake/ycm-0.4.0/modules/AddInstallRPATHSupport.cmake
```diff
@@ -100,6 +99,19 @@ function(ADD_INSTALL_RPATH_SUPPORT)
...
+  # and the INSTALL_NAME_DIR variable is set, then hardcode the install name
+  if(CMAKE_SKIP_RPATH OR CMAKE_SKIP_INSTALL_RPATH)
+    if(DEFINED _ARS_INSTALL_NAME_DIR)
+      set(CMAKE_INSTALL_NAME_DIR ${_ARS_INSTALL_NAME_DIR} PARENT_SCOPE)
+    endif()
```
claudiofantacci: If `_ARS_INSTALL_NAME_DIR` is not defined, shouldn't we raise an error?

cmake/ycm-0.4.0/modules/AddInstallRPATHSupport.cmake
francesco-romano: Thanks! I will update ASAP

francesco-romano:
@claudiofantacci  `if` at line 120 removed. It should be covered by the previous ifs

cmake/ycm-0.4.0/modules/AddInstallRPATHSupport.cmake
francesco-romano: I don't know. There is a possible case which is without RPATH and without absolute path, and I think is the one covered by the fact that we allow the variable to be undefined

claudiofantacci:
@drdanz, by my understanding by default RPATH is enabled either we install YARP in the system or a custom directory.
Shouldn't be good to check whether we are installing in a system directory and disable RPATH?
In this way we have RPATH working only for those cases where it is actually needed and it would also be cleaner when installing through `brew`.

cmake/ycm-0.4.0/modules/AddInstallRPATHSupport.cmake
```diff
@@ -61,6 +62,9 @@
 #  - ``LIB_DIRS`` list of directories to be added to the RPATH. These
 #    directories will be added "relative" w.r.t. the ``BIN_DIRS`` and
 #    ``LIB_DIRS``.
+#  - ``INSTALL_NAME_DIR`` directory where the libraries will be installed.
+#    This variable will be used only if CMAKE_SKIP_RPATH or CMAKE_SKIP_INSTALL_RPATH
+#    is set to TRUE as it will set the INSTALL_NAME_DIR on all targets
```
drdanz: Nitpick: several "``" are missing here...

drdanz:
@claudiofantacci I don't think we should take care of that, normally packaging build script disable the RPATH by setting `CMAKE_SKIP_INSTALL_RPATH`, and the same should be done by the user when installing on a system path

claudiofantacci:
I fear that we are still missing one use case...I can't recall it now :-(

I would suggest to brainstorm for 5 minutes all together.
Anyway, I approve the changes so that you can merge it.

francesco-romano:
@drdanz I updated the documentation. Can you check if it is ok now?

drdanz:
Merged, thanks

PR #1340, opened 2017-09-06
RateThread: add diagnostic message when timing may be an issue.

barbalberto:
Check if RateThread period is negative and in that case an Assert is called.

Add 2 diagnostic messages:

- RateThread period set to 0ms: 0 is not an achievable period and shall not be used. If no specific period is required, a simple `yarp::os::Thread` shall be used instead. This is enclosed as `YARP_NO_DEPRECATED // since YARP 2.3.72` and in future version an Assert will be called as suggested by @drdanz.
- When the `run()` function takes more time to be executed than required period, a warning message will inform the user.

### DIFF
--- a/src/libYARP_OS/src/RateThread.cpp
+++ b/src/libYARP_OS/src/RateThread.cpp
@@ -10,6 +10,7 @@
 #include <yarp/os/impl/Logger.h>
 #include <yarp/os/impl/PlatformTime.h>
 #include <yarp/os/Semaphore.h>
+#include <yarp/os/LogStream.h>
 
 #include <yarp/os/Time.h>
 #include <yarp/os/SystemClock.h>
@@ -21,7 +22,7 @@
 using namespace yarp::os::impl;
 using namespace yarp::os;
 
-//const YARP_timeval _timeout_value(20, 0);    // (20 sec) timeout value for the release (20 sec)
+#define RATETHREAD_NOTIFY_INTERVAL  10.0
 
 class RateThreadCallbackAdapter: public ThreadImpl
 {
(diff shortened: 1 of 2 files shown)


pattacini:
@barbalberto I'd really go for making the **warning message optional** via the API.

barbalberto:
I fear it'll be like not having the message at all. Why do you want to have them optional?

pattacini:
I didn't look at the PR yet 😉 

If the warning message (only the one triggered by the thread overrun) gets printed every once in a while together with other statistics (e.g. `in 1000 run, we had 23 overrun`), then it's fine with me.

By contrast, printing constantly at the thread rate might be critical for certain applications.

barbalberto:
Yes the message is printed at most once every 10seconds (if any) as a gentle reminder, with the number of overrun occurred.

src/libYARP_OS/src/RateThread.cpp
```diff
@@ -299,14 +320,45 @@ RateThread::~RateThread()
...
+
 void RateThread::initWithSystemClock()
 {
     ((RateThreadCallbackAdapter*)implementation)->initWithSystemClock();
 }
 
 bool RateThread::setRate(int period)
 {
-    return ((RateThreadCallbackAdapter*)implementation)->setRate(period);
+    bool ret = checkRequiredPeriod(period);
```
claudiofantacci: The function `checkRequiredPeriod()` either return `true` or stops the application with a `yAssert(false)`.
In my opinion this behaviour it is not clear.
My suggestion is to either change this behaviour, thus allowing a `return false`and then use a `yAssert()` on it or to not a return value at all.

src/libYARP_OS/src/RateThread.cpp
```diff
@@ -299,14 +320,45 @@ RateThread::~RateThread()
     }
 }
 
+bool RateThread::checkRequiredPeriod(int &period)
```
claudiofantacci: My two cents here: since the actual parameter is an `int` and it is not modified by the function, we should simply pass it by value with a `const int`.

barbalberto:
A few other changes have to be done, I'll do when I'll came back from vacation.

lornat75:
Please do not deprecate period=0ms;  it is intended to have a continuous thread which yields the CPU to avoid busy wait. If the documentation is unclear we should improve it. Using a Thread instead requires more coding.

claudiofantacci:
I guess that then we should discuss a bit about this before going any further.

barbalberto:
I agree, as usual the implications are more widespread than expected.

PR #1343, opened 2017-09-13
libYARP_serversql: adds to the help the write option

### LINKED ISSUE #1329: yarpserver's --write option is undocumented in CLI
Typing `yarpserver --help` the option `--write` is not listed/documented.

Bash completion also lacks of this option.

Nicogene:
This PR fixes #1329.

Please review code.

### DIFF
--- a/src/libYARP_serversql/src/yarpserver.cpp
+++ b/src/libYARP_serversql/src/yarpserver.cpp
@@ -205,6 +205,7 @@ yarpserversql_API int yarpserver_main(int argc, char *argv[]) {
 
     if (options.check("help")) {
         printf("Welcome to the YARP name server.\n");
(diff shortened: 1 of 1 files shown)


src/libYARP_serversql/src/yarpserver.cpp
```diff
@@ -205,6 +205,7 @@ yarpserversql_API int yarpserver_main(int argc, char *argv[]) {
 
     if (options.check("help")) {
         printf("Welcome to the YARP name server.\n");
+        printf("  --write                  The configuration file will be written with the address for the name server, if found.\n");
```
drdanz: The length of the line is a bit long compared to the other ones, please split it in 2 lines (see --portdb and --subdb)
Also I'd say "Write IP address and socket on the configuration file" or something similar, for coherence with the other help lines...

PR #1354, opened 2017-09-21
Macos AppNap support

### LINKED ISSUE #1153: macOS AppNap put into idle YARP applications in background
Starting from macOS Mavericks (10.9) macOS supports a new feature (AppNap) with the aim of improving energy efficiency and battery life on notebooks.

Unfortunately, it uses some heuristics to decide which app may be put into "nap" (i.e. in a very low scheduling priority), see official documentation [here](https://developer.apple.com/library/content/documentation/Performance/Conceptual/power_efficiency_guidelines_osx/AppNap.html#//apple_ref/doc/uid/TP40013929-CH2-SW1).

As a consequence some yarp applications (e.g. `yarpdataplayer`) at some point stops sending data to their ports.

As documented in Apple APIs, AppNap may be disabled using (unfortunately only) ObjectiveC calls in [NSProcessInfo](https://developer.apple.com/reference/foundation/nsprocessinfo).

We have to test if this applies also to non UI applications, e.g. a simple main which in a run loop sends random data through an output port.

francesco-romano:
Fixes #1153

### DIFF
--- a/src/libYARP_OS/include/yarp/os/Os.h
+++ b/src/libYARP_OS/include/yarp/os/Os.h
@@ -189,6 +189,20 @@ namespace yarp {
          * @since YARP 2.3.70
          */
         YARP_OS_API int fork(void);
+
+       /**
+        * @brief Toggle the OS energy saving feature
+        *
+        * This function toggle the state of the energy saving feature
+        * that may affect YARP calls.
+        *
+        * For example, in macOS this function toggles the state of the AppNap
+        * feature, which put the process into an idle state.
+        *
+        * @param enable true if the energy saving mode should be enabled. False otherwise.
+        * @since YARP 2.3.72
+        */
+        YARP_OS_API void setEnergySavingModeState(bool enabled);
     }
 }
 
--- a/src/libYARP_OS/src/MacOSAPI.mm
+++ b/src/libYARP_OS/src/MacOSAPI.mm
@@ -1,0 +1,26 @@
+#import "yarp/os/impl/MacOSAPI.h"
+
+#import <Foundation/Foundation.h>
+
+void* disableAppNap()
+{
+    NSProcessInfo *processInfo = [NSProcessInfo processInfo];
+    id activity = nil;
+    // beginActivityWithOptions:reason: is only in macOS 10.9+
+    if ([processInfo respondsToSelector:@selector(beginActivityWithOptions:reason:)]) {
+        activity = [processInfo beginActivityWithOptions:NSActivityLatencyCritical | NSActivityUserInitiated
+                                                 reason:@"YARP requires AppNap off"];
+    }
+    return activity;
+}
+
+void restoreAppNap(void *activityInfo)
+{
+    id activity = (id)activityInfo;
+    if (!activity) return;
+    // There is no need to check for macOS versions here,
+    // as we are passing an object which is obtained only
+    // on the correct version of macOS
+    NSProcessInfo *processInfo = [NSProcessInfo processInfo];
+    [processInfo endActivity:activity];
+}
--- a/src/libYARP_OS/src/Os.cpp
+++ b/src/libYARP_OS/src/Os.cpp
@@ -142,3 +142,22 @@ int yarp::os::fork()
     pid_t pid = yarp::os::impl::fork();
     return pid;
 }
+
+
+#if defined(__APPLE__)
+#include "yarp/os/impl/MacOSAPI.h"
+#endif
+
+
+void yarp::os::setEnergySavingModeState(bool enabled)
+{
+#if defined(__APPLE__)
+    static void* handle = 0;
+    if (!enabled && !handle) {
+        handle = disableAppNap();
+    } else {
+        restoreAppNap(handle);
+    }
+
+#endif
+}
--- a/doc/release/v2_3_72.md
+++ b/doc/release/v2_3_72.md
@@ -38,6 +38,7 @@ New Features
 * Added the property *yarprun* to ports opened through `yarprun --server`.
 * Added the property *nameserver* for the port opened through `yarpserver`.
 * The method `yarp::os::Network::runNameServer()` was removed.
+* Added possibility to enable and disable macOS AppNap (Os.h)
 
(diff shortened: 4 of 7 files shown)


src/libYARP_OS/include/yarp/os/Os.h
```diff
@@ -189,6 +189,20 @@ namespace yarp {
...
+        * This function toggle the state of the energy saving feature
+        * that may affect YARP calls
+        *
+        * For example, in macOS this function toggles the state of the AppNap
+        * feature, which put the process into an idle state.
+        *
+        * @param enable true if the energy saving mode should be enabled. False otherwise.
+        * @since YARP 2.3.72
+        */
+        YARP_OS_API void toggleEnergySavingMode(bool enable);
```
drdanz: "toggle" does not seem the correct verb to use with a boolean switch, since toggle means something that switches between on and off every time you call it...
Since I suspect we will use it only for disabling the energy saving mode, perhaps `disableEnergySavingMode(bool disable = true)` might be better?

src/libYARP_OS/src/Os.cpp
```diff
@@ -142,3 +142,22 @@ int yarp::os::fork()
     pid_t pid = yarp::os::impl::fork();
     return pid;
 }
+
+
+#ifdef __APPLE__
```
drdanz: please use `#if defined(__APPLE__)` for coherence with the other parts of the code

src/libYARP_OS/include/yarp/os/Os.h
```diff
@@ -189,6 +189,20 @@ namespace yarp {
          * @since YARP 2.3.70
          */
         YARP_OS_API int fork(void);
+
+       /**
+        * @brief Toggle the OS energy saving feature
+        *
+        * This function toggle the state of the energy saving feature
+        * that may affect YARP calls
```
claudiofantacci: Missing a . at the end of the phrase.

src/libYARP_OS/include/yarp/os/Os.h
```diff
@@ -189,6 +189,20 @@ namespace yarp {
...
+        *
+        * For example, in macOS this function toggles the state of the AppNap
+        * feature, which put the process into an idle state.
+        *
+        * @param enable true if the energy saving mode should be enabled. False otherwise.
+        * @since YARP 2.3.72
+        */
+        YARP_OS_API void setEnergySavingModeEnabled(bool enabled);
```
claudiofantacci: @drdanz already commented for the name of this method, but I put my two cents here as well.
I may be picky here, but I would not use `set` and `Enable` in an `ON/OFF` method. Simply using `setEnergySavingMode(bool status)` should be self-explanatory: pass `true/false` to enable/disable it.

src/libYARP_OS/src/MacOSAPI.mm
```diff
@@ -0,0 +1,19 @@
+#import "yarp/os/impl/MacOSAPI.h"
+
+#import <Foundation/Foundation.h>
+
+void* enableAppNap()
+{
+    NSProcessInfo *processInfo = [NSProcessInfo processInfo];
+    id activity = [processInfo beginActivityWithOptions:NSActivityLatencyCritical | NSActivityUserInitiated
+                                                 reason:@"YARP requires AppNap off"];
```
claudiofantacci: @francesco-romano I remember when we sketched this piece of code that we misunderstood the how to turn AppNap on off. Here the function name suggest that AppNap will be _enabled_, that is background applications will be put in a "sleep" status, but the `processInfo` reason says _"YARP requires AppNap off"_.
What am I getting wrong here?

src/libYARP_OS/src/Os.cpp
```diff
@@ -142,3 +142,22 @@ int yarp::os::fork()
...
+#if defined(__APPLE__)
+#include "yarp/os/impl/MacOSAPI.h"
+#endif
+
+
+void yarp::os::setEnergySavingModeEnabled(bool enabled)
+{
+#if defined(__APPLE__)
+    static void* handle = 0;
+    if (!enabled && !handle) {
```
claudiofantacci: If we pass `false` here we should set energy saving to `OFF`, am I right?
But if `enable == false` it seems that we are asking to activate AppNap.

src/libYARP_OS/include/yarp/os/Os.h
francesco-romano: Then I would prefer `setEnergySavingModeState(bool status)`,
but please, decide together with @drdanz  😛

src/libYARP_OS/src/MacOSAPI.mm
francesco-romano: I will check it again.
I only rebased the old branch and opened the PR, so I didn't check everything.

Thanks to point this out

src/libYARP_OS/src/Os.cpp
francesco-romano: We should just "invert" the names of the inner methods. I'll do it

src/libYARP_OS/include/yarp/os/Os.h
claudiofantacci: For me `setEnergySavingModeState` is fine.
I understand that I can set a status of something passing a boolean `ON/OFF` parameter.
@drdanz is that ok for you?

src/libYARP_OS/src/MacOSAPI.mm
claudiofantacci: No worries! It may be working just fine, but it looks counterintuitive to me :-)

src/libYARP_OS/src/Os.cpp
drdanz: Is this done in the latest patch? @claudiofantacci when we are happy with the changes, I think we can merge this...

francesco-romano:
@drdanz @claudiofantacci  I (hopefully) modified all the requested changes.

claudiofantacci:
LGTM
Thanks!

francesco-romano:
Ok. I think it is time to merge then.

PR #1355, opened 2017-09-21
Removed register storage specifies

francesco-romano:
It is deprecated in C++11

> 3 - A register specifier is a hint to the implementation that the variable so declared will be heavily used. [ Note: The hint can be ignored and in most implementations it will be ignored if the address of the variable is taken. This use is deprecated (see D.2). — end note ]

`lua_swig.h` still contains `register` but I think it is autogenerated by swig


francesco-romano:
Lexer is autogenerated. 
This PR can be closed

PR #1356, opened 2017-09-21
Fix double truncation in Property::fromString()

### LINKED ISSUE #1057: Loss of precision when converting string to double inside a yarp::os::Property
When a string is converted to a double inside a `yarp::os::Property`, the number of decimal digits is internally truncated.  
```
    Property p1;
    Property p2;
    Property p3;
    p1.fromString("(width 0.00001)");
    p2.fromString("(width 0.000001)");
    p3.fromString("(width 0.0000001)");
    double s1 = p1.find("width").asDouble(); / /ok
    double s2 = p2.find("width").asDouble(); // ok
    double s3 = p3.find("width").asDouble(); // ****** s3 == 0 ******
```
The problem affects also parameters read from file.

Nicogene:
This PR fixes double truncation in `Property::fromString()`, and it substitutes in `Property::putBottle()` the inefficient  `a.fromString(b.toString())` with `a = b`.
Fixes #1057 . 
Please review code.

### DIFF
--- a/doc/release/v2_3_70_2.md
+++ b/doc/release/v2_3_70_2.md
@@ -8,6 +8,12 @@ A (partial) list of bug fixed and issues resolved in this release can be found
 Bug Fixes
 ---------
 
+### Libraries
+
+#### YARP_OS
+
+* Fixed truncation of double in `Property::fromString()`.
+
 
 Contributors
 ------------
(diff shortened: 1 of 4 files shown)


drdanz:
Merged, thanks

PR #1360, opened 2017-09-25
Fix Clang warnings - libYARP_OS

damn1:
libYARP_OS - fix file by file
- after the assignement the var is never read


src/libYARP_OS/include/yarp/os/impl/BufferedConnectionWriter.h
```diff
@@ -364,7 +364,6 @@ class YARP_OS_impl_API yarp::os::impl::BufferedConnectionWriter : public Connect
             appendStringBase(str);
         } else {
             ConstString s = str;
-            str += terminate;
```
drdanz: This fix seems wrong, I think that who wrote this meant to write `s += terminate` instead of `str += terminate`

src/libYARP_OS/src/PortCoreInputUnit.cpp
```diff
@@ -164,12 +164,12 @@ void PortCoreInputUnit::run() {
             //printf("HAVE A REFERENCE\n");
             if (localReader!=YARP_NULLPTR) {
                 bool ok = localReader->read(br);
-                if (!br.isActive()) { done = true; break; }
```
drdanz: Is this `done` variable actually used?
There are other few `done = true; break;` in the code that are not removed.

I wonder if for readability of the code it would be better to change all of them to `done = true; continue;` instead...

drdanz:
Also note for myself, we should remember to change the `YARP_NULLPTR` to `nullptr` when merging this into devel

drdanz:
Merged, thanks.

PR #1361, opened 2017-09-25
Fix Clang warnings - libYARP_sig

damn1:
libYARP_sig - fix file by file

### DIFF
--- a/src/libYARP_sig/include/yarp/sig/ImageNetworkHeader.h
+++ b/src/libYARP_sig/include/yarp/sig/ImageNetworkHeader.h
@@ -46,6 +46,12 @@ public:
     yarp::os::NetInt32 paramBlobTag;
     yarp::os::NetInt32 paramBlobLen;
 
+    ImageNetworkHeader() : listTag(0), listLen(0), paramNameTag(0), 
+                           paramName(0), paramIdTag(0), id(0), 
+                           paramListTag(0), paramListLen(0), depth(0),
+                           imgSize(0), quantum(0), width(0),
+                           height(0), paramBlobTag(0), paramBlobLen(0) {}
+
     void setFromImage(const Image& image) {
         listTag = BOTTLE_TAG_LIST;
         listLen = 4;
--- a/src/libYARP_sig/src/Matrix.cpp
+++ b/src/libYARP_sig/src/Matrix.cpp
@@ -93,6 +93,9 @@ bool yarp::sig::submatrix(const Matrix &in, Matrix &out, int r1, int r2, int c1,
     const double *i=in.data()+in.cols()*r1+c1;
     const int offset=in.cols()-(c2-c1+1);
 
+    if(i == YARP_NULLPTR || t == YARP_NULLPTR)
+        return false;
+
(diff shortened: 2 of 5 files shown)


src/libYARP_sig/include/yarp/sig/ImageNetworkHeader.h
```diff
@@ -46,6 +46,25 @@ class yarp::sig::ImageNetworkHeader
     yarp::os::NetInt32 paramBlobTag;
     yarp::os::NetInt32 paramBlobLen;
 
+    ImageNetworkHeader()
```
drdanz: @Nicogene Why is it complaining about this? Perhaps the actual issue is in NetInt32? Anyway, please use the initialization list instead or even better (if it fixes the issue) the defaulted constructor (`ImageNetworkHeader() = default;`)

drdanz: Oh wait, this is master... no `default`

src/libYARP_sig/src/Matrix.cpp
```diff
@@ -36,6 +36,19 @@ class MatrixPortContentHeader
     yarp::os::NetInt32 cols;
     yarp::os::NetInt32 listTag;
     yarp::os::NetInt32 listLen;
+
+    MatrixPortContentHeader()
```
drdanz: Same

src/libYARP_sig/src/Vector.cpp
```diff
@@ -36,6 +36,10 @@ class VectorPortContentHeader
 public:
     yarp::os::NetInt32 listTag;
     yarp::os::NetInt32 listLen;
+    VectorPortContentHeader(){
```
drdanz: same

src/libYARP_sig/include/yarp/sig/ImageNetworkHeader.h
Nicogene: Clang static analyzer complaints about the missing initialization in the constructor.

src/libYARP_sig/include/yarp/sig/ImageNetworkHeader.h
```diff
@@ -46,6 +46,12 @@ class yarp::sig::ImageNetworkHeader
     yarp::os::NetInt32 paramBlobTag;
     yarp::os::NetInt32 paramBlobLen;
 
+    ImageNetworkHeader() : listTag(0), listLen(0), paramNameTag(0), 
+		                   paramName(0), paramIdTag(0), id(0),
```
drdanz: Please remove the 'tab' characters

src/libYARP_sig/src/Vector.cpp
```diff
@@ -36,6 +36,7 @@ class VectorPortContentHeader
 public:
     yarp::os::NetInt32 listTag;
     yarp::os::NetInt32 listLen;
+	VectorPortContentHeader() : listTag(0), listLen(0) {}
```
drdanz: 'tab'

drdanz:
Merged, thanks.

PR #1362, opened 2017-09-25
Fix Clang warnings - libYARP_math

damn1:
libYARP_math - fix file by file

### DIFF
--- a/src/libYARP_math/src/Quaternion.cpp
+++ b/src/libYARP_math/src/Quaternion.cpp
@@ -17,6 +17,7 @@ class QuaternionPortContentHeader
 public:
     yarp::os::NetInt32 listTag;
     yarp::os::NetInt32 listLen;
+    QuaternionPortContentHeader() : listTag(0), listLen(0) {}
 };
 YARP_END_PACK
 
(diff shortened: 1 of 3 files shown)


src/libYARP_math/src/Quaternion.cpp
```diff
@@ -17,6 +17,11 @@ class QuaternionPortContentHeader
 public:
     yarp::os::NetInt32 listTag;
     yarp::os::NetInt32 listLen;
+    QuaternionPortContentHeader()
+    {
+        listTag = 0;
+        listLen = 0;
+    }
```
drdanz: @Nicogene What is the purpose of this? 
Can we use "QuaternionPortContentHeader() = default" instead? Or at least you should use the initialization list...

src/libYARP_math/src/Vec2D.cpp
```diff
@@ -21,6 +21,11 @@ class Vec2DPortContentHeader
 public:
     yarp::os::NetInt32 listTag;
     yarp::os::NetInt32 listLen;
+    Vec2DPortContentHeader()
```
drdanz: Same here...

src/libYARP_math/src/Quaternion.cpp
Nicogene: The purpose is to initialise `listTag` and `listLen`, yes we can use the initialization list.

drdanz:
Merged, thanks.

PR #1363, opened 2017-09-25
Fix Clang warnings - other libraries

damn1:
libYARP_name, libYARP_manager, yarpmanager, yarpmotorgui - fix file by file

- shift operator with undefined behaviour

### DIFF
--- a/src/libYARP_manager/include/yarp/manager/impl/textparser.h
+++ b/src/libYARP_manager/include/yarp/manager/impl/textparser.h
@@ -38,7 +38,6 @@ public:
 
         std::string ret, startKeyword, endKeyword;
         size_t s, e;
-        bool   badSymbol;
 
         ret = "";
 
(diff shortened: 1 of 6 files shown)


src/libYARP_manager/include/yarp/manager/impl/textparser.h
```diff
@@ -38,7 +38,7 @@ class TextParser
 
         std::string ret, startKeyword, endKeyword;
         size_t s, e;
-        bool   badSymbol;
+        bool   badSymbol = false;
```
drdanz: please move the variable creation inside the following if, i.e. `bool badSymbol = ret.find("$") != std::string::npos;`

src/libYARP_manager/src/binexparser.cpp
```diff
@@ -8,21 +8,26 @@
...
 #include <cstdio>
 #include <cstring>
 #include <iostream>
 #include <fstream>
 #include <string>
 #include <algorithm>
 #include <cmath>
+#include <inttypes.h>
+#include <limits.h>
+#include <stddef.h>
```
drdanz: please use `<cinttypes>`, `<climits>`, and `<cstddef>`

drdanz:
Also please rebase the branch over the latest master

drdanz:
Merged, thanks.

PR #1364, opened 2017-09-25
Fix Clang warnings - libYARP_dev

damn1:
libYARP_dev - fix file by file


drdanz:
Merged, thanks.

PR #1370, opened 2017-09-26
Fix mcast yarpserver

Nicogene:
Since `v2_3_70` we support mcast also without ACE.
This PR add the open of the `fallback` port also if `yarp` is compiled with `SKIP_ACE`.

Please review code.


drdanz:
Merged, thanks.

PR #1396, opened 2017-10-09
Changes to YarpLogger

francesco-romano:
This PR does a couple of changes  to Yarplogger.
The current logger behaviour is that when a message arrives and the log has reached the maximum number of messages, the new message is discarded.
With the proposed changes, if the message list is full, the oldest message is removed.

In the meantime there is another useful feature that I want to propose (and I will take care of that):
changing the timestamps to be human readable. Too many time, we look at the logger and we are not able to understand when an error occurred.
Currently there are two timestamp:
- local timestamp (time of the message when it is read, which is transformed into a string)
- yarprun timestamp (a static integer that is incremented every time).

Is somebody using these timing information? I would like to add a "real" timing variable which can be transformed into a `QDateTime` object in the UI.
@drdanz can we have a chat on this? I see two issues: what kind of variables should i add in the header file? How to serialize/deserialize to file in a portable way.

This PR also changes how the log table manages its entries, by using a `QAbstractTableModel` object.


francesco-romano:
@drdanz I will write here the discussion about the time.

https://github.com/robotology/yarp/blob/master/src/libYARP_logger/include/yarp/logger/YarpLogger.h#L124-L125 declares timestamps to be a string. These values are filled from the logger, when a message is received at the following lines:
https://github.com/robotology/yarp/blob/master/src/libYARP_logger/src/YarpLogger.cpp#L299-L300 
but the timestamps are generated either from a simple counter (integer) or from a `time_t` structure.

Leaving aside the `yarprun` timestamp (which is a counter, I don't know if somebody is using it or not. Probably it make sense to just remove this data), I would like to focus on the local timestamp.
In this case I want to save a "native" time and to convert it to a string representation only at the time of saving it to a file. In this way we preserve the actual datetime information.
The problem is: what type of data can I use? I see the following options:
- double (returned by `yarp::os::Time::now()`).  In this case, I can serialise this directly to the file (and read it back). I can then use the [QDateTime::setMSecsSinceEpoch](http://doc.qt.io/qt-5/qdatetime.html#setMSecsSinceEpoch) to retrieve the DateTime. 
- work with `std::chrono` directly. I don't like to put a `time_point` in the header though. 

In every case, I think we will miss the microseconds accuracy when setting the output on the table. 
But I don't know if we are actually using the microseconds.

randaz81:
> This PR also changes how the log table manages its entries, by using a QAbstractTableModel object.
This change is too complex for me do be reviewed just looking at the code. Can you show me a demo?

randaz81:
> @drdanz I will write here the discussion about the time.
As above. If the PR is too big or it changes too many features, I'm not going to merge it. Please split in multiple PRs, each of them addressing a single feature that you want to improve...

francesco-romano:
@randaz81 ok, so we can move the "timestamp" modification to a new PR.
In the end this PR modifies two things (and the feature of removing old messages is just two lines of code).

We can start testing on the robot then for a couple of days before actually merging.

francesco-romano:
Travis failed for a test (on macOS) on the transform client (@drdanz)

drdanz:
@randaz81 Can you please review this?

francesco-romano:
Yarplogger is running on the green iCub since monday afternoon. @gabrielenava can confirm if it is still running fine.
In that case I think it is safe to rebase and merge

gabrielenava:
Yes, yarplogger is still running fine on Green iCub.

francesco-romano:
Thanks. I will do a last test on exporting the log to disk. If it works for me is a "merge" (btw I remember @aikolina was not able to reload a log from the file even in the old version, so I will check also that)

francesco-romano:
Export tested (both single log and full session) and working.
Loading of session does not work, but this seems to confirm what I remember before this PR.
Debugging it seems the call to `tellg()` returns -1 in https://github.com/francesco-romano/yarp/blob/master/src/libYARP_logger/src/YarpLogger.cpp#L768

I don't think this PR should fix this issue, as it is out of scope.

gabrielenava:
While I was doing YOGA++ experiments on iCubGenova04 with @fjandrad, the robot went in hardware fault. We tried to look at the logger for understanding the source of the issue, but it crashed. We open it again and it crashed again. The first time it crashed,  the logger was open on a terminal, so I can attach here the output of its terminal. It seems a segmentation fault problem. 

[logger_terminal.txt](https://github.com/robotology/yarp/files/1398508/logger_terminal.txt)

Furthermore, there is another bug. As soon as the log size is greater than 10000, **all** messages (errors, warning, info) are printed even if the user selects only "errors" (see attached figure). 

screenshot from 2017-10-19 15-09-49

francesco-romano:
Thank you @gabrielenava .
Let's pospone then the merging of this PR after the release.
We can then have more tests.

francesco-romano:
@gabrielenava testing with valgrind for memory problems.
In the meantime I found the issue of the filter. When new messages arrive, the filter is not refreshed.

francesco-romano:
Closing as to properly clean the logger it requires too many modification at the architecture level

PR #1399, opened 2017-10-10
Coverity fix - libYARP_manager

damn1:
Fix file by file of issues raised by Coverity static analyzer on libYARP_manager.

### DIFF
--- a/src/libYARP_manager/include/yarp/manager/application.h
+++ b/src/libYARP_manager/include/yarp/manager/application.h
@@ -364,6 +364,7 @@ public:
         strXmlFile.clear();
         Iapplications.clear();
         resources.clear();
+        appOwner=YARP_NULLPTR;
     }
 
     GraphicModel& getModelBase(void) { return modelBase;}
--- a/src/libYARP_manager/src/localbroker.cpp
+++ b/src/libYARP_manager/src/localbroker.cpp
@@ -838,8 +838,12 @@ bool LocalBroker::startStdout(void)
     }
 
     int oflags = fcntl(pipe_to_stdout[READ_FROM_PIPE], F_GETFL);
-    fcntl(pipe_to_stdout[READ_FROM_PIPE], F_SETFL, oflags|O_NONBLOCK);
-
+    if(fcntl(pipe_to_stdout[READ_FROM_PIPE], F_SETFL, oflags|O_NONBLOCK) == -1)
+    {
+        strError = "cannot set flag on pipe: " + string(strerror(errno));
+        //close(pipe_to_stdout[READ_FROM_PIPE]);
+        return false;
+    }
     Thread::start();
     return true;
 }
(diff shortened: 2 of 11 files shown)


src/libYARP_manager/src/kbase.cpp
```diff
@@ -1625,7 +1625,7 @@ bool KnowledgeBase::constrainSatisfied(Node* node,
         {
             OSTRINGSTREAM msg;
             msg<<"Some resource dependencies of ";
-            msg<<dynamic_cast<Module*>(resource->owner())->getName();
+            msg<< (dynamic_cast<Module*>(resource->owner()) ? dynamic_cast<Module*>(resource->owner())->getName() : "");
```
Nicogene: Why this is marked as warning ? `dynamic_cast<Module*>(resource->owner()` is already checked in the outer if

src/libYARP_manager/src/localbroker.cpp
```diff
@@ -885,8 +889,14 @@ int LocalBroker::ExecuteCmd(void)
         //int saved_stderr = dup(STDERR_FILENO);
         dup2(pipe_to_stdout[WRITE_TO_PIPE], STDOUT_FILENO);
         dup2(pipe_to_stdout[WRITE_TO_PIPE], STDERR_FILENO);
-        fcntl(STDOUT_FILENO, F_SETFL, fcntl(STDOUT_FILENO, F_GETFL) | O_NONBLOCK);
-        fcntl(STDERR_FILENO, F_SETFL, fcntl(STDERR_FILENO, F_GETFL) | O_NONBLOCK);
+        if (fcntl(STDOUT_FILENO, F_SETFL, fcntl(STDOUT_FILENO, F_GETFL) | O_NONBLOCK) == -1) {
```
Nicogene: Instead of `return 0`, in case of error I would return a different number, 0 usually means "everything is ok"

src/libYARP_manager/include/yarp/manager/application.h
```diff
@@ -364,6 +364,7 @@ class Application : public Node{
         strXmlFile.clear();
         Iapplications.clear();
         resources.clear();
+        appOwner=NULL;
```
drdanz: please use YARP_NULLPTR

src/libYARP_manager/src/primresource.cpp
```diff
@@ -166,6 +166,7 @@ Processor::Processor(void) : GenericResource("Processor")
     cpuload.loadAverage1 = (double)0.0;
     cpuload.loadAverage5 = (double)0.0;
     cpuload.loadAverage15 = (double)0.0;
+    cpuload.loadAverageInstant = (double)0.0;
```
drdanz: Sorry, I never noticed this before... These `(double)` casts do not make sense... 0.0 is a double, see http://www.cplusplus.com/doc/tutorial/constants/
@Nicogene I think these should be removed after merging this patch, what do you think?

Nicogene: Good point @drdanz , I agree that we should remove this explicit cast

src/libYARP_manager/src/localbroker.cpp
drdanz: Actually, this function is returning 0 everywhere, so I suppose it's fine here...
@Nicogene you can eventually change the function later, I'm going to merge this now!

PR #1403, opened 2017-10-11
joypadControlClient: recursive call fix

aerydna:
fixed a wrong typo-caused recursive call on JoypadControl


drdanz:
Merged, thanks.

PR #1405, opened 2017-10-12
ros msgs autogeneration bug fix

damn1:
Previous code was generating problems when the message had constant fields declared as last fields.
This fix provides support for this case and also for void messages or messages with only constant fields.


drdanz:
It would be good to have one test that actually falls in this case...

drdanz:
Merged, thanks.

PR #1407, opened 2017-10-13
Yarpviz cleanup

Nicogene:
This PR ectract all the profiling utilities of `yarpviz` to the new library `libYARP_profiler`.
Moreover some cleanup in ex-yarpviz classes has been done

Please review code.

PR #1410, opened 2017-10-16
Improvement yarpmanager

### LINKED ISSUE #1401: The "Check All" function of the Cluster Management Window in yarpmanager doesn't check if the yarprun port is still active
If we launch ``yarprun`` on the robot's head (using ``yarpmanager``) and then we shut the head down, the status in the ``Cluster Management Window`` remains green, even if we press ``Check All``.

@Nicogene @gabrielenava

Nicogene:
After #1407 , it has been possible to introduces some new nice features:
- `yarp clean` button: before actually performing it, it asks which timeout do you want to use. It fixes #1401 
- `yarp name list` button:  "Ports" has been added in the entities tree, it is populated by this button.
:warning:  at the startup `yarpmanager` executes a `yarp name list` command.

image

Please review code.


drdanz:
please be coherent with the "standard" icon names:
* Do not use "_" (`action_yarp_name_list.svg` should be `action-yarp-name-list.svg`) (BTW I noticed that you used that also in the "all" action, please fix them as well)
* Do not use CamelCase (`action-yarpClean.svg` should be `action-yarp-clean.svg`
* Do not use abbreviations (`res`, `mod`, `app`)

Also I'd remove the red "bookmark" symbol (and its shadow) from the yarp name list button

Nicogene:
Moved to #1414, for some reason github refused to update the PR after `push --force`

pattacini:
@Nicogene don't [`push --force`][1] within a PR 😉

[1]: https://help.github.com/articles/about-pull-requests

drdanz:
> @Nicogene [don't `push --force`][1] within a PR 😉

Not sure where that comment on the documentation comes from, but if you cannot push force on PR, then we should move to some other hosting

pattacini:
You can certainly push --force within a PR but then - obviously - you spoil the history of the commits displayed within the PR, as expected, I'd say.

PR #1413, opened 2017-10-17
Coverity fix - libYARP_dev #2

damn1:
Commits left out from https://github.com/robotology/yarp/pull/1387


drdanz:
I wish there was an option in github to add "ignore-all-space" to the diffs 🙏

PR #1414, opened 2017-10-17
Improvement yarpmanager

### LINKED ISSUE #1089: Reduce wait state in yarpmanager when stopping a subset of modules
Hi @apaikan 

@Tobias-Fischer and I have a follow-up to #1051 for the case when **only a subset of modules** is selected to be stopped.

Instead of relying directly on the values specified for the `stop` option within the XML file, it would be better to substract the minimum value among those selected. This way, we are not going to wait unnecessary seconds upon clicking on stop.

#### Example

Say we have 3 selected modules within an application containing many more, and the following settings for the three modules are given in order:

```xml
<ensure>
   <wait when='stop'> 3 </wait> 
</ensure>

<ensure>
   <wait when='stop'> 5 </wait> 
</ensure>

<ensure>
   <wait when='stop'> 10 </wait> 
</ensure>
```

When we stop just the three modules, we could better wait:

- `0` seconds for the first, instead of `3`;
- `2` seconds for the second, instead of `5`;
- `7` seconds for the third, instead of `10`.

We can achieve the above behavior by just subtracting `3` (the minimum) from all of them.
The same should apply also in the case we select only one module: just wait `0` second.

### LINKED ISSUE #1401: The "Check All" function of the Cluster Management Window in yarpmanager doesn't check if the yarprun port is still active

Nicogene:
After #1407 , it has been possible to introduces some new nice features:
- `yarp clean` button: before actually performing it, it asks which timeout do you want to use. It fixes #1401 
- `yarp name list` button:  "Ports" has been added in the entities tree, it is populated by this button.
:warning:  at the startup `yarpmanager` executes a `yarp name list` command.

image

Moreover it fixes #1089 .
Please review code.

demoRedBall TESTED.


drdanz:
Merged, thanks.

PR #1418, opened 2017-10-18
Replace yarp::os::Random

damn1:
Substituted all usages of `yarp::os::Random` API with utilities from C++11's `<random>` library.

### DIFF
--- a/example/game/game_clientmt/main.cpp
+++ b/example/game/game_clientmt/main.cpp
@@ -8,11 +8,11 @@
 #include <time.h>
 #include <stdlib.h>
 #include <string.h>
+#include <random>
 
 #include <yarp/os/all.h>
 
 #include <yarp/os/RateThread.h>
-#include <yarp/os/Random.h>
 #include <yarp/os/Semaphore.h>
 
 #include <yarp/os/Time.h>
@@ -31,6 +31,8 @@ const char GO[]="go";
 class MyPlayer: public RateThread
 {
 public:
+    std::default_random_engine randengine;
+
     MyPlayer(const char *n, int rate):RateThread(rate)
     {
         myX=0;
@@ -57,7 +59,7 @@ public:
 
         myLife=6;
 
-        Random::seed(time(0));
+        randengine.seed(0);
     
         mutex.post();
     }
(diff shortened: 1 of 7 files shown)


src/libYARP_OS/include/yarp/os/Random.h
```diff
@@ -7,6 +7,8 @@
 #ifndef YARP_OS_RANDOM_H
 #define YARP_OS_RANDOM_H
 
+#ifndef YARP_NO_DEPRECATED
```
drdanz: Please add
```diff
+#ifndef YARP_NO_DEPRECATED // Since YARP 2.3.72
```
So that we will find it easily when we will remove the deprecated features in the next release

pattacini:
Please keep this PR on hold. We could
I'd like to discuss the policy to deprecate API since it has clearly a deep impact on user code.

Managing one repository is not as complicated as managing our ecosystem; we need therefore commitment to keep the ecosystem neat as well.

@drdanz @lornat75 @vtikha @traversaro @randaz81

lornat75:
Let's keep this on hold and revise it in person.

drdanz:
Please update the examples as well...

drdanz:
...and yarpviz

damn1:
I revised and completed the PR as @drdanz requested, in case it will have to be accepted.

src/libYARP_OS/include/yarp/os/Random.h
```diff
@@ -26,7 +28,9 @@ namespace yarp {
  *
  * A collection of basic random number generation algorithms.
  */
-class YARP_OS_API yarp::os::Random {
+class YARP_OS_API YARP_OS_DEPRECATED_API YARP_OS_DEPRECATED_API_MSG("use instead utilities from C++11's <random> lib") yarp::os::Random {
```
drdanz: `YARP_OS_DEPRECATED_API` and `YARP_OS_DEPRECATED_API_MSG` are not supposed to be used together, the first is used to deprecate with a standard message, the second with a custom message.
please remove `YARP_OS_DEPRECATED_API`

damn1: :+1: , https://github.com/robotology/yarp/pull/1418/commits/dac3e586fc372f5174712721f0968cfa472b3da3

drdanz:
Merged, thanks.

PR #1420, opened 2017-10-19
Patch for C++11 delay(0)

barbalberto:
After F2F chat with @lornat75 and @pattacini, it has been decided to introduce a patch to restore the same behavior as ACE.

ACE version of `sleep(0)` was forcing a `yield`.
With the introduction of C++11, `sleep(0)` does not force `yield` anymore.
This patch is to reintroduce old behavior and avoid thread starvation.

### DIFF
--- a/src/libYARP_OS/include/yarp/os/RFModule.h
+++ b/src/libYARP_OS/include/yarp/os/RFModule.h
@@ -52,6 +52,15 @@ public:
      * work could be done during this call, or it could just check the
      * state of a thread running in the background.
      *
+     * The thread calls the updateModule() function every <period> seconds.
+     * At the end of each run, the thread will sleep the amounth of time
+     * required, taking into account the time spent inside the loop function.
+     * Example:  requested period is 10ms, the updateModule() function take
+     * 3ms to be executed, the thread will sleep for 7ms.
+     *
+     * Note: after each run is completed, the thread will call a yield()
+     * in order to facilitate other threads to run.
+     *
      * @return true iff module should continue
     */
     virtual bool updateModule() = 0;
--- a/src/libYARP_OS/include/yarp/os/RateThread.h
+++ b/src/libYARP_OS/include/yarp/os/RateThread.h
@@ -60,6 +60,14 @@ public:
 
     /**
      * Loop function. This is the thread itself.
+     * The thread calls the run() function every <period> ms.
+     * At the end of each run, the thread will sleep the amounth of time
+     * required, taking into account the time spent inside the loop function.
+     * Example:  requested period is 10ms, the run() function take 3ms to
+     * be executed, the thread will sleep for 7ms.
+     *
+     * Note: after each run is completed, the thread will call a yield()
+     * in order to facilitate other threads to run.
      */
     virtual void run() = 0;
 
--- a/src/libYARP_OS/src/RFModule.cpp
+++ b/src/libYARP_OS/src/RFModule.cpp
@@ -307,6 +307,13 @@ int RFModule::runModule() {
             break;
         }
 
+        // At the end of each run of updateModule function, the thread is supposed
+        // to be suspended and release CPU to other threads.
+        // Calling a yield here will help the threads to alternate in the execution.
+        // Note: call yield BEFORE computing elapsed time, so that any time spent due to
+        // yield is took into account and the sleep time is correct.
+        yarp::os::Time::yield();
+
         // The module is stopped for getPeriod() seconds.
         // If getPeriod() returns a time > 1 second, we check every second if
         // the module stopping, and eventually we exit the main loop.
--- a/src/libYARP_OS/src/RateThread.cpp
+++ b/src/libYARP_OS/src/RateThread.cpp
@@ -13,7 +13,6 @@
 
 #include <yarp/os/Time.h>
 #include <yarp/os/SystemClock.h>
-
 #include <cmath> //sqrt
 
 //added threadRelease/threadInit methods and synchronization -nat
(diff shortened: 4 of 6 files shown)


pattacini:
Just want to add a consideration on this.

If we leave aside for a moment all the discussion we had related to the misuse of a mutex within a thread whose duration exceeds the period, I'd like to point out that - at least for me - what is reasonable to expect when the single call to run ends is that the scheduler takes over the control and decides who is the next to be running.

Assume that we set up a thread to run each `10 ms`,  but then the body of the run is as follows:
```c++
void yarp::os::RateThread::run()
{
    // thread body
    yarp::os::Time::delay(1.0);
}
```
For me, it's more reasonable that when the single instance of `run()` ends, the task-scheduler brings up and decides which thread to wake up, instead of calling again `run()` straightaway.

claudiofantacci:
> For me, it's more reasonable that when the single instance of run() ends, the task-scheduler brings up and decides which thread to wake up

It's reasonable.

For the record, let me motivate this non-trivial "issue"/"discussion".

In general, everything sums up to:  
_**Does the scheduler need to be invoked when sleep is invoked with 0?**_

Why 0?
Well, as an example, if we look at [line 180 of RateThread::SingleStep()](http://yarp.it/RateThread_8cpp_source.html#l00180) you will notice that may be invoked with 0 (or maybe even with a negative number).
That leads us to [sleepThread()](http://yarp.it/PlatformTime_8cpp_source.html#l00034) which has 3 different kind of sleep calls:
 - `ACE_OS::sleep(sleep_period)`
 - `usleep(sleep_period.tv_sec * 1000000 + sleep_period.tv_usec )` that is converted in a `nanosleep()` call
 - `Time::delay(toDouble(sleep_period))` that brings to [`SystemClock::delaySystem`](http://yarp.it/SystemClock_8cpp_source.html#l00014) and finally to a `std::this_thread::sleep_for()` call.

Now, from the [`ACE_OS::sleep()`](http://www.aoc.nrao.edu/php/tjuerges/ALMA/ACE-5.5.2/html/ace/namespaceACE__OS.html#a521) and [`nanosleep()`](http://man7.org/linux/man-pages/man2/nanosleep.2.html) implementation/documentation the two calls activate the scheduler passing argument 0, and in particular ACE uses `select()`/`Sleep()`/`nanosecond()` for this purpose.
The C++11 implementation, however, _**seems not to have this behaviour at all**_.
Furthermore, trying to force a `yield()` seems not to solve the problem, in particular on **Linux** OSs (Windows and macOS seems not to be affected by this, behaving as expected).
@aerydna and @barbalberto made some tests that I run on different systems.

Concluding, what we have had so far is a _**particular behaviour**_: invoking the scheduler with a `sleep(0)`. This patch may not be principled or elegant or whatever, but it necessary to have the behaviour we always have had (and maybe that should actually be there).
Taking this into account, we should decide what we want for YARP.

drdanz:
For the records, I'm against this patch

barbalberto:
I believe this is not a good place to discuss about complicated issue, so I'll write only some short clarifications:
 > If we leave aside for a moment all the discussion we had related to the misuse of a mutex within a thread ...

Then all this patch does not solve anything, since the starvation happens only when both conditions are together. sleep(0) by his own is not a problem.

> For me, it's more reasonable that the end of the single instance of run() brings up the task-scheduler that decides which thread to wake up, instead of calling again run() straightaway.

This is an opinion, and I do care about user's experience, but the scheduler don't :smile: 
The work done by the scheduler is unrelated to how the code is written, it will give and take cpu time to threads based on his own algorithm and thread priority.
What will probably happen is that scheduler wakes up other threads *during* the run execution if that run takes a long time to complete. The starvation happens in cases where, when other threads are woken up, a shared mutex is still locked.

@claudiofantacci the code you are pointing to is the old master, the devel only has C++11 sleep. SleepThread is not used anymore in devel. It was so messy...

The test I run was showing that the sleep(0) by his own is not the critical difference, also in mac a sleep(0) does not calls yeld or rescheduling. What actually change between OSs is the order in which  threads are woken up after the mutex_unlock.

pattacini:
@barbalberto we have better clarified F2F the problem here 👍 

So the key point is how we intend our RateThread layer should work from a user's perspective; we're not dealing with the job of the scheduler.

src/libYARP_OS/include/yarp/os/SystemClock.h
```diff
@@ -30,6 +30,7 @@ class YARP_OS_API yarp::os::SystemClock : public Clock {
 
     static double nowSystem();
     static void delaySystem(double seconds);
+    static void yieldSystem();
```
drdanz: The `yieldSystem()` method is in my opinion wrong... there is no need for this in `SystemClock`, `yield` is not something related to `time` or to the clock, but to the execution of the thread, hence it should call `yarp::os::Thread::yield()` and `yarp::os::Time::yield()` should be deprecated...

drdanz: or even better, call the c++11 yield() function directly...

drdanz: damn, @traversaro you are fast!

barbalberto: Make sense. Done in the last version of the commit.

claudiofantacci:
@barbalberto asked me to run the 25th test, the RateThread one, on my mac.
My settings should be the same of test 5005.21, using clan 8.0.0, and I am able to pass it all the time.
I just restarted job 5005.3 which uses clang 7 and and job 5005.21 to see whether it was some kind of Travis error related to thread usage or not.

claudiofantacci:
Also, should we rename this PR in lieu of the actual changes that will be made in YARP?

claudiofantacci:
The test passed without errors.
Maybe test involving thread may be the cause of accidental crashes on Travis.

barbalberto:
I think next step is to test the branch on the robot

drdanz:
Merged, thanks.

PR #1426, opened 2017-10-24
pixelsize bug fix

### LINKED ISSUE #1408: ImageOf - FlexImage conversion on read
by sending a `yarp::sig::ImageOf<PixelRgb>` and receiving it as a `yarp::sig::FlexImage`, some of the metadata of the original image are lost. the metadata tested at the moment are `pixelsize` and `padding`.

for example code see #1409

aerydna:
fixes #1408

basically the pixelSize is a datum dependent on the pixel code and have to be always deduced from it, not set manually. we should deprecate the `setPixelSize` method

### DIFF
--- a/src/libYARP_sig/include/yarp/sig/Image.h
+++ b/src/libYARP_sig/include/yarp/sig/Image.h
@@ -267,7 +267,7 @@ public:
      * Returns IPL/OpenCV view of image, if possible.
      * Not possible if the image is the wrong size, with no padding.
      * This method is currently not well documented.
-     * @return pointer to an IplImage structure
+     * @return pointer to an IplImage structure or nullptr
      */
     void *getIplImage();
 
@@ -275,7 +275,7 @@ public:
      * Returns IPL/OpenCV view of image, if possible.
      * Not possible if the image is the wrong size, with no padding.
      * This method is currently not well documented.
-     * @return pointer to an IplImage structure
+     * @return pointer to an IplImage structure or nullptr
      */
     const void *getIplImage() const;
 
@@ -346,6 +346,9 @@ protected:
 
     void setPixelCode(int imgPixelCode);
 
+    //pixelCode and pixelsSize should be linked together consistently.
+    //since setPixelCode set also the corresponding pixelSize setPixelSize should not be used at all except for
+    //setting an arbitrary pixelSize with no corresponding pixel code (in that case the pixelCode will be set to -pixelSize).
     void setPixelSize(int imgPixelSize);
 
 
@@ -378,8 +381,12 @@ public:
         Image::setPixelCode(imgPixelCode);
     }
 
+
     void setPixelSize(int imgPixelSize) {
         Image::setPixelSize(imgPixelSize);
+    //pixelCode and pixelsSize should be linked together consistently.
+    //since setPixelCode set also the corresponding pixelSize setPixelSize should not be used at all except for
+    //setting an arbitrary pixelSize with no corresponding pixel code (in that case the pixelCode will be set to -pixelSize).
     }
 
     void setQuantum(int imgQuantum) {
--- a/src/libYARP_sig/include/yarp/sig/Image.h
+++ b/src/libYARP_sig/include/yarp/sig/Image.h
@@ -12,6 +12,7 @@
 #include <yarp/os/Vocab.h>
 #include <yarp/os/NetUint16.h>
 #include <yarp/sig/api.h>
+#include <map>
 
 namespace yarp {
     /**
@@ -35,6 +36,38 @@ namespace yarp {
     }
 }
 
+// the image types partially reflect the IPL image types.
+// There must be a pixel type for every ImageType entry.
+enum YarpVocabPixelTypesEnum
+{
+    VOCAB_PIXEL_INVALID = 0,
+    VOCAB_PIXEL_MONO = VOCAB4('m','o','n','o'),
+    VOCAB_PIXEL_MONO16 = VOCAB4('m','o','1','6'),
+    VOCAB_PIXEL_RGB = VOCAB3('r','g','b'),
+    VOCAB_PIXEL_RGBA = VOCAB4('r','g','b','a'),
+    VOCAB_PIXEL_BGRA = VOCAB4(98/*'b'*/,'g','r','a'), /* SWIG BUG */
+    VOCAB_PIXEL_INT = VOCAB3('i','n','t'),
+    VOCAB_PIXEL_HSV = VOCAB3('h','s','v'),
+    VOCAB_PIXEL_BGR = VOCAB3(98/*'b'*/,'g','r'), /* SWIG BUG */
+    VOCAB_PIXEL_MONO_SIGNED = VOCAB4('s','i','g','n'),
+    VOCAB_PIXEL_RGB_SIGNED = VOCAB4('r','g','b','-'),
+    VOCAB_PIXEL_RGB_INT = VOCAB4('r','g','b','i'),
+    VOCAB_PIXEL_MONO_FLOAT = VOCAB3('d','e','c'),
+    VOCAB_PIXEL_RGB_FLOAT = VOCAB4('r','g','b','.'),
+    VOCAB_PIXEL_HSV_FLOAT = VOCAB4('h','s','v','.'),
+    VOCAB_PIXEL_ENCODING_BAYER_GRBG8 = VOCAB4('g', 'r', 'b', 'g'),   //grbg8
+    VOCAB_PIXEL_ENCODING_BAYER_GRBG16 = VOCAB4('g', 'r', '1', '6'),  //grbg16
+    VOCAB_PIXEL_ENCODING_BAYER_BGGR8 = VOCAB4(98/*'b'*/, 'g', 'g', 'r'),     //bggr8
+    VOCAB_PIXEL_ENCODING_BAYER_BGGR16 = VOCAB4(98/*'b'*/, 'g', '1', '6'),  //bggr16
+    VOCAB_PIXEL_ENCODING_BAYER_GBRG8 = VOCAB4('g', 'b', 'r', 'g'),  //gbrg8
+    VOCAB_PIXEL_ENCODING_BAYER_GBRG16 = VOCAB4('g', 'b', '1', '6'),  //gbrg16
+    VOCAB_PIXEL_ENCODING_BAYER_RGGB8 = -VOCAB4('r', 'g', 'g', 'b'),   //rggb8
+    VOCAB_PIXEL_ENCODING_BAYER_RGGB16 = VOCAB4('r', 'g', '1', '6'),  //rggb16
+    VOCAB_PIXEL_YUV_420 = VOCAB4('y','u','v','a'),
+    VOCAB_PIXEL_YUV_444 = VOCAB4('y','u','v','b'),
+    VOCAB_PIXEL_YUV_422 = VOCAB4('y','u','v','c'),
+    VOCAB_PIXEL_YUV_411 = VOCAB4('y','u','v','d')
+};
 
 /**
  * \ingroup sig_class
@@ -317,7 +350,7 @@ protected:
 
 
 private:
-
+    static const std::map<YarpVocabPixelTypesEnum, unsigned int> pixelCode2Size;
     int imgWidth, imgHeight, imgPixelSize, imgRowSize, imgPixelCode, imgQuantum;
     bool topIsLow;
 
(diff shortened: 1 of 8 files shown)


src/libYARP_sig/include/yarp/sig/Image.h
```diff
@@ -35,6 +36,38 @@ namespace yarp {
     }
 }
 
+// the image types partially reflect the IPL image types.
```
randaz81: Uhm, maybe inside a namespace would be better

src/libYARP_sig/src/Image.cpp
```diff
@@ -609,7 +609,40 @@ int ImageStorage::_pad_bytes (int linesize, int align) const
 
 
 
+std::map<YarpVocabPixelTypesEnum, unsigned int> Image::create_map()
```
randaz81: check ImageStorage::_set_ipl_header()

src/libYARP_sig/include/yarp/sig/Image.h
drdanz: if (!restoreObject()) {
        unreferenceStorage();
    }

drdanz: I think this was already outside of the namespace and it was just moved above, but @aerydna please confirm this...

src/libYARP_sig/src/Image.cpp
```diff
@@ -609,7 +609,40 @@ int ImageStorage::_pad_bytes (int linesize, int align) const
...
+    m[VOCAB_PIXEL_ENCODING_BAYER_BGGR8]  = 4;
+    m[VOCAB_PIXEL_ENCODING_BAYER_BGGR16] = 8;
+    m[VOCAB_PIXEL_ENCODING_BAYER_GBRG8]  = 4;
+    m[VOCAB_PIXEL_ENCODING_BAYER_GBRG16] = 8;
+    m[VOCAB_PIXEL_ENCODING_BAYER_RGGB8]  = 4;
+    m[VOCAB_PIXEL_ENCODING_BAYER_RGGB16] = 8;
+    m[VOCAB_PIXEL_YUV_420]               = 3;
+    m[VOCAB_PIXEL_YUV_444]               = 3;
+    m[VOCAB_PIXEL_YUV_422]               = 3;
+    m[VOCAB_PIXEL_YUV_411]               = 3;
```
drdanz: please use `sizeof(<type>)` instead of constants, at least for the ones that correspond to a specific type.

src/libYARP_sig/include/yarp/sig/Image.h
```diff
@@ -345,9 +379,8 @@ class YARP_sig_API yarp::sig::FlexImage : public yarp::sig::Image {
         Image::setPixelCode(imgPixelCode);
     }
 
-    void setPixelSize(int imgPixelSize) {
-        Image::setPixelSize(imgPixelSize);
-    }
+    //should be deprecated since the pixelType have to be deduced from the pixelCode and not set manually..
```
drdanz: As discussed, I don't agree with this comment...

src/libYARP_sig/include/yarp/sig/Image.h
```diff
@@ -173,7 +206,7 @@ class YARP_sig_API yarp::sig::Image : public yarp::os::Portable {
      * @return address of pixel in memory
      */
     inline unsigned char *getPixelAddress(int x, int y) const {
-        return (unsigned char *)(data[y] + x*imgPixelSize);
+        return (unsigned char *)(data[y] + x*getPixelSize());
```
drdanz: Sorry, but I'm a bit afraid that not having this as a variable and calling this function instead, might have a heavy impact on user code performances... Just image a for loop calling `getPixelAddress()` for each pixel!

I'd rather restore the `imgPixelSize` variable and use it as a cache whenever the PixelCode is changed

tests/libYARP_sig/ImageTest.cpp
```diff
@@ -226,7 +226,6 @@ class ImageTest : public UnitTest {
                         yAssert(data);
                         img.setQuantum(pad1);
                         img.setPixelCode(VOCAB_PIXEL_RGB);
-                        img.setPixelSize(3);
```
drdanz: This should still work after your changes, if it is not working you are doing something wrong.
Please restore this line in the unit test...

src/libYARP_sig/src/Image.cpp
drdanz: Also we are full c++11 now, you should be initializing it using c++11 syntax, i.e. something like this:

```c++
const std::map<YarpVocabPixelTypesEnum, unsigned int> Image::pixelCode2Size = {
   {VOCAB_PIXEL_INVALID, 0},
   {VOCAB_PIXEL_MONO,    sizeof(yarp::sig::PixelMono)},
   {VOCAB_PIXEL_MONO16,  sizeof(yarp::sig::PixelMono16)},
// ... 
};

File: src/libYARP_sig/include/yarp/sig/Image.h (hunk 1 above)
Comment (aerydna) [reply src/libYARP_sig/include/yarp/sig/Image.h]: i confirm

File: src/libYARP_sig/src/Image.cpp [hunk 7]
```diff
@@ -126,7 +126,7 @@ class ImageStorage {
     }
 
     void resize(int x, int y, int pixel_type,
-                int pixel_size, int quantum, bool topIsLow);
+                int quantum, bool topIsLow);
```
Nicogene: I'm afraid that this function is largely used in the downstream project( `icub-main` for sure). We should leave the old signature deprecated and add this one.

src/libYARP_sig/src/Image.cpp
```diff
@@ -135,7 +135,7 @@ class ImageStorage {
 
 
 void ImageStorage::resize(int x, int y, int pixel_type,
-                          int pixel_size, int quantum, bool topIsLow) {
+                          int quantum, bool topIsLow) {
```
Nicogene: Same as above

src/libYARP_sig/src/Image.cpp
Nicogene: Forget it, I thought was `Image::resize` :sweat_smile:

drdanz: I think this method is hidden in the c++ file, so it cannot be used outside

src/libYARP_sig/src/Image.cpp
```diff
@@ -669,33 +707,30 @@ void Image::resize(int imgWidth, int imgHeight) {
...
     }
 }
 
+void Image::setPixelSize(int imgPixelSize) {
+    if(imgPixelSize == getPixelSize())
+        return;
 
-void Image::setPixelCode(int imgPixelCode) {
-    this->imgPixelCode = imgPixelCode;
+    return setPixelCode(-imgPixelSize);
```
Nicogene: ??

Nicogene:
Check `travis` and `appveyor` fails:
```
98% tests passed, 2 tests failed out of 85

Total Test time (real) =  74.29 sec

The following tests FAILED:

	 45 - sig::ImageTest (SEGFAULT)

	 46 - sig::SoundTest (SEGFAULT)
``` 
Why `Sound` and `Image` are linked ? :sweat_smile:

src/libYARP_sig/include/yarp/sig/Image.h
```diff
@@ -313,11 +346,12 @@ class YARP_sig_API yarp::sig::Image : public yarp::os::Portable {
 
     void setPixelCode(int imgPixelCode);
 
+    //should be deprecated since the pixelType have to be deduced from the pixelCode and not set manually..
```
drdanz: I still don't like this comment, please remove it

src/libYARP_sig/src/Image.cpp
drdanz: @Nicogene Don't worry, this is pure magic... 🤣 
@aerydna Please add some comments to explain this

src/libYARP_sig/include/yarp/sig/Image.h
```diff
@@ -139,7 +172,7 @@ class YARP_sig_API yarp::sig::Image : public yarp::os::Portable {
      */
     inline int getPadding() const
     {
-        const int ret=imgRowSize-imgWidth*imgPixelSize;
+        const int ret=imgRowSize-imgWidth*getPixelSize();
```
drdanz: is there a reason for not using `imgPixelSize` here?

src/libYARP_sig/include/yarp/sig/Image.h
```diff
@@ -313,11 +346,12 @@ class YARP_sig_API yarp::sig::Image : public yarp::os::Portable {
 
     void setPixelCode(int imgPixelCode);
 
+    //should be deprecated since the pixelType have to be deduced from the pixelCode and not set manually..
     void setPixelSize(int imgPixelSize);
 
-
 private:
-
+    static std::map<YarpVocabPixelTypesEnum, unsigned int> create_map();
```
drdanz: Is this method still useful? I think it is no longer implemented...

src/libYARP_sig/src/Image.cpp
```diff
@@ -657,7 +686,6 @@ void Image::resize(int imgWidth, int imgHeight) {
     yAssert(imgWidth>=0 && imgHeight>=0);
 
     int code = getPixelCode();
-    int size = getPixelSize();
     bool change = false;
     if (code!=imgPixelCode) {
         imgPixelCode = code;
```
drdanz: I think you should be setting `imgPixelSize` as well here

src/libYARP_sig/src/Image.cpp
```diff
@@ -637,7 +660,13 @@ Image::~Image() {
...
 
 int Image::getPixelSize() const {
-    return imgPixelSize;
+
+    if(pixelCode2Size.find((YarpVocabPixelTypesEnum)imgPixelCode) != pixelCode2Size.end())
+    {
+        return -imgPixelCode;
+    }
+
+    return pixelCode2Size.at((YarpVocabPixelTypesEnum)imgPixelCode);
```
drdanz: I don't like this...
The `getPixelSize` method should just return `imgPixelSize`. Instead this logic should be used when setting the pixelCode

src/libYARP_sig/src/Image.cpp
```diff
@@ -619,7 +642,7 @@ void Image::initialize() {
     implementation = nullptr;
     data = nullptr;
     imgWidth = imgHeight = 0;
-    imgPixelSize = imgRowSize = 0;
+    imgRowSize = 0;
```
drdanz: How does this build???

src/libYARP_sig/src/Image.cpp
aerydna: i think it is better to leave the getter method directly linked to the pixelCode value. for someone who read the code the binding between the two is clearer this way. The cached variable is used only in the getPixelAdress Loop.

src/libYARP_sig/src/Image.cpp
aerydna: again here i think it's better use the method (that is linked to the pixelSize property) for making clear the binding between the two. (someone who read this, by going to the implementation of getPixelSize() sees immediatly the binding)

src/libYARP_sig/src/Image.cpp
aerydna: ?

src/libYARP_sig/include/yarp/sig/Image.h
aerydna: see the replies to similar questions below

aerydna:
@drdanz it think it is better to leave the method where possible (where performances is not a problem) for it to show clearly the binding between the two properties (pixelCode and pixelSize)

it it for performance that you prefer the cached variable?

src/libYARP_sig/src/Image.cpp
```diff
@@ -604,12 +606,35 @@ int ImageStorage::_pad_bytes (int linesize, int align) const
...
+    {VOCAB_PIXEL_ENCODING_BAYER_BGGR8,  8 },
+    {VOCAB_PIXEL_ENCODING_BAYER_BGGR16, 16},
+    {VOCAB_PIXEL_ENCODING_BAYER_GBRG8,  8 },
+    {VOCAB_PIXEL_ENCODING_BAYER_GBRG16, 16},
+    {VOCAB_PIXEL_ENCODING_BAYER_RGGB8,  8 },
+    {VOCAB_PIXEL_ENCODING_BAYER_RGGB16, 16},
+    {VOCAB_PIXEL_YUV_420,               3 },
+    {VOCAB_PIXEL_YUV_444,               3 },
+    {VOCAB_PIXEL_YUV_422,               3 },
+    {VOCAB_PIXEL_YUV_411,               3 },
```
barbalberto: The sizes of pixels for bayer and yuv does not seem right to me. Where did you get those values?
I think they should be:
{VOCAB_PIXEL_ENCODING_BAYER_GRBG8,       1 },   // Values are in bytes, 8/16 were bits instead
{VOCAB_PIXEL_ENCODING_BAYER_GRBG16,     2},
{VOCAB_PIXEL_ENCODING_BAYER_BGGR8,       1 },
{VOCAB_PIXEL_ENCODING_BAYER_BGGR16,     2},
{VOCAB_PIXEL_ENCODING_BAYER_GBRG8,       1 },
{VOCAB_PIXEL_ENCODING_BAYER_GBRG16,     2},
{VOCAB_PIXEL_ENCODING_BAYER_RGGB8,       1 },
{VOCAB_PIXEL_ENCODING_BAYER_RGGB16,     2},

for yuv is a bit more tricky. The size should be:

{VOCAB_PIXEL_YUV_444,               3 },
{VOCAB_PIXEL_YUV_422,               2 },
{VOCAB_PIXEL_YUV_411,               1.5 },  --> how can we handle this? increase it to 2?
{VOCAB_PIXEL_YUV_420,               1.5},   --> how can we handle this? increase it to 2?  Is it ever been used?

aerydna: ok.. i fixed up the bayer ones.. it was just a slip. 

For the YUV ones i don't get the code..  if the numbers are bit used we would have 
VOCAB_PIXEL_YUV_444, 1.5
VOCAB_PIXEL_YUV_422, 1
VOCAB_PIXEL_YUV_411, 0.75
VOCAB_PIXEL_YUV_420, 0.75

barbalberto: Actually the code does not refers to bits, but to subsampling.
Here is a short video explaining them, for reminder
http://www.personal-view.com/talks/discussion/6144/easy-video-explaining-444-422-and-420-for-noobs/p1

src/libYARP_sig/src/Image.cpp
```diff
@@ -298,44 +297,42 @@ void ImageStorage::_set_ipl_header(int x, int y, int pixel_type, int quantum,
     int origin = topIsLow?IPL_ORIGIN_TL:IPL_ORIGIN_BL;
     int implemented_yet = 1;
     // used to allocate the ipl header.
+#define iplCreateImageHeaderMono   pImage = iplCreateImageHeader( 1, 0, IPL_DEPTH_8U,  (char *)"GRAY", (char *)"GRAY", IPL_DATA_ORDER_PIXEL, origin, quantum, x, y, nullptr, nullptr, nullptr, nullptr)
+#define iplCreateImageHeaderMono16 pImage = iplCreateImageHeader( 1, 0, IPL_DEPTH_16U, (char *)"GRAY", (char *)"GRAY", IPL_DATA_ORDER_PIXEL, origin, quantum, x, y, nullptr, nullptr, nullptr, nullptr)
```
drdanz: Please remove this macro, I don't want to see new macro magic added to a code that is already almost unreadable...

aerydna: no macro magic here.. it is simply the method already called many times in that portion of code.. in my case i've added 8 types that requires the method called with the same parameters (4 and 4).. are you sure the code to be more readable removing these macro?.. cause in this case i think it is time for refactoring..

barbalberto: I do agree with @drdanz, defines shall not have to be used to simply _rename_ things. That's a bad practise and does not help readability, please stop doing it. Let the code speak for itself

aerydna: in this case are used for not repeating code and connecting the same code under a common flag, to let the reader know that all the involved cases calls the same function.. and this concept is not as clear when you call a method with 14 parameters! but since you insist, lets refactor!

src/libYARP_sig/src/Image.cpp
```diff
@@ -717,7 +517,6 @@ void Image::synchronize() {
         data = nullptr;
         imgWidth = imgHeight = 0;
     }
-    imgPixelSize = getPixelSize();
```
drdanz: This removal does not look correct to me, but I'm not sure about what's happening here, can you please double check this part?

aerydna: now that the behaviour of getPixelSize is restored (it returns imgPixelSize) that call is equivalent to imgPixelSize = imgPixelSize.

aerydna:
i didn't really close the pr.. i pushed a totally different local branch (recommitted) and for some reason i found the pr closed

drdanz:
Merged (finally), thanks.

PR #1432, opened 2017-10-27
Small improvements in yarpmanager

Nicogene:
This PR introduces some small enhancements:

- If `<node>` and  `<protocol>` are not specified in the xml file, `localhost` and `tcp` will be added respectively while loading the application.
- Added ip details of the port nodes: 
image
- Remove duplication of resources, before if two(or more modules) had the same `<dependencies>` the resource was diplayed twice:
image

Please review code.


drdanz:
Merged, thanks.

PR #1449, opened 2017-11-06
Improve using of parameters in carriers

valegagge:
Currently, if source or destination port is not competent (ie it can't perform yarp handshaking), carrier's paramters given in connect command are lost.

Since could be usefull to specify some resize paramters for h264 carrier in connect command, I needed to change how the parameters are passed to the carrier.

now if user gives following commands:

```
yarp name register /srcPort MyCarrier+param1.value1 IP PORTNUM
yarp name register /dstPort MyCarrier+param2.value2 IP PORTNUM
yarp connect /srcPort /dstPort MyCarrier+param3.value3
```

The used carrier contains all three parameters: +param1.value1+param2.value2+param3.value3.

We decided that Network object does NOT any check on parameters: their verification will be done in the carrier.

### DIFF
--- a/src/libYARP_OS/src/Network.cpp
+++ b/src/libYARP_OS/src/Network.cpp
@@ -126,6 +126,14 @@ static int noteDud(const Contact& src)
     return ok ? 0 : 1;
  }
 
+
+//#define DEBUG_CONNECT_CARRIER
+#ifdef DEBUG_CONNECT_CARRIER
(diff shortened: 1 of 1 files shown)


drdanz:
Merged, thanks.

drdanz:
@valegagge I forgot, can you please add a line to the changelog briefly explaining this change?

PR #1451, opened 2017-11-08
Change YARP CMake file destination on Linux/macOs to "lib/cmake/YARP"

### LINKED ISSUE #1081: Same folder used for plugins and CMake config files
Given that

1. CMake config files are installed in `<prefix>/lib/YARP`
2. Plugin libraries are installed in `<prefix>/lib/yarp`
3. On macOS, the file system is case insensitive.

This means that on macOS both CMake files and plugin libraries are installed in the same folder.

This is not optimal. Also on case sensitive file system it is not really nice to have `/usr/lib/YARP` and `/usr/lib/yarp`

I suggest to change CMake folder to one of:

* `<prefix>/lib/cmake/YARP`
* `<prefix>/share/yarp/cmake` (this folder is already used for CMake modules installed by yarp)
* `<prefix>/share/yarp` (this is the folder used by vcpkg on windows)

Also, since CMake v3.7.0, on windows, the same paths are supported, therefore we might consider changing the folder on windows as well (at the moment `<prefix>/cmake`), maybe adding a check on CMake version.

Anyone against this change?

cc @francesco-romano, @lornat75, @traversaro

cc also @mbrunettini since this change will have an impact on packaging

drdanz:
Fixes #1081 

As discussed with @mbrunettini, we will merge this when the release scripts are updated as well.

@francesco-romano can you check if brew formulas need to be updated?


francesco-romano:
@drdanz the formula should be ok

drdanz:
@mbrunettini I had a look at [robotology/yarp-packaging](https://github.com/robotology/yarp-packaging) and since this patch only affects linux, I think it is ok to merge it, since I don't see anything related to these files in linux scripts. Can you please confirm?

PR #1455, opened 2017-11-17
ServerGrabberDual: fix error while handling RPC calls for right image

barbalberto:
RPC were not working when connectiong to the right image port, in case split is set to true.


Nicogene:
@barbalberto `TestFrameGrabberTest` with valgrind is failing, can you please check it ?

barbalberto:
Yep. Also bindings are not happy...

drdanz:
Merged, thanks.

PR #1459, opened 2017-11-20
Fix web example issues.

jgvictores:
- Contact(name) -> Contact contact(name)
- ConstString::toString -> std::stringstream implementation


drdanz:
The test failed is unrelated... merging

drdanz:
Merged, thanks.

PR #1467, opened 2017-11-24
Coverity fixes - initializations_3

damn1:
Fixes some 'uninitialized pointer or scalar field' issues raised by Coverity.
Affect components
 - YARP_manager
 - yarpmanager

### DIFF
--- a/src/yarpmanager/src-builder/arrow.cpp
+++ b/src/yarpmanager/src-builder/arrow.cpp
@@ -57,35 +57,29 @@ const qreal Pi = 3.14;
 
 
 
-Arrow::Arrow(BuilderItem *startItem, BuilderItem *endItem,int id, Manager *safeManager,
-             bool isInApp,bool editingMode, BuilderItem *parent)
-    : BuilderItem(parent),textLbl("",this)
+Arrow::Arrow(BuilderItem *startItem, BuilderItem *endItem,int _id, Manager *safeManager,
+             bool isInApp,bool editingMode, BuilderItem *parent) :
+    BuilderItem(parent),
+    manager(safeManager),
+    externalSelection(false),
+    editingMode(editingMode),
+    connected(false),
+    myStartItem(startItem),
+    myEndItem(endItem),
+    myColor(Qt::black),
+    textLbl("",this),
+    textWidth(0),
+    id(_id),
+    firstTime(true)
 {
-    sigHandler = new ItemSignalHandler();
     itemType = ConnectionItemType;
-    connected = false;
-    myStartItem = startItem;
-    myEndItem = endItem;
-    this->manager = safeManager;
+    nestedInApp = isInApp;
+    sigHandler = new ItemSignalHandler();
     setFlag(ItemIsSelectable,true);
     setFlag(ItemClipsToShape,false);
-    myColor = Qt::black;
-    this->id = id;
-    this->nestedInApp = isInApp;
-    this->editingMode = editingMode;
-
-
-    externalSelection = false;
-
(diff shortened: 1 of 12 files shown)


src/yarpmanager/src-manager/logwidget.cpp
```diff
@@ -3,7 +3,8 @@
 #include <QCoreApplication>
 
 LogWidget::LogWidget(QWidget *parent) :
-    QListWidget(parent)
+    QListWidget(parent),
+    contextMenu(nullptr)
```
Nicogene: In `master` you have to put `YARP_NULLPTR`

src/yarpmanager/src-manager/safe_manager.cpp
```diff
@@ -16,9 +16,12 @@ using namespace yarp::manager;
 #define POST_SEMAPHOR() postSemaphore();
 
 
-SafeManager::SafeManager()
-    :action(MNOTHING), eventReceiver(NULL), busyAction(false)
-{ }
+SafeManager::SafeManager() :
+    m_pConfig(YARP_NULLPTR),
+    action(MNOTHING),
+    eventReceiver(NULL),
```
Nicogene: Please use `YARP_NULLPTR` for making easier search and replace when merging `master` in `devel`

src/yarpmanager/src-manager/logwidget.cpp
damn1: Here using `YARP_NULLPTR` was giving an 'undefined symbol' error

damn1: Ok I included `compiler.h` to avoid using `nullptr` https://github.com/robotology/yarp/pull/1467/commits/6ae143d8afb84690ad94174b1de1ee659eae4fc6

src/yarpmanager/src-builder/arrow.cpp
```diff
@@ -57,35 +57,29 @@ const qreal Pi = 3.14;
...
+    firstTime(true)
 {
-    sigHandler = new ItemSignalHandler();
     itemType = ConnectionItemType;
-    connected = false;
-    myStartItem = startItem;
-    myEndItem = endItem;
-    this->manager = safeManager;
+    nestedInApp = isInApp;
+    sigHandler = new ItemSignalHandler();
```
drdanz: Is there a reason why you didn't move these in the initialization list as well?

damn1: Yes they're base class members and i could not init them in the list.

src/yarpmanager/src-builder/arrow.cpp
```diff
@@ -57,35 +57,29 @@ const qreal Pi = 3.14;
 
 
 
-Arrow::Arrow(BuilderItem *startItem, BuilderItem *endItem,int id, Manager *safeManager,
-             bool isInApp,bool editingMode, BuilderItem *parent)
-    : BuilderItem(parent),textLbl("",this)
+Arrow::Arrow(BuilderItem *startItem, BuilderItem *endItem,int _id, Manager *safeManager,
```
drdanz: I didn't notice this kind of changes, and I merged it already, but there is no need to rename the `id` variable, you can safely initialize it as `id(id)`

damn1: Yes I know but I saw somewhere in yarp the `id(_id)` style and it seemed more clear to me so I adopted it. If you prefer the other one I will keep in mind.

drdanz:
Merged, thanks.

PR #1473, opened 2017-11-28
"--help" parameter opens actual device

aerydna:
at the moment when a device have a wrapper associated it's impossible to simply see the common help output (the parameter is parsed by the wrapper). this pr fixes this by opening the actual device without automatically wrapping it if the keyword "--help" is provided.

note that if the device doesn't take care of the "help" parameter it will start normally, so this could also be used as a (unwanted) workaround to bypass the wrapping system of yarpdev.

### DIFF
--- a/src/libYARP_dev/src/PolyDriver.cpp
+++ b/src/libYARP_dev/src/PolyDriver.cpp
@@ -239,7 +239,7 @@ bool PolyDriver::coreOpen(yarp::os::Searchable& prop) {
                 p.fromString(config->toString());
                 p.unput("wrapped");
                 config = &p;
-                if (wrapCreator!=creator) {
+                if (wrapCreator!=creator && !p.check("help")) {
                     p.put("subdevice",str.c_str());
                     p.put("device",wrapper.c_str());
                     p.setMonitor(prop.getMonitor(),


drdanz:
I agree that there is some confusion... at the moment:

* `yarpdev --help` prints yarpdev help (using yInfo) and exits with this warning: "[WARNING]--device option not specified. Closing."
* `yarpdev --help --device test_grabber` starts test_grabber
* `yarpdev --verbose --device test_grabber` prints all the parameters supported by test_grabber and starts the test_grabber

Can you explain how this will change using your patch?

--------

`<rant>`

By the way, at the moment, the documentation is added using inside the `open()` method one of the overloads of the `check()` method.

In my opinion this is "an useless complication" (read "madness"), the only advantage is that documentation cannot be outdated, but this goes against const correctness (the Searchable passed in the open should be const, but it cannot be, because it is modified by the `check()` calls, it has a big impact on readability of the code, and (i think) this is prone to errors. For example I think that if you have 2 dependent parameters you might miss one, e.g.

```c++
if (config.check("foo", "foo doc") {
    // ...
    if (config.check("bar", "bar doc") {
        // ...
    }
}
```

`bar` documentation is probably not printed by --verbose, unless `--foo` is used

Moreover, Searchable code is a lot more complicated than it should be (SearchReport, SearchMonitor) due to this corner case usage.

`</rant>`

aerydna:
i agree the situation to be really confusing. This deserve a F2F discussion i think. @drdanz @barbalberto  @Nicogene @randaz81

Nicogene:
What we want to do with this PR?

aerydna:
we should discuss it. I'll put it in the meeting topics to discuss in our next meeting.

drdanz:
Closing as discussed

PR #1482, opened 2017-12-04
Rate thread improvements

### LINKED ISSUE #1481: Methods that should be protected in RateThread
The following methods in `yarp::os::RateThread` are supposed to be `protected`, but are `public`instead:

```c++
virtual bool threadInit();
virtual void threadRelease();
virtual void run() = 0;
virtual void beforeStart();
virtual void afterStart(bool success);
```

I'd like to make them `protected` but this is an API change and I don't know if this will break stuff around.
@traversaro @pattacini @lornat75 @barbalberto @vtikha are you aware of any module, device or other, actually *calling* these methods? (Just *overloading* them is ok, even though I would recommend to make them protected also in derived classes)

drdanz:
OS/RateThread: Improvements

* Cleanup
* Avoid useless call
* Improve PIMPL implementation
* Remove useless private method

OS/RateThread: Protect methods that should not be public.

The following methods in `yarp::os::RateThread` are now `protected` instead
of `public`:
* `virtual bool threadInit();`
* `virtual void threadRelease();`
* `virtual void run() = 0;`
* `virtual void beforeStart();`
* `virtual void afterStart(bool success);`

Fixes #1481

### DIFF
--- a/doc/release/v2_3_72.md
+++ b/doc/release/v2_3_72.md
@@ -35,6 +35,13 @@ Important Changes
   `yarp::os::Network` is initialized will abort the program in order to avoid
   undefined behaviour of uninitialized `Time` object
 * `Contactable::open(void)` was deprecated. Use `open("...")` instead.
+* The following methods in `yarp::os::RateThread` are now `protected` instead
+  of `public`:
+  * `virtual bool threadInit();`
+  * `virtual void threadRelease();`
+  * `virtual void run() = 0;`
+  * `virtual void beforeStart();`
+  * `virtual void afterStart(bool success);`
 
 
 New Features
--- a/src/libYARP_OS/include/yarp/os/RateThread.h
+++ b/src/libYARP_OS/include/yarp/os/RateThread.h
@@ -22,7 +22,8 @@ namespace yarp {
  *
  * An abstraction for a periodic thread.
  */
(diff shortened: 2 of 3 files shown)


pattacini:
@Nicogene you know how much concerned I am about breaking compatibility downstream 😉 
However, this case is very particular and I feel that calling directly those methods is deeply erratic per se.

Thus far, we didn't identify problems around. If we'll come across one, we'll fix it.

claudiofantacci:
> I totally agree to change these functions from public to protected, but here we are changing wide used header. I know, it is wrong to call directly run threadInit etc. but unfortunately we have to support also bad user implementations 😞

I agree only partially with you. As discussed several times before this issue, we should also care to have good public APIs.
To me having those interfaces publics is both conter-intuitive and most of all error prone.

What we should do, for example, is to try this PR on our systems to see whether there are issues or bad implemented applications. I only see positive outcomes with this patch.

Nicogene:
I know for sure that one thing will be broken after this merge but I can't tell because *he says the sin but not the sinner* 😛 Not a big problem this guy will fix it, I was just wandering if there are not other that did the same mistake ! 
I totally agree to give a good API to enforce users to write code in the right way ! 😉

drdanz:
Rebased over latest devel (travis should be fixed on devel).

PR #1500, opened 2017-12-20
yarpmanager-console: display correct name in help

gsaponaro:
Use the name `yarpmanager-console` consistently in the help message.


drdanz:
Merged, thanks.

PR #1503, opened 2017-12-24
Add missing <cstring> include in yarp/dev/CanBusInterface.h for std::memset

PeterBowman:
P.S. No changelog file available in `doc/release/`.


drdanz:
Merged, thanks.

PR #1505, opened 2017-12-24
Overload yarp::dev::CanMessage::operator[], add constness for getters

PeterBowman:
Use case:

```c++
// implements yarp::dev::ICanBus::canWrite
bool MyClass::canWrite(const yarp::dev::CanBuffer &msgs, unsigned int size, unsigned int *sent, bool wait)
{
    // ideally, we would like to avoid this:
    //const yarp::dev::CanMessage &msg = const_cast<yarp::dev::CanBuffer &>(msgs)[0];

    // ...and do this instead:
    const yarp::dev::CanMessage &msg = msgs[0];

    // send msg
}
```

### DIFF
--- a/doc/release/v2_3_72.md
+++ b/doc/release/v2_3_72.md
@@ -72,6 +72,7 @@ New Features
 #### `YARP_dev`
 
 * Added a new interface for visual servoing: `IVisualServoing.h`.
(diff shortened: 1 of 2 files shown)


drdanz:
Merged, thanks.

PR #1507, opened 2017-12-29
Add support for carrier specification on registration

damn1:
This patch should allow to open a port with an associated custom carrier different from tcp, if this carrier is available and it provides a custom implementation of `yarp::os::Face` that can replace default handshake.
Example:
```
yarp::os::Port p;
yarp::os::Contact c("/port", "carrier");
p.open(c);
```

### DIFF
--- a/src/libYARP_OS/src/Network.cpp
+++ b/src/libYARP_OS/src/Network.cpp
@@ -232,11 +232,6 @@ static int enactConnection(const Contact& src,
     if (c2.getPort()<=0) {
         c2 = NetworkBase::queryName(c2.getName());
     }
-    if (c2.getCarrier()!="tcp") {
-        YARP_SPRINTF2(Logger::get(), debug, "would have asked %s: %s",
-                      src.toString().c_str(), cmd.toString().c_str());
-        return 1;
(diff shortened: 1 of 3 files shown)


Nicogene:
I would add in `PortTest` or in `NetworkTest` the test that checks this patch
- Create a `Conctact` with carrier `tcp` and check that the port with this contact is opened
- Create a `Conctact` **without** carrier  and check that the port with this contact is opened
- Create a `Conctact` with carrier `udp` and check that the port with this contact is **NOT** opened as expected

BTW, nice patch!

src/libYARP_OS/src/Network.cpp
```diff
@@ -1589,7 +1584,7 @@ class StubCarrier : public ForwardingCarrier {
             ncar = nullptr;
             return nullptr;
         }
-        return ncar;
```
Nicogene: The only possible breaking change could be this one..
The motivation behind this change is that `ncar` is a sort of wrapper of the carriers that are loaded as plugin, returning `->getContent()` we have the actual carrier that has been loaded and not his wrapper.

PR #1520, opened 2018-01-23
Few fixes on RGBD devices and add documentation

barbalberto:
This PR fixes in the RGBD devices
- RPC responder in the wrapper
- timestamp segfault in the fakeDepthCamera
- add usage and code examples in the /example/dev/rgbd folder


src/devices/depthCamera/depthCameraDriver.h
```diff
@@ -46,12 +46,11 @@ namespace yarp
...
- *  @ingroup dev_impl
- *
+ *  @ingroup dev_impl_media
  *
  *
  * This device is a YARP plugin for OpenNI2 compatible devices, and exposes the IRGBDSensor and IFrameGrabberControls2
- * interfaces to read the images and operate on available settings.
+ * interfaces to read the images and operate on the available settings.
  *
  * See they documentation for more details about each interface.
```
gsaponaro: Hi @barbalberto, 'they' looks like a small typo here.

gsaponaro: I was referring to "See they documentation"

barbalberto: :+1: thanks

.gitignore
```diff
@@ -3,6 +3,8 @@
 /packaging/windows/build/
 /packaging/linux/build/
 /bindings/build*/
+/example/dev/build*/
```
Nicogene: we should add to gitignore all the build directories of the examples but personally I didn't find a smart way to do it.

example/dev/RGBD/README.md
```diff
@@ -0,0 +1,137 @@
...
+cmake .. -DCREATE_GUIS=true -DCREATE_OPTIONAL_CARRIERS=true -DENABLE_yarpcar_portmonitor=true -DENABLE_yarpcar_depthimage=true -DENABLE_yarpmod_fakeDepthCamera=true
+```
+
+To fully run the test, we need to open 5 terminals for command line input.
+
+First we need to run the yarpserver and the device driver producing the images. In this example we will use a fake device, for testing purposes.
+
+```
+Terminal 1: yarpserver
+Terminal 2: yarpdev --device RGBDSensorWrapper --name /server --subdevice fakeDepthCamera --mode ball
```
jgvictores: Command could be shorter if we appended `wrapper RGBDSensorWrapper` to `src/devices/fakeDepthCamera/fakeDepthCamera.ini` (I think touching `src/devices/fakeDepthCamera/CMakeLists.txt` was for static libs). Would also be nicer in general.

src/devices/fakeDepthCamera/fakeDepthCameraDriver.cpp
```diff
@@ -40,6 +40,8 @@ fakeDepthCameraDriver::~fakeDepthCameraDriver()
 bool fakeDepthCameraDriver::open(Searchable& config)
 {
     Property cfg;
+    cfg.fromString(config.toString());
```
jgvictores: Thanks for this! Had to hard-code things here for debugging a couple of days ago (bug turned out to be on my side).

example/dev/RGBD/README.md
barbalberto: Yes, you are right, I'll add it.
Anyway in this specific case I put everything on purpose because it is an example, so I prefer to explicitly call every device involved. This helps to understand how and way devices are created.

jgvictores: Understood and agreed! Thank you!!

.gitignore
barbalberto: IMHO the best would be to integrate example compilation in the main CMakeLists, so that examples will go in the main build folder along with all other stuff.

PR #1523, opened 2018-01-25
OS/Bottle: make operator== and operator!= const

drdanz:


src/libYARP_OS/src/Bottle.cpp
```diff
@@ -290,7 +290,7 @@ void Bottle::fini()
     }
 }
 
-bool Bottle::operator==(const Bottle& alt)
+bool Bottle::operator==(const Bottle& alt) const
 {
     return ConstString(toString().c_str()) == alt.toString().c_str();
```
drdanz: BTW, I doubt that this implementation still make sense, I think that `toString() == alt.toString()` will just work avoiding the creation of at least 2 temporary `ConstString`s

PR #1525, opened 2018-01-29
[WIP]Log In Yarpmanager

aerydna:
this pr intend to show the little effort necessary to integrate a log retriever into yarp manager. I've integrated it in the tooltip (that currently shows the last 10 lines of log) as an example, despite it to fullfill already the necessity to quickly check the status of an application (ex. the reason an application failed to start).

from here on, we can discuss what to do. the possibility i thought of are:

1. leave this as less invasive as possible.. showing only the last log lines, just to have a quick feedback from the applications and leaving the deep log inspection to yarplogger (basically leaving it as it is now)

2. integrating some functions of yarp logger but keeping it light and simple, leaving the deep log inspection to yarplogger

3. integrate all the functions of yarplogger, leaving yarplogger usage to the setups where no yarp managers are involved (command line yarprun setups) 

tooltip

### DIFF
--- a/src/libYARP_OS/src/Run.cpp
+++ b/src/libYARP_OS/src/Run.cpp
@@ -1945,7 +1945,6 @@ int yarp::os::Run::executeCmdStdout(Bottle& msg, Bottle& result, yarp::os::Const
     stdout_startup_info.hStdOutput=GetStdHandle(STD_OUTPUT_HANDLE);
     stdout_startup_info.hStdInput=read_from_pipe_cmd_to_stdout;
     stdout_startup_info.dwFlags|=STARTF_USESTDHANDLES;
-
     BOOL bSuccess=CreateProcess(nullptr,  // command name
                                 (char*)(yarp::os::ConstString("yarprun --log ")+loggerName+yarp::os::ConstString(" --write ")+portName).c_str(), // command line
                                 nullptr,  // process security attributes
@@ -1956,7 +1955,7 @@ int yarp::os::Run::executeCmdStdout(Bottle& msg, Bottle& result, yarp::os::Const
                                 nullptr,  // use parent's current directory
                                 &stdout_startup_info,   // STARTUPINFO pointer
                                 &stdout_process_info);  // receives PROCESS_INFORMATION
-
+    
     if (!bSuccess)
     {
         yarp::os::ConstString strError=yarp::os::ConstString("ABORTED: server=")+mPortName
@@ -2120,6 +2119,7 @@ int yarp::os::Run::executeCmdStdout(Bottle& msg, Bottle& result, yarp::os::Const
                              +yarp::os::ConstString("\n");
 
     result.addString(out.c_str());
+    portName = portName + "/" + std::to_string(stdout_process_info.dwProcessId);
     result.addString(portName.c_str());
     fprintf(stderr, "%s", out.c_str());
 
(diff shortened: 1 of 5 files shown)


randaz81:
Can we discuss this activity next week?
I can see the label "Test on hardware - Required", so I think that you really needed it urgently. So, it's ok for me if you use it. 
But on Yarp I always thought to have something completely different, plugin-based (rqt-like), not a monolithic yarpmanager. I have to say that I'm also generally skeptical about tooltips (in general, not about your implementation). Sorry about this 😔. 
You know we already discussed manager-logger integration many times, but unfortunately, we always missed the time to bring this to reality. (again...😔). I would happy to recap some ideas with you and all the group and to make a plan about this work as soon as I'm back in Genoa.

pattacini:
Don't want to debate now the choice between integrating everything inside yarpmanager vs. keeping yarplogger standalone, but I'm wondering what is the rationale behind this PR, which introduces apparently a very minimal subset of the features exposed by yarplogger.

Nicogene:
I don't think that we need it urgently, the "Test on hardware" label is only to clarify that we need to test it on a `iCub` setup and see how it behaves when you have an application with a lot of modules(e.g. `iol`).

aerydna:
actually i don't need this urgently. it is something i developed in the weekend to establish a simple connection between yarp manager and application logs (in this pr every application stores his log). i've put it in the tooltip to show that but from this point we can manage the log the way we prefer. this is why the pr is WIP (it will probably change the name over time). 

it's a way to say "ok we can have the log inside yarp manager with a very little effort. what can we do with it?".

traversaro:
Related to: https://github.com/robotology/yarp/issues/731 and https://github.com/robotology/yarp/issues/1527 .

aerydna:
@Nicogene the choice of limiting the lines to 9 was made on pourpose (its usage could be to inspect why an application as failed.. and for some cases this is the main and only reason for opening the logger). the effort for making a scrollable pop-up window is very little though.

pattacini:
@aerydna now we're talking the same language ;)
Thanks for amending the PR description: it was really necessary for who wasn't in the loop of the discussion to understand the motivations that pushed you to open it up.
Well done for this 👍

randaz81:
We agreed to discuss f2f, tentative date 06Feb

Nicogene:
I'm sorry, I will attend to the vvv-school from 6-8 February, shall we do it the 9??

aerydna:
@drdanz will not be present on 06Feb and probably also on 07Feb, so lets try on 09Feb @randaz81 @Nicogene @barbalberto @damn1

aerydna:
I took advantage of my insomnia to (while waiting for the f2f discussion) develop an accesorized  logger window widget. see gif below

managerlog

randaz81:
I'm really disappointed by the fact that you continued on working on this, despite the fact I explicitly told you twice that I do not like this work. I'm not going to merge any blobblish yarpmanger-logger integration, no matter how you change the widget appearance. I'll approve only a plugin-based approach, which I know is more complicated and requires more work, but I never asked you to do it (actually, I already mentioned that I'm inclined to implement it by myself). Additionally, I find despicable using sentences like "I did it in my spare time". To me, it just sounds like a way to force me to approve your PR. I discourage you to insist on this.

lornat75:
This PR has been put on hold in view of f2f discussion.

PR #1529, opened 2018-01-31
Coverity fix

damn1:
Some fixes for Coverity analyzer in 
* `libYARP_OS`
* `libYARP_dev`

### DIFF
--- a/src/libYARP_OS/src/DgramTwoWayStream.cpp
+++ b/src/libYARP_OS/src/DgramTwoWayStream.cpp
@@ -222,7 +222,9 @@ void DgramTwoWayStream::allocate(int readSize, int writeSize) {
                 int result = dgram->get_option(SOL_SOCKET, SO_RCVBUF, &_read_size, &len);
                 if (result < 0) {
                     YARP_ERROR(Logger::get(), ConstString("Failed to read buffer size from RCVBUF socket with error: ") +
-                               ConstString(strerror(errno)));
+                               ConstString(strerror(errno)) +
+                               ConstString(". Setting read buffer size to UDP_MAX_DATAGRAM_SIZE."));
+                    _read_size = UDP_MAX_DATAGRAM_SIZE;
                 }
             }
         #else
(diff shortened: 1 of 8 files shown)


src/libYARP_dev/src/MapGrid2D.cpp
```diff
@@ -1006,9 +1006,7 @@ bool MapGrid2D::getOccupancyData(XYCell cell, double& occupancy) const
         yError() << "Invalid cell requested " << cell.x << " " << cell.y;
         return false;
     }
-    if (m_map_occupancy.safePixel(cell.x, cell.y)==-1 ||
-        m_map_occupancy.safePixel(cell.x, cell.y)==255
```
barbalberto: @randaz81 If I see correctly, the CellData is a PixelMono, which in turns is a typedef of unsigned char, so it cannot be `-1`. Coverity suggests to remove the check because it is never false.
Maybe the purpose of the check was different ... can you double check this change?

src/libYARP_OS/src/DgramTwoWayStream.cpp
```diff
@@ -222,7 +222,9 @@ void DgramTwoWayStream::allocate(int readSize, int writeSize) {
                 int result = dgram->get_option(SOL_SOCKET, SO_RCVBUF, &_read_size, &len);
                 if (result < 0) {
                     YARP_ERROR(Logger::get(), ConstString("Failed to read buffer size from RCVBUF socket with error: ") +
-                               ConstString(strerror(errno)));
+                               ConstString(strerror(errno)) +
+                               ConstString(". Setting read buffer size to UDP_MAX_DATAGRAM_SIZE."));
+                    _read_size = UDP_MAX_DATAGRAM_SIZE;
```
Nicogene: The only doubt is about this assignment, should be better to put it to 0?? Anyway it is quite impossible that `result < 0` so it is not a big deal.

barbalberto: The ratio here is that `_read_size` is used afterwards to allocate memory.
Allocating zero memory will lead to an unusable buffer, while the client code expects something usable, but the API does not allow to notify the user that something went wrong.

Few lines above, at 211, a similar solution has been adopted for `_write_size`
```
if (_write_size > UDP_MAX_DATAGRAM_SIZE || _write_size < 0)
{
         _write_size = UDP_MAX_DATAGRAM_SIZE;
}
```
so I think it is safe to adopt for `_read_size` as well. Anyway I agree with you that it is a corner case that should never happen (and I believe never was).

Nicogene: All this stuff has been added by me in this [commit](https://github.com/robotology/yarp/commit/ed59342b6e96c973c80e7ba7cc6ce375b5df8a33) and I was uncertain how handle this error 😅 
BTW I agree to `_read_size = UDP_MAX_DATAGRAM_SIZE`  for symmetry 👍

src/libYARP_dev/src/MapGrid2D.cpp
randaz81: ok!

drdanz:
Merged, thanks

PR #1530, opened 2018-02-02
fixed YarpPlugin cmake macro

aerydna:
the macro YARP_ADD_PLUGIN_YARPDEV_EXECUTABLE was configuring a cpp file using the c++ macro YARP_DECLARE_DEVICES (which was deprecated and has been recently cleanupped) causing any project using it to not compile succesfully


drdanz:
Merged, thanks.

PR #1553, opened 2018-02-20
Re-license as BSD-3-Clause YARP libraries and several tools

drdanz:
@lornat75 please check (not every file, but at least each commit).


Nicogene:
@drdanz you didn't change the README.md, and the license files... is it intentional ?

drdanz:
@Nicogene I just added the `LICENSE` and `README.md` files in the last commits.

drdanz:
I'm about to merge this huge change.

Please note that I did my best and double checked everything, but the number of files is huge (1862 files modified), therefore I might have missed something.

If you notice any mistake, please don't sue me, but instead let me know, and I'll be happy to fix it... :stuck_out_tongue_closed_eyes:

PR #1563, opened 2018-02-23
Fix server expectText stuck

damn1:
Related to #1542 this could be a less invasive fix than placing a default timeout on the stream.
Unless we want to go for https://github.com/robotology/yarp/issues/1542#issuecomment-368011127, let me know.
On the fly I also moved [StreamConnectionReader] implementation from h to cpp.

### DIFF
--- a/src/libYARP_OS/src/StreamConnectionReader.cpp
+++ b/src/libYARP_OS/src/StreamConnectionReader.cpp
@@ -247,7 +247,9 @@ yarp::os::ConstString StreamConnectionReader::expectText(int terminatingChar)
     }
     yAssert(in!=nullptr);
     bool lsuccess = false;
+    in->setReadTimeout(2.0);
     ConstString result = in->readLine(terminatingChar, &lsuccess);
+    in->setReadTimeout(0.0);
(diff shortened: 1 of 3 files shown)


src/libYARP_OS/src/StreamConnectionReader.cpp
```diff
@@ -247,7 +247,9 @@ yarp::os::ConstString StreamConnectionReader::expectText(int terminatingChar)
     }
     yAssert(in!=nullptr);
     bool lsuccess = false;
+    in->setReadTimeout(2.0);
     ConstString result = in->readLine(terminatingChar, &lsuccess);
+    in->setReadTimeout(0.0);
```
drdanz: We should probably get current timeout and reset the previous value

damn1: Right, I didn't think about that, but now I checked and there is no easy way to get that: no `getReadTimeout()` or `haveReadTimeout()` method in the `InputStream`. Do you think is this the case to add one?

drdanz:
Cherry-picked the first commit

PR #1583, opened 2018-03-11
Fix subdevice not being created for BatteryWrapper

David-Estevez:
Hi!

I've been developing a new IBattery device but couldn't get it to work properly. After a closer inspection to the YARP code, I found out that the code for the BatteryWrapper was not creating the corresponding subdevice, and that's why I couldn't get my device to work. 

I've changed the BatteryWrapper based on other wrappers (such as the AnalogWrapper or the ControlBoardWrapper), fixing the issue, and now it works.

This PR contains the aforementioned fix.

### DIFF
--- a/src/devices/BatteryWrapper/BatteryWrapper.cpp
+++ b/src/devices/BatteryWrapper/BatteryWrapper.cpp
@@ -36,6 +36,7 @@ BatteryWrapper::BatteryWrapper() : RateThread(DEFAULT_THREAD_PERIOD)
 
 BatteryWrapper::~BatteryWrapper()
 {
+    threadRelease();
     battery_p = nullptr;
 }
 
--- a/src/devices/BatteryWrapper/BatteryWrapper.cpp
+++ b/src/devices/BatteryWrapper/BatteryWrapper.cpp
@@ -1,6 +1,6 @@
 /*
  * Copyright (C) 2013 Istituto Italiano di Tecnologia (IIT)
- * Authors: Marco Randazzo <marco.randazzo@iit.it>
+ * Authors: Marco Randazzo <marco.randazzo@iit.it>, David Estevez <destevez@ing.uc3m.es>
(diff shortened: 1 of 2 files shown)


Nicogene:
I'm not able to review this PR, @randaz81, @barbalberto, @traversaro any comment ??

src/devices/BatteryWrapper/BatteryWrapper.cpp
```diff
@@ -135,17 +141,48 @@ bool BatteryWrapper::read(yarp::os::ConnectionReader& connection)
     return true;
 }
 
-bool BatteryWrapper::threadInit()
```
Nicogene: Why did you removed the `threadInit` function?

src/devices/BatteryWrapper/BatteryWrapper.cpp
```diff
@@ -36,7 +36,6 @@ BatteryWrapper::BatteryWrapper() : RateThread(DEFAULT_THREAD_PERIOD)
 
 BatteryWrapper::~BatteryWrapper()
 {
-    threadRelease();
```
Nicogene: This function gave any problem?

drdanz:
> I'm not able to review this PR

Same here, ping @randaz81

David-Estevez:
Hi! Sorry for the late reply, I've been a bit busy during the last week. Let me answer both questions with this comment.

I removed `threadInit()` and `threadRelease()` because, as opposed to `AnalogWrapper`, the `BatteryWrapper` didn't need to extend their funcionality, so the implementation from the parent class was just fine. 

Removing the call to `threadRelase()` in the destructor was a lapse though, as it was supposed to be called anyways. I'm fixing that ASAP.

drdanz:
@randaz81 ping

drdanz:
@randaz81 ping

randaz81:
Suggested modifications included in #1916

PR #1590, opened 2018-03-14
Bugfix/yarpscope

prashanthr05:
Fix to handle bottles properly while displaying multiple indeces data on the yarpscope


pattacini:
@prashanthr05 I've edited the PR to merge into `devel`, which I think was your original intention 😉

PR #1593, opened 2018-03-14
Make shmem carrier a plugin

drdanz:
The `shmem` carrier is not strictly necessary in YARP_OS, moreover it is not built when ACE is not available.

This patch makes it a plugin (enabled by default), from the user point of view it doesn't change anything.

Also the code was cleaned, implementation moved in the proper files, and the old code commented out for several years (i.e. version 1 of the shmem carrier) was removed.


drdanz:
@Nicogene It's not really used anywhere, afaik...

drdanz:
And no, there are no unit tests

PR #1596, opened 2018-03-15
Fix rateThread example

barbalberto:
Fix segfault in the rateThread example

### DIFF
--- a/example/os/rateThreadTiming.cpp
+++ b/example/os/rateThreadTiming.cpp
@@ -11,7 +11,7 @@
 #include <yarp/os/RateThread.h>
 #include <yarp/os/Time.h>
 #include <yarp/os/Thread.h>
-
+#include <yarp/os/Network.h>
 #include <yarp/sig/Matrix.h>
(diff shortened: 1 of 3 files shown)


jgvictores:
Cool, tested and works.

@barbalberto Could you please apply the same on [example/os/threads.cpp](https://github.com/robotology/yarp/blob/45ec3a37ed717592bc512638be6073e26a23fd0b/example/os/threads.cpp)?
(I do not have push permissions; while forking for this is overkill, it is required to patch this to avoid segfaults)

drdanz:
@barbalberto Let me know if you want me to merge this, or if you want to fix example/os/threads.cpp as well, as suggested by @jgvictores

barbalberto:
ok, I'll fix also thread

barbalberto:
Updated all `os` examples

jgvictores:
@barbalberto Thanks a lot!

drdanz:
@barbalberto Awesome, thanks!

PR #1597, opened 2018-03-15
[WIP] PointCloud in YARP

Nicogene:
This PR introduces the `PointCloud` type in YARP :tada: 

A nice documentation page has been added under "YARP tutorials"

image

Probably the most important feature is the **total** compatibility with the PCL PointCloud.
In this sense in the documentation has been added a section where is explained how to convert a yarp cloud in a pcl cloud and viceversa.

image

Please review code.


src/libYARP_pcl/include/yarp/pcl/Pcl.h
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Convert a yarp::sig::PointCloud to a pcl::PointCloud object
+ * @param yarpCloud yarp::sig::PointCloud input
+ * @return a pcl PointCloud filled with data contained in the yarp cloud.
+ */
+template< class T1, class T2 >
+inline bool toPCL(yarp::sig::PointCloud< T1 > &yarpCloud, ::pcl::PointCloud< T2 > &pclCloud)
+{
+    yAssert(sizeof(T1) == sizeof(T2));
```
traversaro: [`static_assert`](http://en.cppreference.com/w/cpp/language/static_assert) ? 
As you are directly copyng the raw memory, perhaps it is even safer to just defined the template to operate on one consistent type.

Nicogene: The problem here is that we can't have any control on `T2` because it is a pcl's type, the only thing that I can check is that the `sizeof` of the yarp type and the pcl type are the same.

I used `static_assert` inside the yarp PointCloud for defining which T1 are allowed and supported for now, the problem is that with the `static_assert` we can add the corresponding list of pcl supported types but inside these conversion functions (`toPcl` and `fromPcl`) we cannot check the correspondence between the yarp type and pcl type (e.g. `yarp::sig::XYZ_DATA` and `pcl::PointXYZ`

Note that it is possible that `sizeof(T1)==sizeof(T2)` but the two types can be incompatible.
Example:
`sizeof(yarp::sig::XYZ_DATA)` is equal to `sizeof(pcl::Normal)` but they have different fields.

This check is not sufficient to be 100% safe but better than nothing :sweat_smile: 
I hope that the user will these functions with the correct types and not try some exotic conversions like yarp XYZ to pcl NxNyNz (Normals):sweat_smile:

traversaro: I see, T1 is something like `struct vec3 { double x; double y; double z; }` and in memcpy you are relying of the memory structure of the two types to be the same, I get it. 
However, I think it is is desirable to transform this in `static_assert(sizeof(T1) == sizeof(T2))`.

Nicogene: Ah ok now I get your point, yes probably is better `static_assert` in this case :+1:

src/libYARP_pcl/include/yarp/pcl/Pcl.h
```diff
@@ -0,0 +1,85 @@
...
+    pclCloud.points.resize(yarpCloud.size());
+    pclCloud.width  = yarpCloud.width();
+    pclCloud.height = yarpCloud.height();
+    memcpy((char*)& pclCloud.points.at(0), yarpCloud.getRawData(), yarpCloud.dataSizeBytes());
+    return true;
+}
+
+/**
+ * Convert a  pcl::PointCloud  to a yarp::sig::PointCloud  object
```
claudiofantacci: There is an extra space before and after `pcl::PointCloud` and after `yarp::sig::PointCloud`.

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,595 @@
...
+         * @param alt the image to clone
+         */
+        const PointCloud<T>& operator=(const PointCloud<T1>& alt)
+        {
+            copy(alt);
+            return *this;
+        }
+
+
+        /** \brief Concatenate a point cloud to the current cloud.
```
claudiofantacci: Probably missing a newline after `/**`?

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,595 @@
...
+                header.isDense = true;
+            }
+            else
+            {
+                header.isDense = false;
+            }
+            return (*this);
+        }
+
+        /** \brief Concatenate a point cloud to another cloud.
```
claudiofantacci: Probably missing a newline after `/**`?

src/libYARP_sig/include/yarp/sig/PointCloudNetworkHeader.h
```diff
@@ -0,0 +1,62 @@
...
+
+
+// YARP_BEGIN_PACK
+/**
+ * @brief The yarp::sig::PointCloudNetworkHeader class
+ */
+class yarp::sig::PointCloudNetworkHeader
+{
+public:
+    PointCloudNetworkHeader() :    width(10),
```
claudiofantacci: Here and the following lines the indentation seems somwhat wrong.

traversaro:
Why `XY_DATA` and all the other data structures are named using `ALL_CAPITAL_LETTERS_WITH_UNDERSCORE` style (typically used for macros in C/C++) while the rest of classes in YARP are named using the `CamelCaseClassName` style?

Nicogene:
@traversaro because they are `struct` and not `class` (?).
The names have been chosen by @barbalberto :sweat_smile:

src/libYARP_sig/include/yarp/sig/PointCloudNetworkHeader.h
Nicogene: It seems ok :thinking:

claudiofantacci: This is how I see it.
schermata 2018-03-15 alle 19 40 08
There are extra spaces here and there after the default constructor semicolon 😄

Nicogene: Ah ok I see, I thought was referred to the identation of public 😅

pattacini:
Thanks @Nicogene and @barbalberto for this awesome integration 💖 

We have the urge to merge this PR in order to start testing an online viewer of point clouds.
Any outstanding point still to discuss?

Nicogene:
All tests, for me the PR is ready to be merged.

doc/tutorials.dox
```diff
@@ -49,6 +49,7 @@ Using YARP devices:
 \li \subpage yarp_config_files
 
 More software tutorials:
+\li \subpage yarp_pointcloud
```
traversaro: I would add  this page also on the  Interoperability and advanced use section in the main page of YARP.

src/libYARP_pcl/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+install(TARGETS YARP_pcl
+        EXPORT YARP
+        COMPONENT YARP_pcl
+        PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/yarp/pcl)
+
+install(DIRECTORY ${PROJECT_SOURCE_DIR}/include/yarp/pcl
+        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/yarp)
+
+set_property(GLOBAL APPEND PROPERTY YARP_LIBS YARP_pcl)
+#set_property(TARGET YARP_pcl PROPERTY FOLDER "Libraries")
```
traversaro: This line can be removed?

src/libYARP_pcl/include/yarp/pcl/Pcl.h
```diff
@@ -0,0 +1,85 @@
...
+#include <yarp/sig/PointCloud.h>
+
+namespace yarp
+{
+namespace pcl
+{
+
+/**
+ * Convert a yarp::sig::PointCloud to a pcl::PointCloud object
+ * @param yarpCloud yarp::sig::PointCloud input
```
traversaro: @param[in] ?

src/libYARP_pcl/include/yarp/pcl/Pcl.h
```diff
@@ -0,0 +1,85 @@
...
+ * @return a pcl PointCloud filled with data contained in the yarp cloud.
+ */
+template< class T1, class T2 >
+inline bool toPCL(yarp::sig::PointCloud< T1 > &yarpCloud, ::pcl::PointCloud< T2 > &pclCloud)
```
traversaro: If `yarpCloud` is just an input argument, perhaps we can have `const yarp::sig::PointCloud< T1 > &yarpCloud`?

src/libYARP_pcl/include/yarp/pcl/Pcl.h
```diff
@@ -0,0 +1,85 @@
...
+    pclCloud.points.resize(yarpCloud.size());
+    pclCloud.width  = yarpCloud.width();
+    pclCloud.height = yarpCloud.height();
+    memcpy((char*)& pclCloud.points.at(0), yarpCloud.getRawData(), yarpCloud.dataSizeBytes());
+    return true;
+}
+
+/**
+ * Convert a pcl::PointCloud to a yarp::sig::PointCloud object
+ * @param pclCloud pcl::PointCloud input
```
traversaro: `@param[in]`

doc/tutorials.dox
Nicogene: Is it possible to do it without restarting all the tests ? :grimacing:

src/libYARP_pcl/include/yarp/pcl/Pcl.h
```diff
@@ -0,0 +1,85 @@
...
+ * @return a yarp cloud filled with data contained in the pcl cloud.
```
traversaro: Missing yarpCloud docs.

src/libYARP_pcl/include/yarp/pcl/Pcl.h
```diff
@@ -0,0 +1,85 @@
...
+
+namespace yarp
+{
+namespace pcl
+{
+
+/**
+ * Convert a yarp::sig::PointCloud to a pcl::PointCloud object
+ * @param yarpCloud yarp::sig::PointCloud input
+ * @return a pcl PointCloud filled with data contained in the yarp cloud.
```
traversaro: No, the return value is bool, this should be the docs of the `@param[out] pclCloud` argument.

src/libYARP_pcl/include/yarp/pcl/Pcl.h
```diff
@@ -0,0 +1,85 @@
...
+inline bool fromPCL(::pcl::PointCloud< T1 > &pclCloud, yarp::sig::PointCloud< T2 > &yarpCloud)
+{
+    static_assert(sizeof(T1) == sizeof(T2), "yarp::pcl::fromPCL: T1 and T2 are incompatible");
+    yarpCloud.fromExternalPC((char*) &pclCloud(0,0), yarpCloud.getPointType(), pclCloud.width, pclCloud.height, pclCloud.is_dense);
+    return true;
+}
+
+/**
+ * Save a yarp::sig::PointCloud to PCD file, ASCII format
+ * @param file_name name of the file to be created wth the cloud
```
traversaro: `@param[in]`

src/libYARP_pcl/include/yarp/pcl/Pcl.h
```diff
@@ -0,0 +1,85 @@
...
+ * @param yarpCloud yarp::sig::PointCloud input
```
traversaro: `@param[out]`

src/libYARP_pcl/include/yarp/pcl/Pcl.h
```diff
@@ -0,0 +1,85 @@
...
+    ::pcl::PointCloud<T2> pclCloud(yarpCloud.width(), yarpCloud.height());
+    yarp::pcl::toPCL< T1, T2 >(yarpCloud, pclCloud);
+    return ::pcl::io::savePCDFile(file_name, pclCloud);
+}
+
+/**
+ * Load a yarp::sig::PointCloud from a PCD file, ASCII format
+ * @param file_name of the PCD file containing the cloud
+ * @param yarpCloud yarp::sig::PointCloud obtained from the PCD file
+ * @return result of the load operation
```
traversaro: It is not clear that `load` is the `::pcl::io::loadPCDFile` method, I would put it with the full method (even because the meaning of the `int` return value is not documented even in PCL documentation)!

src/libYARP_sig/CMakeLists.txt
```diff
@@ -16,7 +16,10 @@ set(YARP_sig_HDRS include/yarp/sig/all.h
                   include/yarp/sig/Matrix.h
                   include/yarp/sig/SoundFile.h
                   include/yarp/sig/Sound.h
-                  include/yarp/sig/Vector.h)
+                  include/yarp/sig/Vector.h
+                  include/yarp/sig/PointCloud.h
+                  include/yarp/sig/PointCloudNetworkHeader.h
+                  include/yarp/sig/PointCloudTypes.h)
```
traversaro: The list was in alphabetical order, I would preserve this.

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,597 @@
...
+
+
+namespace yarp {
+    namespace sig {
+        template <class T> class PointCloud;
+        class PointCloudBase;
+    }
+}
+
+// TBD: Copy constructor (needed by compilers such as Intel C++)
```
traversaro: Not sure what you mean here. Can you open a separate issue if this is a problem?

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,597 @@
...
+
+        virtual bool read(yarp::os::ConnectionReader& connection) override = 0;
+
+        virtual bool write(yarp::os::ConnectionWriter& writer) override = 0;
+
+        virtual int getBottleTag() const = 0;
+
+        /**
+         * @brief
+         * @return the height specified int the yarp::sig::PointCloudNetworkHeader.
```
traversaro: `int the` --> `in the`, repeated

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,597 @@
...
+        /**
+         * @brief
+         * @return the enum representing the point type.
+         */
+        virtual int getPointType() const;
+
+        virtual yarp::os::Type getType() override;
+
+        /**
+         * @brief
```
traversaro: If there is not @brief , just remove it.

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,597 @@
...
+         * @return true if the point cloud is organized in an image-like structure
+         */
+        virtual bool isOrganized();
+
+        /**
+         * @brief isDense
```
traversaro: Remove?

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,597 @@
...
+        {
+            header.width = width;
+            header.height = height;
+            data.resize(width * height);
+        }
+
+        /**
+         * @brief Resize the PointCloud.
+         * @param width.
+         * Calling this function the cloud will become NOT organizedS
```
traversaro: I am not sure what you mean by "will become NOT organizedS"

doc/tutorials.dox
pattacini: @Nicogene let's be flexible and add this change after the merge 😉 (if you like to have the green lights within the PR)

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,597 @@
...
+        }
+
+        /**
+         * @brief Obtain the point given by the (column, row) coordinates.
+         * Only works on organized clouds (those that have height != 1).
+         * @param u, column coordinate
+         * @param v, row coordinate
+         */
+        inline T& operator()(size_t u, size_t v) {
+            yAssert(isOrganized());
```
traversaro: You have already a error handling behaviour (even if not documented) in the following for, why not to use it also for the `isOrganized` case? Crashing is not always a desired behavior on error.

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,597 @@
...
+
+        /**
+         * \brief Concatenate a point cloud to the current cloud.
+         *  \param rhs the cloud to add to the current cloud
+         * \return the new cloud as a concatenation of the current cloud and the new given cloud
+         */
+        inline PointCloud<T>&
+        operator+=(const PointCloud<T>& rhs)
+        {
+            //TODO: take the newest timestamp
```
traversaro: If this is an actual TODO, can you open a issue in the YARP issue tracker?

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,597 @@
...
+
+            yAssert(getPointType() == rhs.getPointType());
```
traversaro: Some gentler error, or perhaps document this behavior in the docs?

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,597 @@
...
+         * \return the new cloud as a concatenation of the current cloud and the new given cloud
+         */
+        inline const PointCloud<T>
+        operator + (const PointCloud<T>& rhs)
+        {
+          return (PointCloud<T> (*this) += rhs);
+        }
+
+        inline void push_back (const T& pt)
+           {
```
traversaro: The indentation of this method is different.

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,597 @@
...
+         * @brief Copy the content of an external PointCloud.
+         * @param source, pointer to the source data.
+         * @param type, enum representing the type of the source cloud.
+         * @param width, width of the source cloud.
+         * @param height, height of the source cloud.
+         * @param isDense
+         */
+        virtual void fromExternalPC(const char* source, int type, size_t width, size_t height, bool isDense = true)
+        {
+            yAssert(source);
```
traversaro: As before, are you sure that crashing is a sensible behavior (especially if not documented) for wrong parameters?

src/libYARP_sig/include/yarp/sig/PointCloudNetworkHeader.h
```diff
@@ -0,0 +1,63 @@
...
+    // PCL like fields
+    yarp::os::NetInt32  width;
+    yarp::os::NetInt32  height;
+    yarp::os::NetInt32  pointType;       // bitwise of all possible informations -> could also be int64 or just an enum, but I thin bitwise gives more freedom about all possible combinations
+    bool                isDense;         // the point cloud is dense if not contains NaN or Inf values
+//     yarp::sig::Vector   pose;            // translation from origin -- could be an Eigen::Vector4f for better PCL compatibility if yarp can afford to depend from it
+//     yarp::sig::Vector   orientation;     // orientation wrt origin  -- could be an Eigen::Quaternion for better PCL compatibility if yarp can afford to depend from it
+
+    // YARPish fileds
+//     char                *data;           // actual pointCloud data.
```
traversaro: If all this commented line are not necessary, can we remove them?

traversaro:
Added some comments but:
* I was not able to review all the PR
* Comments  are quite picky, can be ignored if the mantainer is ok, or can be addressed after the merge

src/libYARP_pcl/CMakeLists.txt
Nicogene: Yes it is a copy-paste from the CMakeLists.txt of  `libYARP_eigen`.
(If you uncomment this line the CMake complaints, so I think it is better to remove it :thinking:)

src/libYARP_sig/include/yarp/sig/PointCloud.h
Nicogene: I think that I can remove it, in PCL the `operator+=` takes the newest timestamp because the timestamp is in the pointcloud header, but in `YARP`  is handled on another level.

src/libYARP_sig/include/yarp/sig/PointCloud.h
Nicogene: @barbalberto is it an issue ?
BTW the copy constructor is defined.

src/libYARP_sig/include/yarp/sig/PointCloud.h
Nicogene: A point cloud is organized if `height > 1`. If you resize it with only one argument it will become not organized.
An organized point cloud is treatable like an image, accessing the pixels with u and v.

src/libYARP_sig/include/yarp/sig/PointCloud.h
Nicogene:
```c++
inline const PointT&
at (int column, int row) const
   {
       if (this->height > 1)
           return (points.at (row * this->width + column));
       else
           throw IsNotDenseException ("Can't use 2D indexing with a unorganized point cloud");
  }
```
This is how PCL handles the access through `u` and `v` in a pointcloud that it is NOT organized.
I can add an error message?

src/libYARP_sig/include/yarp/sig/PointCloudNetworkHeader.h
Nicogene: I think that it is better to keep it for the future improvements

src/libYARP_sig/include/yarp/sig/PointCloud.h
Nicogene: `yarp::sig::Image*` classes are plenty of these `yAssert` :sweat_smile:
I think that the assert is necessary in these cases, but probably it is better to document it

Nicogene:
I can do the changes requested by @traversaro, and I think that tomorrow should be ready.
I know that @pattacini needs it, I don't know how much it is urgent!
@drdanz any comment?

src/libYARP_sig/include/yarp/sig/PointCloud.h
traversaro: I see, then it is a discussion worth having on another place.

src/libYARP_sig/include/yarp/sig/PointCloud.h
claudiofantacci: My suggestion here is to rewrite the sentence as follows:
"This function resizes the point cloud and has the side effect of reducing the height to 1. Thus, by definition, the point cloud becomes unorganized."

Nicogene: Even better with `@note`

Nicogene:
Rebased over last `devel`, waiting that passes all tests...

barbalberto:
Tests are ok, the only one failing is 
` Test #48: dev::ControlBoardRemapperTest ......***Exception: SegFault  1.34 `
which is not related. @traversaro can you check why that test keeps failing randomly?

P.R. looks ok, but I see it introduces a lot of warnings
`: conversion from 'double' to 'float', possible loss of data  src\libyarp_sig\include\yarp\sigpointcloudtypes.h` on windows. @Nicogene can you check?

Nicogene:
I have to check on a windows machine, the lines pointed by appveyor seems wrong

Nicogene:
I fixed all the `conversion from double to float` warnings(not pushed yet).

I have to see  how to fix this warning : `'yarp::sig::PointCloudBase::header': class 'yarp::sig::PointCloudNetworkHeader' needs to have dll-interface to be used by clients of class 'yarp::sig::PointCloudBase'	YARP_sig	C:\Users\isaac\Workspace\yarp_temp_nicogene\src\libYARP_sig\include\yarp\sig\PointCloud.h	96` that seems more toxic ☠️ 
`

Nicogene:
Fixed ALL windows warning.
We can mergr after the test passes.

src/libYARP_sig/include/yarp/sig/PointCloudTypes.h
```diff
@@ -0,0 +1,934 @@
...
+                ret.addInt(g);
+                ret.addInt(b);
+                ret.addInt(a);
+                return ret;
+            }
+            void fromBottle(const yarp::os::Bottle &bt, size_t i)
+            {
+                yarp::os::Bottle* intBt = bt.get(i).asList();
+
+                if(!intBt) return;
```
drdanz: nitpick: please, always use brackets

src/libYARP_sig/include/yarp/sig/Vector.h
```diff
@@ -457,16 +475,15 @@ class YARP_sig_API yarp::sig::Vector : public yarp::os::Portable
     void clear ()
     { storage.clear();}
 
-    ///////// Serialization methods
-    /*
+    /**
     * Read vector from a connection.
-    * return true iff a vector was read correctly
+    * @return true if a vector was read correctly
```
drdanz: iff had a meaning here, if we want to change this, we can discuss it, but this change does not belong in this PR, please keep the changes as small as possible

src/libYARP_sig/include/yarp/sig/Vector.h
```diff
@@ -457,16 +475,15 @@ class YARP_sig_API yarp::sig::Vector : public yarp::os::Portable
...
     */
     virtual bool read(yarp::os::ConnectionReader& connection) override;
 
     /**
     * Write vector to a connection.
-    * return true iff a vector was written correctly
+    * @return true if a vector was written correctly
```
drdanz: same

src/libYARP_sig/src/PointCloud.cpp
```diff
@@ -0,0 +1,185 @@
...
+    return yarp::os::Type::byName("yarp/pointCloud");
+}
+
+bool PointCloudBase::isOrganized()
+{
+    return height() > 1;
+}
+
+PointCloudBase::PointCloudBase()
+{}
```
drdanz: ` = default`

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,590 @@
...
+    /**
+     * @brief The PointCloudBase class
+     *
+     * A Base class for a yarp::sig::PointCloud<T>, provide default implementation for
+     * some methods and hides some internal implementation to the user.
+     * It can't be instantiated, only inherited.
+     */
+    class YARP_sig_API PointCloudBase : public yarp::os::Portable
+    {
+    public:
```
drdanz: You should have a virtual destructor in this class, otherwise, if you want to delete an array of `PointCloudBase*` you will not call the right dtor.

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,590 @@
...
+                return;
+            }
+
+    //        RGBA_DATA       has sense to implement them?
+    //        intensity       has sense to implement them?
+    //        VIEWPOINT_DATA  has sense to implement them?
+
+            header.pointType = 0;
+
+            return;
```
drdanz: Are those return at the end of the void methods here for some reason? I'd rather remove them, since they are useless, and they are just going to confuse the user (I think I saw another one somewhere, but it might have been this one).

src/libYARP_sig/include/yarp/sig/PointCloudTypes.h
```diff
@@ -0,0 +1,934 @@
...
+                        yarp::os::NetInt32 rgba;
+                    };
+                    float curvature;
+                };
+                float filler_others[4];
+            };
+            yarp::os::ConstString toString(int precision, int width)
+            {
+                yarp::os::ConstString ret = "";
+                char tmp[100];
```
drdanz: (This and all the other **100**s). Why 100? please use a power of 2. Also it is probably worth to define this somewhere)

src/libYARP_sig/include/yarp/sig/PointCloudTypes.h
```diff
@@ -0,0 +1,934 @@
...
+                if(width<0)
+                {
+                    sprintf(tmp, "% .*lf % .*lf % .*lf ", precision, x,
```
drdanz: (This and all the other **`sprintf`**s). Please use `snprintf`

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,590 @@
...
+         * @return the enum representing the point type.
+         */
+        virtual int getPointType() const;
+
+        virtual yarp::os::Type getType() override;
+
+        /**
+         * @return true if the point cloud is organized in an image-like structure
+         */
+        virtual bool isOrganized();
```
drdanz: const?

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,590 @@
...
+
+        virtual void copyFromRawData(const char* dst, const char* source, std::vector<int> &recipe);
+
+        virtual std::vector<int> getComposition(int type_composite) const;
+
+        virtual size_t pointType2Size(int type) const;
+
+        virtual size_t getOffset(int type_composite, int type_basic) const;
+
+        yarp::sig::PointCloudNetworkHeader    header;
```
drdanz: This member should be hidden using pimpl, and perhaps accessed by subclasses using some method

src/libYARP_sig/include/yarp/sig/PointCloudNetworkHeader.h
```diff
@@ -0,0 +1,63 @@
...
+#include <yarp/sig/PointCloudTypes.h>
+
+namespace yarp {
+    namespace sig {
+        class PointCloudNetworkHeader;
+    }
+}
+
+
+// YARP_BEGIN_PACK
```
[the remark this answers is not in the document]
barbalberto: The only thing that needs to be packed is the point type used as template argument for the vector. The whole class itself does not.

src/libYARP_sig/include/yarp/sig/Vector.h
barbalberto: This is the very first commit of this PR and everything after rely on it. I understand it is about Vector and not PC, but this context is the one giving a meaning to that commit.
A separated PR with only the change on the vector will be meaningless and confusing since it does not change the 'standard' behavior of the vector.

IMHO I believe it is better to have the change and it's usage together to provide a better understanding of why it was done.

drdanz: Sorry, I was referring just to the documentation change `iff` -> `if`, not to. Here "return true iff a vector was read correctly" doesn't mean "return true if a vector was read correctly" but "return true if a vector was read correctly, false otherwise.
Maybe the meaning of this *biconditional logical connective* (see [`iff` ](https://en.wikipedia.org/wiki/If_and_only_if), used btw in a lot of places (`git grep " iff " | wc -l` -> 114))  is not very clear and we can consider replacing them if we want to make it more explicit, but in this case this is changing 2 conditions to 1. The correct change would be:
```diff
-    * return true iff a vector was read correctly
+    * return true if and only if a vector was read correctly
```
or
```diff
-    * return true iff a vector was read correctly
+    * return true if a vector was read correctly, false otherwise
```
and in case, we should change it everywhere, not just here.

lornat75: can this method return true if the vector is not read correctly?

src/libYARP_sig/include/yarp/sig/Vector.h
```diff
@@ -257,6 +257,24 @@ class yarp::sig::VectorOf : public VectorBase
         len = 0;
         first = nullptr;
     }
+
+    /**
+    * Read vector from a connection.
+    * return true if a vector was read correctly
+    */
+    virtual bool read(yarp::os::ConnectionReader& connection) override
```
drdanz: What is the point in re-implementing these methods? If I understand it correctly, they just return the result of the parent's method, that is exactly what would happen if you don't re-implement them, but at the cost of an extra method call. Am I missing something?

Also I totally missed this change at my first review, this confirms that this kind of changes (a change to a well established class) does not belong to such a pull request (new feature), and it should be done in a separate PR.

Finally if such a change is required, there should be an entry in the release notes.

lornat75:
Ø  What is the point in re-implementing these methods? If I understand it correctly, they just return the result of the parent's method, that is exactly what would happen if you don't re-implement them, but at the cost of an extra method call. Am I missing something?

Maybe it would be useful to specialize these functions for types for which we can provide compatibility with Bottle as it is done for Vector. To be discussed separately?

src/libYARP_sig/include/yarp/sig/Vector.h
lornat75: Maybe it would be useful to specialize these functions for types for which we can provide compatibility with Bottle as it is done for Vector. To be discussed separately?

Nicogene:
After discussing with @barbalberto we decided to remove the commits involving `Vector`.
These commit were made at the beginning(2016) when the idea was to use `Vector::read/write` specialization for the data type e.g.:
```c++
Vector::write<XYZ_DATA>::write()
```
But we changed idea and the `PointCloud::read/write` are agnostic of the type, @barbalberto and I did some black magic 🔮

@drdanz now the PR should be clean, but we have to wait the tests, because I didn't manage to use valgrind on macOs 😅

src/libYARP_sig/include/yarp/sig/PointCloudTypes.h
```diff
@@ -0,0 +1,945 @@
...
+                g = intBt->get(1).asInt();
+                b = intBt->get(2).asInt();
+                a = intBt->get(3).asInt();
+                return;
+            }
+        };
+        YARP_END_PACK
+
+        // Intensity
+        typedef float Intensity;
```
drdanz: Polluting `yarp::sig`

src/libYARP_sig/include/yarp/sig/PointCloudTypes.h
```diff
@@ -0,0 +1,945 @@
...
+
+        // Intensity
+        typedef float Strength;
```
drdanz: Polluting `yarp::sig`

src/libYARP_sig/include/yarp/sig/PointCloud.h
```diff
@@ -0,0 +1,590 @@
...
+
+    };
+
+    } // sig
+
+
+} // yarp
+
+
+#define TagMap(X) \
```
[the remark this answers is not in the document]
Nicogene: you wrote that define :stuck_out_tongue_closed_eyes:

doc/tutorials.dox
Nicogene: https://github.com/robotology/yarp/pull/1597/files#diff-64c78214510ead1279a75f3949512cc3 
done

src/libYARP_sig/include/yarp/sig/PointCloud.h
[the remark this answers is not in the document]
drdanz: :-1: for me!

src/libYARP_sig/include/yarp/sig/PointCloudTypes.h
```diff
@@ -0,0 +1,945 @@
...
+            BORDER_TRAIT__VEIL_POINT_LEFT,
+        };
+
+        // Definition of single fields data structures
+        YARP_BEGIN_PACK
+        struct XYData
+        {
+            union
+            {
+                float _xy[2];
```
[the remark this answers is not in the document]
Nicogene: Vector and Image doesn't check the endianess:
>
```/**
* \ingroup sig_class
*
* A Base class for a VectorOf<T>, provide default implementation for
* read/write methods. Warning: the current implementation assumes the same
* representation for data type (endianess).
*/ 
```
Since we are using internally a `yarp::sig::VectorOf` the PointCloud is affected by the same problem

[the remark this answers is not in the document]
drdanz: Ok, let's forget about this... The NetworkHeader comment is important though, it's not just endianness, also size of bool can change on different architectures

PR #1606, opened 2018-03-16
Fix yarp::OS::getprogname

randaz81:
This PR fixes the implementation of yarp::OS::getprogname() that is not working on Windows. 
Indeed, the method was just a wrapper for an ACE function that was ok on Linux, but was not implemented for Windows (it was just able to return a string previously set by yarp::OS::setprogname()).

The new implementations is now a wrapper around yarp::os::SystemInfo::getProcessInfo(), which is correctly implemented using a windows.h function call. I was tempted to remove yarp::OS::getprogname() completely, but I left it for backward compatibility.

Instead, I removed yarp::OS::setprogname() because it is neither useful nor meaningful to me.

To be tested on MAC_OS.

### DIFF
--- a/src/libYARP_OS/include/yarp/os/Os.h
+++ b/src/libYARP_OS/include/yarp/os/Os.h
@@ -78,15 +78,6 @@ YARP_OS_API const char* getenv(const char* var);
 YARP_OS_API int getpid();
 
 /**
- * @brief Portable wrapper for the setprogname() function.
- *
- * Set the program name.
- *
- * @param[in] progname the program name
- */
-YARP_OS_API void setprogname(const char* progname);
-
-/**
  * @brief Portable wrapper for the getprogname() function.
  *
  * Get the program name.
--- a/src/libYARP_OS/src/Os.cpp
+++ b/src/libYARP_OS/src/Os.cpp
@@ -14,6 +14,7 @@
 #include <yarp/os/impl/PlatformSysStat.h>
 #include <yarp/os/impl/PlatformStdlib.h>
 #include <yarp/os/impl/PlatformSignal.h>
+#include <yarp/os/SystemInfo.h>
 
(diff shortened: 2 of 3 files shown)


randaz81:
sorry for the not-related commit. The plan to merge it with a separated PR, anyway it's already here...

randaz81:
Please note that this PR will be push-forced soon to remove some merging errors.

claudiofantacci:
I tested the PR on macOS with @randaz81 and we found out the following limitation.

The name of the process is parsed via `sysctl()` and the maximum size of the process name is limited to `16 bytes`.  
The constrain can be found in the header file `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/proc.h`, in the `struct extern_proc` that has member `char p_comm[MAXCOMLEN+1]`.  
`MAXCOMLEN` is defined in `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/param.h` to as `16`.

As a result the `yarplogger` reports, on macOS only, a process name that is truncated to 16 characters.

drdanz:
### @randaz81 It doesn't matter if `setprogname` "is neither useful nor meaningful to you", you cannot delete public API methods, especially in a PR entitled "**FIX** *something*", where "*something*" is not even the method you are removing, and without a changelog entry!!!
Also you changed the behaviour on those systems where setprogname is required.



I reverted this change, please propose a proper patch.

randaz81:
I'm thus going to mark as **deprecated** the aforementioned methods, despite the fact that their implementation is **wrong** (in this sense this PR was fixing a severe bug).

I'll replace all occurrence of these methods in user code in order to use an API that is correctly working.

PR #1608, opened 2018-03-17
Actually pass monitor object to wrapped subdevices in CBW

PeterBowman:
Subdevice options did not show up on stdout when using `yarpdev --verbose`. A missing `Searchable::setMonitor` call prevented the monitor object from being passed on to any wrapped device. By the way, the *subdevice* generic name has been replaced so that the actual device name is picked by said monitor.

### DIFF
--- a/doc/release/v2_3_72_1.md
+++ b/doc/release/v2_3_72_1.md
@@ -23,6 +23,10 @@ Bug Fixes
 #### `YARP_dev`
 
 * ControlBoardWrapper now correctly passes on any monitoring to wrapped subdevices.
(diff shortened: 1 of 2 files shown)


Nicogene:
Related to #1473.
@aerydna

jgvictores:
Note: Also happens with `RGBDSensorWrapper`, which contains very [similar code](https://github.com/robotology/yarp/blob/3a4778976fa6c36292aadf4066bc2fe99f15cd2e/src/libYARP_dev/src/devices/RGBDSensorWrapper/RGBDSensorWrapper.cpp#L353) to the lines changed here.

PeterBowman:
> Could you please rebase to latest master?

Done.

PR #1624, opened 2018-03-26
OS: Fix race condition in Nodes

drdanz:
* Refactor the whole class implementation using c++11
* Make the critical sections as small as possible
* Cleanup

(I'm testing the git-hub command line tool to create this PR, so please
forgive me if I mess this up)


drdanz:
> (I'm testing the git-hub command line tool to create this PR, so please forgive me if I mess this up)

Cool, it worked (except that I created it on the `master` branch instead of `devel`). For those who are interested, the correct syntax should be `git hub pull new -b devel`

PR #1631, opened 2018-03-30
[WIP] Introduce yarp protocol version check

barbalberto:
This PR introduce a mechanism to verify that two ports are using the same protocol version before connecting them. If the versions do not match the connection is aborted.

If the port is connecting to a non-YARP entity (ROS, mjpeg stream) the check is not meaningful therefore it is skipped.


barbalberto:
PR updated as requested on last YARP council

doc/protocolVersion_check.dox
```diff
@@ -0,0 +1,157 @@
...
+\li \ref deal_with_changes
+<br>
+
+\section yarp_protocols How YARP ports communicate
+The communication of user data is achieved by means of specialized a class called \ref port_expert_data "Portable", once
+the sender and receiver ports are connected the data will be sent over the network using the specified carrier (tcp, udp,
+mjpeg etc). The ports can optionally send addictional metadata along with the data, like for example the timestamp.
+
+In order to establish a connection betwen two ports, an handshake needs to be performed between them and the yarpserver
+(when present) and this is achieved by mean of an \c internal protocol, referred to as the \b admin protocol.
```
gsaponaro: Minor suggestions:
* an handshake -> a handshake
* by mean of -> by means of

doc/protocolVersion_check.dox
```diff
@@ -0,0 +1,157 @@
...
+    i       Interrupt parent process (unix only)
+    r       Reverse connection type to be a reader
+    /port   Requests to send output to /port
+    !/port  Requests to stop sending output to /port
+    ~/port  Requests to stop receiving input from /port
+    a       Signals the beginning of an administrative message
+    ?       Gives this help
+\endverbatim
+
+Writing the commad \c a we can have access to the admin functionalities through telnet connection.
```
gsaponaro: commad -> command

doc/protocolVersion_check.dox
```diff
@@ -0,0 +1,157 @@
...
+- The Minor number is updated when a compatibility breaking change is applied to the protocol. <br>
+        Anyway the developer <b> must provide a backward compatibility layer for Minor changes</b>,
+        so that YARP communication between two machines running different versions of YARP protocol can still be succesfully performed.
+        See \ref deal_with_changes for more information. <br>
+- The Major number is updated when a new release of the protocol is delivered. Backward compatibility may or may not
+    be guaranteed, depending case by case. <br>
+
+
+\note Please note this is \b NOT the YARP library version, but it is a separated versioning specific for the port protocol. <br>
+YARP connections can be instantiate with different carriers, like tcp, udp, mjpeg, h264, portmonitor and many others.
```
gsaponaro: can be instantiate -> can be instantiated

doc/protocolVersion_check.dox
```diff
@@ -0,0 +1,157 @@
...
+Users can write their own carriers and use them when connecting ports. The YARP protocol version is not related to any
+carrier used for the connection, instead is all about how YARP establish the communication and serialize the data, and it
```
gsaponaro: instead is all about how YARP establish the communication and serialize the data -> instead it is all about how YARP establishes the communication and serializes the data

src/libYARP_OS/src/Network.cpp
```diff
@@ -152,6 +153,94 @@ yarp::os::ProtocolVersion::ProtocolVersion(int Major, int Minor, int Patch) :
...
+                                       const std::string& style,
+                                       std::string &error_msg_p)
+{
+    ProtocolVersion srcVersion, destVersion;
+    ProtocolVersion match;
+    std::string msg;
+
+    /*
+     * We need to discriminate between YARP to YARP connections or YARP to non YARP connection.
+     * The 'canEscape()' function is called for this purpose. In oreder to call it, we need to
```
gsaponaro: minor typo: oreder -> order

src/libYARP_OS/src/Network.cpp
```diff
@@ -140,6 +141,17 @@ static int noteDud(const Contact& src)
 # define CARRIER_DEBUG(fmt, ...)
 #endif
 
+yarp::os::ProtocolVersion::ProtocolVersion() :  protoMajor(YARP_PROTOCOL_VERSION_MAJOR),
+                                                protoMinor(YARP_PROTOCOL_VERSION_MINOR),
+                                                protoPatch(YARP_PROTOCOL_VERSION_PATCH)
+{}
+
+yarp::os::ProtocolVersion::ProtocolVersion(int Major, int Minor, int Patch) :
```
gsaponaro: Hi @barbalberto , in the function declaration (.h) the three parameters Major,Minor,Patch are `const`, however in the function definition (.cpp) they are not. Is it intended?

barbalberto:
Updated, thanks

cmake/YarpVersion.cmake
```diff
@@ -11,6 +11,10 @@ include(GitInfo)
 
 set(YARP_VERSION_ABI "1")
 
+set(YARP_PROTOCOL_VERSION_MAJOR "1")
+set(YARP_PROTOCOL_VERSION_MINOR "2")
+set(YARP_PROTOCOL_VERSION_PATCH "4")
```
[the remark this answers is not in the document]
barbalberto: All three of them were already present, I just moved them. I have no opinion about using 2 or 3 number, I just have not found any reason to remove it. Instead, I think that when a small change is done and it does not break the compatibility (like an optimization or similar), it can be done without enforcing the user to update the code.

The point here however is not simply add or remove messages, consider also case in which the message is the same, but the binary representation change. Like it is gonna happen to envelope soon.

doc/release/v3_0_0.md
```diff
@@ -177,6 +177,7 @@ New Features
   [C++11 Mutex concept](http://en.cppreference.com/w/cpp/concept/Mutex).
 * Added a Timer class to execute a static or member function in a separate
   thread after a period x, every y milliseconds, for z times or w seconds.
+* Introduce YARP protocol version check upon connection.
```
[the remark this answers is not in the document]
barbalberto: Well, this does not make yarp incompatible with previous versions. Keeping the compatibility was one of the requirements

cmake/YarpVersion.cmake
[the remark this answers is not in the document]
drdanz: > All three of them were already present, I just moved them.

Yes, I know, and I was probably the one who suggested that in the first place, to maintain compatibility, sorry about that...
But since we are breaking things, I suggest we handle this properly...
Usually you have 3 number to represent versions, the third is the "implementation" level (private), but on a *protocol* you just have a "public" level, therefore I believe that "MAJOR.MINOR" is enough (also it is what actual protocol use, see HTTP, for example)

> The point here however is not simply add or remove messages, consider also case in which the message is the same, but the binary representation change. Like it is gonna happen to envelope soon.

You are making the example of a *very* breaking change...
If the binary representation changes, that's one of the cases when the `MAJOR` number should be increased...

---

[OT] In the case of the envelope, are you sure you have  I think the way to go would be to add a new message that specifies that you are sending a message with a binary envelope, that means a `MINOR` change. We should discuss this...

barbalberto:
> This is just a first "general" review, I still need to study the messages exchanged, because there is still something that does not convince me...

There are no message involved, just a "get version" and this cannot be removed.
I'll wait for a "proper" review before doing any change

drdanz:
As discussed, we are going in a complete different direction, I think we should close this for now, eventually we can always recover this branch later (perhaps push it to robotology-playground?). @barbalberto do you agree?

barbalberto:
Let's discuss this F2F

drdanz:
The big problem I was talking about F2F is that the check here is not performed in the same connection as the actual connection that will transfer the data.
The actual connection could technically happen with a port that is not the same as the one you requested the version. This is a security issue, and probably easily exploitable.

The version check should be executed at every connection, possibly in the "greeting" message, before the functionalities of the port are actually exposed. The `ver` command itself is subject to changes, therefore, it might disappear, or change return values, or the way to access the administrative interface might change, therefore you cannot call the `ver` command until the 2 sides agree on the protocol version used.

Also my comment about the "PATCH" number for the version is still valid, I don't think we need it, actually I believe it is counterproductive and we should **not** have it.

PR #1632, opened 2018-04-01
OS: Deprecate Port and BufferedPort

drdanz:
As discussed at the latest YARP developers meeting with @lornat75, @barbalberto, @pattacini, @vtikha, @randaz81, @traversaro, @Nicogene, @aerydna and @damn1 YARP 3.0 will no longer support `Port` and `BufferedPort`.

This decision comes fro the continuous discussions that arise every few days on why **ROS** is more widely adopted than **YARP** and on whether **YARP** is better than **ROS** or not, and if it is not why we keep using **YARP** for the iCub robot. This lead us to think that the publish-subscribe paradigm is the way to go, and therefore we decided to concentrate our efforts in this direction for the next major release.

This patch deprecates `yarp::os::Port` and `yarp::os::BufferedPort` in favour of `yarp::os::Publisher` and `yarp::os::Subscriber`.

Please note that this is just the first step towards this new direction, and therefore this patch will cause a lot of deprecation warnings even while compiling YARP. In order to fix these, we will move the `Port` and `BufferedPort` classes in the "impl" folder, and use them internally, and we will add a deprecated compatibility class. Anyway this is not ready yet, but we decided to merge this patch as fast as possible, since the YARP 3.0 release is really close, and we want to allow @robotology/all-robotology-developers to migrate their code as soon as possible.

### DIFF
--- a/doc/release/v3_0_0.md
+++ b/doc/release/v3_0_0.md
@@ -27,6 +27,8 @@ Important Changes
 * The methods `yarp::os::Mutex::tryLock()` and
   `yarp::os::RecursiveMutex::tryLock()` were deprecated in favour of
   `try_lock()`.
+* The `yarp::os::Port` and `yarp::os::BufferedPort` classes are now deprecated
+  in favour of `yarp::os::Publisher` and `yarp::os::Subscriber`.
 
 #### `YARP_sig`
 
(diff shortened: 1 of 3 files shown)


serena-ivaldi:
Could you write a guide for developers to instruct us in how to do the porting in a smooth way?
What does this change means practically?

ps: I hope it's not a April's fish..

drdanz:
This was obviously an April fools, closing this before people actually starts porting their code :rofl:

caicaixia:
:(

PR #1633, opened 2018-04-05
Fix depthCamera and RGBDSensorWrapper

Nicogene:
This PR fixes the CMakelists of `depthCamera` and the compatibility of `RGBDSensorWrapper` with frameGrabberGui2.

Please review code.


drdanz:
Merged, thanks.

PR #1634, opened 2018-04-05
Add realsense2 device driver

Nicogene:
This PR introduce:

- `realsense` device driver.
- `RGBDParamParser` class to unify the parsing of the depth cameras parameters.
- consequent refactor of `depthCamera`

Example of ini file:

```
device       RGBDSensorWrapper
subdevice    realsense2
name         /depthCamera

[SETTINGS]
depthResolution (640 480)    #Note the parentesys
rgbResolution   (640 480)
framerate       30
enableEmitter   true

[HW_DESCRIPTION]
clipPlanes (0.2 10.0)
```

Before merging:

- [x] merge #1633 
- [x] test astra and xtion with ROS (with the help of @aerydna)
- [x] make codacy happy

Probably for a future PR I could investigate if using a callback could improve the framerate.


src/devices/depthCamera/depthCameraDriver.cpp
```diff
@@ -100,36 +112,29 @@ void streamFrameListener::onNewFrame(openni::VideoStream& stream)
 }
 
 
-depthCameraDriver::depthCameraDriver() : m_depthFrame(NULL), m_imageFrame(NULL), m_cameraDescription(NULL)
+depthCameraDriver::depthCameraDriver() : m_depthFrame(NULL), m_imageFrame(NULL), m_paramParser(NULL)
```
pattacini: You might consider using `nullptr` in place of `NULL` throughout the code.
Of course, this would be only optional in case you wanted to adhere to the already existing code.

src/devices/depthCamera/depthCameraDriver.cpp
```diff
@@ -100,36 +112,29 @@ void streamFrameListener::onNewFrame(openni::VideoStream& stream)
...
+        {rgb_Fov,        RGBDSensorParamParser::RGBDParam(rgb_Fov,         2)},
+        {rgbRes,         RGBDSensorParamParser::RGBDParam(rgbRes,          2)},
+        {rgbMirroring,   RGBDSensorParamParser::RGBDParam(rgbMirroring,    1)},
+        {depthMirroring, RGBDSensorParamParser::RGBDParam(depthMirroring,  1)}
+    };
     m_depthRegistration = true;
     m_depthFrame        = new streamFrameListener();
     m_imageFrame        = new streamFrameListener();
-    m_cameraDescription = new CameraParameters();
+    m_paramParser = new RGBDSensorParamParser();
```
pattacini: Keep the existing indentation.

src/devices/depthCamera/depthCameraDriver.cpp
```diff
@@ -139,8 +144,8 @@ depthCameraDriver::~depthCameraDriver()
         delete m_depthFrame;
     if (m_imageFrame)
         delete m_imageFrame;
-    if (m_cameraDescription)
-        delete m_cameraDescription;
+    if (m_paramParser)
```
pattacini: You might want to refactor the destructor like this:
```c++
delete m_depthFrame;
delete m_imageFrame;
delete m_paramParser;
```
This is doable since `nullptr` pointers won't be deleted.

src/devices/depthCamera/depthCameraDriver.cpp
```diff
@@ -139,8 +144,8 @@ depthCameraDriver::~depthCameraDriver()
...
+        delete m_paramParser;
     return;
```
pattacini: `return` to be removed.

src/devices/depthCamera/depthCameraDriver.cpp
```diff
@@ -744,22 +537,22 @@ bool depthCameraDriver::setRgbFOV(double horizontalFov, double verticalFov)
...
         return false;
     }
     return setFOV(horizontalFov, verticalFov, m_depthStream);
 }
 
-bool depthCameraDriver::setDepthAccuracy(double accuracy)
+bool depthCameraDriver::setDepthAccuracy(double _accuracy)
 {
-    if (m_cameraDescription->accuracy.isDescription)
+    if (m_params_map[accuracy].isDescription)
```
pattacini: Didn't go through the code and from the PR it is not immediate to check this out but is `accuracy` chosen on purpose in place of `_accuracy`?

src/devices/depthCamera/depthCameraDriver.h
```diff
@@ -277,9 +273,10 @@ class yarp::dev::depthCameraDriver : public yarp::dev::DeviceDriver,
     impl::streamFrameListener*      m_depthFrame;
     impl::streamFrameListener*      m_imageFrame;
     yarp::os::ConstString           m_lastError;
-    impl::CameraParameters*         m_cameraDescription;
+    yarp::dev::RGBDSensorParamParser* m_paramParser;
```
pattacini: Keep the indentation when possible.

src/devices/realsense2/realsense2Driver.h
```diff
@@ -0,0 +1,226 @@
...
+ * This software may be modified and distributed under the terms of the
+ * BSD-3-Clause license. See the accompanying LICENSE file for details.
+ */
+
+
+#ifndef REALSENSE2_DRIVER_H
+#define REALSENSE2_DRIVER_H
+
+#ifdef WIN32
+#define _USE_MATH_DEFINES
```
pattacini: It is usually safer to add `_USE_MATH_DEFINES` from the CMakeLists.txt file
```cmake
```
in order to avoid unwanted interactions among multiple `#include <cmath>`

Btw, you need to include `cmath` within this file anyway to get `M_PI` define.

src/devices/realsense2/realsense2Driver.h
```diff
@@ -0,0 +1,226 @@
...
+public:
+    realsense2Driver();
+    ~realsense2Driver();
+
+    // DeviceDriver
+    virtual bool open(yarp::os::Searchable& config) override;
+    virtual bool close() override;
+
+    // IRGBDSensor
+    virtual int    getRgbHeight() override;
```
pattacini: If you use `override`, usually, you don't have to specify `virtual` 😉 
That's exactly the purpose people introduced `override` attribute.

src/devices/depthCamera/depthCameraDriver.cpp
Nicogene: Yes because this function is called also in the configuration phase, and in case the accuracy is present as description in the ini file the set has to be skipped :wink:

src/devices/realsense2/realsense2Driver.h
Nicogene: you are right, unfortunately is plenty of these `virtual` `override`, I can change it later in a cleanup commit!

src/devices/depthCamera/depthCameraDriver.cpp
pattacini: Then you might want to highlight this fact by using `this->accuracy`.

Nicogene:
The development can be considered almost finished. 
There is room for enhancement trying to use the callback to get frames from the device.
I don't know if it is necessary, the perfomance on the laptop is good, I have to test it on the NVIDIA jetson(ARM) to see if it can be shipped without callback for now.
The problem is that the installation of the realsense SDK on ARM is not straightforward, and probably it is bugged.
I'd wait to test it on the jetson before merging if everyone agrees.

pattacini:
Uhm @Nicogene if it's working on the laptop/desktop I'd say it'd be worth merging already so that I'll be able to start off using the device driver in the Blacky's system (where `openPose` is installed as well - I've gotta use realsense along with it indeed) without being bothered with handling a branch other than `devel`, you know.

I believe possible bug fixes as resulting from testing on the Jetson can be committed later on.

Of course, my proposal holds in the assumption that the testing activity on the Jetson might take some time.

pattacini:
Alternatively, if you rebase on `devel`, I may be considering switching the Blacky's system to this branch in the meantime.

Nicogene:
@pattacini I see your point, in this case I'm in favour of merging this version of the driver that we can define instable since I will continue working on 😉

pattacini:
@Nicogene @barbalberto 
device driver tested on Blacky successfully 👍 
Thanks for the nice job 🍻

PR #1635, opened 2018-04-05
Fix macos guis compilation

Nicogene:
This PR remove `IMPORTED_GLOBAL` property in `YarpOSXUtilities.cmake` 
This property has been introduced in `cmake 3.11`.

Error: 
```
IMPORTED_GLOBAL property can't be set on non-imported targets
```
CMake documentation : https://cmake.org/cmake/help/v3.11/prop_tgt/IMPORTED_GLOBAL.html
Please review code.


drdanz:
Merged, thanks!

PR #1642, opened 2018-04-10
Improve realsense

Nicogene:
This PR introduces some enhancements to the `realsense2` device.

The most important:
- support for stereo stream (@randaz81)
- flag for eventually disable the alignment of the depth over the rgb.

Example of ini file to use the `realsense2` with the `grabberDual` 
```
device         grabberDual
subdevice   realsense2
capabilities  RAW
stereoMode true

[SETTINGS]
rgbResolution       (640 480)
depthResolution   (640 480)
framerate             30
enableEmitter       true

[HW_DESCRIPTION]
```

Please review code

### DIFF
--- a/src/devices/realsense2/realsense2Driver.cpp
+++ b/src/devices/realsense2/realsense2Driver.cpp
@@ -450,7 +450,11 @@ void realsense2Driver::fallback()
 
 bool realsense2Driver::initializeRealsenseDevice()
 {
-    // TODO get configurations of the device, and read the value from the conf file
+    if (!params_map[rgbRes].isSetting || !params_map[depthRes].isSetting)
+    {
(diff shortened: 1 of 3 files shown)


src/devices/realsense2/realsense2Driver.cpp
```diff
@@ -705,6 +739,7 @@ bool realsense2Driver::setDepthResolution(int width, int height)
         if (m_initialized)
         {
             fallback();
+            return false;
         }
     }
```
barbalberto: The function is `bool` but I don't see any `return true`. What will return the first `if`?

Nicogene: The return `true` is at the end of the function :wink:

barbalberto: Sorry, I missed it. Ok, fine with me

PR #1644, opened 2018-04-12
Remove unmaintained and long deprecated libyarpcxx library

drdanz:
This library is an unmaintained c++ wrapper around libyarpc that is a c wrapper around YARP (c++)
It does not make a lot of sense to keep it...


lornat75:
:+1:

PR #1646, opened 2018-04-13
os/impl/Logger: Remove ACE implementation

drdanz:
Currently we have 2 different implementations for the "deprecated" internal `yarp::os::impl::Logger` class (that at some point will be removed and replaced with the newer `yInfo`, `yDebug` etc. logging methods). One implementation uses "standard" c methods, and works on every platform since it does not require any non-portable library, the other one uses ACE adding a callback and it requires ACE initialization in `Network`. Both implementation have been used and we never had complains about the non-ACE version, I don't think the callback is actually used anywhere.

In this patch the ACE/callback implementation is removed, since I believe it is unused.

The defines were renamed, since otherwise they would conflict when including Logger.h in the same file with an ACE header.

@lornat75 Please let me know if you know something more about this Logger class, and if you think there is some reason why the ACE implementation should not be removed.

### DIFF
--- a/src/libYARP_OS/include/yarp/os/impl/Logger.h
+++ b/src/libYARP_OS/include/yarp/os/impl/Logger.h
@@ -16,20 +16,10 @@
 #include <yarp/os/Log.h>
 #include <yarp/os/impl/PlatformStdio.h>
 
-#ifdef YARP_HAS_ACE
-# include <ace/Log_Priority.h>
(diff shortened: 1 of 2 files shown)


lornat75:
I confirm, as far as I know the ACE implementation can be removed.

drdanz:
Cool, thanks! I will merge this, since I've already run appveyor and travis tests on my fork

PR #1650, opened 2018-04-16
Let yarpdataplayer read ImageOf<PixelFloat> via OpenCV

pattacini:
This PR aims to let `yarpdataplayer` deal with `ImageOf<PixelFloat>` also when OpenCV is used. Indeed, `cvLoadImage()` alone does not load any float image, actually.

Further, I've tried to streamline all the possible flows, saved an unnecessary copy, and made use of `unique_ptr<>` in order to facilitate memory deallocation.

Anyway, writing/reading a float image - which is done through `yarp::sig::file::write()/read()` - appears to be overly inefficient, at least with 640x480 images. Nevertheless, we could now encode handier ways to store/retrieve float images. To get an impression, just take a look at:
https://github.com/robotology/yarp/blob/b806be8c6c3c6e67190e8ec1e428cf3e9e1c2b64/src/libYARP_sig/src/ImageFile.cpp#L26-L115

I ran several tests with standard RGB images successfully. Perhaps, a more extensive campaign would be necessary.

### DIFF
--- a/doc/release/v3_0_0.md
+++ b/doc/release/v3_0_0.md
@@ -143,6 +143,10 @@ Bug Fixes
 
 * Fixed race condition in yarp::os::Nodes.
 
+#### `GUIs`
+
+* Made `yarpdataplayer` able to deal with `ImageOf<PixelFloat>` also with OpenCV.
+
 
 Contributors
 ------------
--- a/src/yarpdataplayer/src/worker.cpp
+++ b/src/yarpdataplayer/src/worker.cpp
@@ -21,6 +21,7 @@
     #pragma warning (disable : 4520)
 #endif
 
+#include <memory>
 #include "yarp/os/Stamp.h"
 #include "include/worker.h"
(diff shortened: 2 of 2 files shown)


lornat75:
After checking yarpview extensively I concluded there may still be a bug on how the datasetplayer streams images, any chance you noticed a possible race condition?

pattacini:
@lornat75 not really caught any race condition, apparently, even though I was not focusing on that particularly.

However, the PR fixed a memory leak in the case where OpenCV was not used (Codacy pointed it out as well), plus an unnecessary copy should be spared. Although, as said, I didn't notice anything special with regards to race conditions.

Aside from this, If you grant me the green light 🚥 I'd like to proceed by refactoring the `yarp::sig::file::read/write` functionalities by using binary formats rather than plain text. I believe no one is storing float images on files. This code turns to be pretty much uncovered in our daily activities.

PR #1652, opened 2018-04-17
Storing binary files for ImageOf<PixelFloat>

pattacini:
Read/write services over `ImageOf<PixelFloat>` do produce binary files now. We should definitely gain something in terms of performance.


pattacini:
@Nicogene I still need to run a couple of tests!

pattacini:
Tested: working nicely!

PR #1662, opened 2018-04-25
Updated/fixed LUA documentation

pattacini:
As of now, we've successfully tested `lua5.2` with yarp, plus the path where the binding gets saved is different from what stated in the present doc and thus needs a fix.


drdanz:
Merged, thanks.

PR #1673, opened 2018-05-04
Fix yarpmanager res/conn list

### LINKED ISSUE #964: mismatch of port connections when not ordered by ID
Hey there, 
I noticed that when changing the default ordering of the port connections in YARPMANAGER from ID to any other (from, to, state, etc), very often the information displayed is not matched anymore. That is, the green and red indications of which ports are open and closed, and which pairs are connected does not correspond to reality anymore.

Best, 
Tanis

Nicogene:
This PR fixes #964 .

Moreover it fixes the same bug in the resources list.

please review code


drdanz:
Rebased and merged, thanks.

PR #1674, opened 2018-05-04
Implement improvements of Lua tests proposed at #1656

jgvictores:
This PR implements the proposal of #1656:
- Rename `test_string.lua` to more representative `test_port.lua`, enable it to  run without the `yarpserver`, and integrate it in the test infrastructure (and thus also removed from `bindings/lua/tests/run.sh`).


bindings/lua/tests/run.sh
```diff
@@ -14,4 +14,4 @@ fi
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-lua $DIR/test_string.lua
+# lua $DIR/test_that_requires_yarpserver.lua
```
drdanz: Is this a leftover?

drdanz: Maybe add a comment, i.e.
```diff
-lua $DIR/test_string.lua
+# Any test that requires yarpserver would go here" 
+# lua $DIR/test_that_requires_yarpserver.lua

jgvictores: Sure, added at 8b7aa10d720c17d08aa7b49fac45765f84a02fd9 : )

drdanz:
Rebased and merged, thanks.

PR #1677, opened 2018-05-07
Properly handle 8,16,32,64 bit integers and 32,64 bit floating point values

drdanz:
This was supposed to be a rather trivial change, but it ended up in the classic "can of worms"...

The handling of 8,16,32, and 64 bit integers was rather incomplete and there was no way to transfer 8 and 16 bit integers, for example yarpidl_rosmsg was bypassing this by sending 8 and 16 bit as "blocks".

Moreover, all the "Int" and "Double" functions, were not architecture safe, on different architectures, these types could have different size, therefore this might lead to unexpected bugs.

The list of changes is rather long, please check each commit, for details, I'll summarize the most important changes here:

* Replace `YARP_INT{8,16,32,64}` with `std::int{8,16,32,64}_t` from `cstdint`
* Use defines from `cinttypes` instead of `YARP_INT64_FMT`
* Add `yarp::conf::float{32,64,128}_t` (128 only if supported)
* Add `yarp::conf::ssize_t`
* Deprecate `YARP_INT{8,16,32,64}` in favour of `std::int{8,16,32,64}_t`
* Deprecate `YARP_FLOAT{32,64}` in favour of `yarp::conf::float{32,64}_t`
* Deprecate `YARP_INT64_FMT` in favour of `PRId64`
* Deprecate `YARP_CONF_SSIZE_T` in favour of `yarp::conf::ssize_t`

* Add the following methods:
  * `yarp::os::Bottle::addInt{8,16,32,64}()`
  * `yarp::os::Bottle::addFloat{32,64}()`
  * `yarp::os::Value::{is,as,make}Int{8,16,32,64}()`
  * `yarp::os::Value::{is,as,make}Float{32,64}()`
  * `yarp::os::ConnectionWriter::appendInt{8,16,32,64}()`
  * `yarp::os::ConnectionWriter::appendFloat{32,64}()`
  * `yarp::os::ConnectionReader::expectInt{8,16,32,64}()`
  * `yarp::os::ConnectionReader::expectFloat{32,64}()`
  * `BOTTLE_TAG_INT{8,16,32,64}` (`BOTTLE_TAG_INT32` = `BOTTLE_TAG_INT`)
  * `BOTTLE_TAG_FLOAT{32,64}` (`BOTTLE_TAG_FLOAT64` = `BOTTLE_TAG_DOUBLE`)

* The following methods are not deprecated due to the large codebase where they are used, but their usage in new code is discouraged, and it is forbidden inside YARP. Anyway, all these methods are now calling the corresponding Int32 and Float64 version, but are not architecture safe, since they still accept `int` and `double`.
  * `yarp::os::Bottle::addInt()`
  * `yarp::os::Bottle::addDouble()`
  * `yarp::os::Value::isInt()`
  * `yarp::os::Value::isDouble()`
  * `yarp::os::Value::asInt()`
  * `yarp::os::Value::asDouble()`
  * `yarp::os::Value::makeInt()`
  * `yarp::os::Value::makeDouble()`
  * `yarp::os::ConnectionWriter::appendInt()`
  * `yarp::os::ConnectionWriter::appendDouble()`
  * `yarp::os::ConnectionReader::expectInt()`
  * `yarp::os::ConnectionReader::expectDouble()`
  * `BOTTLE_TAG_INT`
  * `BOTTLE_TAG_DOUBLE`

**WARNING - Behaviour change**: An `Int64` value no longer returns true when `isInt()` is called. I'm not sure if this was an intended behaviour or a bug, but it does not make sense.

I added a "fake" unit test, whose only goal is to print this table, that is really useful for debugging.

screenshot_20180507_182639

In a second moment, it could be useful to extend this table with the other types supported by `Value`

-------

Edit:
* `yarp::os::Value` ctors are now explicit
* Added `ResourceFinder::setDefault()` overloads for `int32_t` and `float64_t`.


drdanz:
Tested on the iCub black

PR #1679, opened 2018-05-09
Add check on carriers for yarpserver --ros

### LINKED ISSUE #722: yarpserver used with --ros should not start if required carriers are not found/compiled
yarpserver used with --ros option is meant to work together with roscore and therefore it requires the following carriers:
- rossrv
- tcpros
- xmlrpc

right now the server does not check if the plugins are available, and at the connection time it prints an error

yarp: Failed to find library support for carrier xmlrpc
yarp: Could not find carrier "xmlrpc"

which is kind of difficult to spot. 
IMHO it'd be better if the server does not start if the carriers are not available in order to catch the problem as soon as possible.

Nicogene:
It fixes #722 .

This PR makes `xmlrpc` `rossrv` `tcpros` carriers mandatory for running `yarpserver --ros`.

Please review code.

### DIFF
--- a/doc/release/v2_3_72_1.md
+++ b/doc/release/v2_3_72_1.md
@@ -34,6 +34,12 @@ Bug Fixes
 * Fixed `RGBDSensorWrapper` compatibilty with `frameGrabberGui2`.
 * Fixed joint remapping in method `getRefVelocity()`.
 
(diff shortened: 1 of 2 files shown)


drdanz:
Merged, thanks.

PR #1638, opened 2018-04-09
add timeout support for TcpConnector::connect()

damn1:
This is an attempt to fix #1636.

If timeout is not not explicitly set, `select()` function remains blocking with a default timeout of ~1min maintaining behaviour of `TcpConnector::connect()` as before.

PR #1681, opened 2018-05-09
change TcpConnector print method to avoid problems with remote logging

damn1:
Fixes (another) bug introduced by #1638.


Nicogene:
Merged, thanks!

PR #1685, opened 2018-05-10
Cleanup singleton

Nicogene:
This PR c++11-fy all(?) the singletons in `yarp`.

Please review code.

### DIFF
--- a/src/libYARP_init/src/CustomInit.cpp
+++ b/src/libYARP_init/src/CustomInit.cpp
@@ -38,9 +38,7 @@ extern "C" int __yarp_is_initialized;
 
 extern "C" void yarpCustomFini()
 {
-    yarp::os::LogForwarder::clearInstance();
 }
-
 
 yarp::os::Network::Network() {
     Network::init();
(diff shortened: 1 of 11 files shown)


src/libYARP_init/src/CustomInit.cpp
```diff
@@ -36,11 +36,6 @@ extern "C" void yarpCustomInit() {
 
 extern "C" int __yarp_is_initialized;
 
-extern "C" void yarpCustomFini()
-{
-    yarp::os::LogForwarder::clearInstance();
-}
-
```
drdanz: yarpCustomFini() has been there since the origin of YARP, probably for a symmetry reason, see

https://github.com/robotology/yarp/blob/ea3403b353a5da80e0912063ed74fba39c13583b/src/libYARP_init/src/CustomInit.cpp#L33-L34

Also the LogForwarder should have never been initialized there, the right place would be finiMinimum, this call is supposed to be there for initializing stuff initialized in yarpCustomInit.

Anyway, I'd rather just remove the `yarp::os::LogForwarder::clearInstance();` call for now and restore the original empty body of the call (the compiler will probably optimize it anyway...)

Nicogene:
👍  done

PR #1686, opened 2018-05-10
Enhance CMake output

drdanz:
Current output

screenshot_20180510_204303

Default output with this patch:

screenshot_20180510_204117

Output with this patch and `export CLICOLOR_FORCE=1`

screenshot_20180510_204015
screenshot_20180510_204044


claudiofantacci:
Out of sheer curiosity: what's the output with CMake GUIs?

drdanz:
> Out of sheer curiosity: what's the output with CMake GUIs?

"Default output" if `CLICOLOR_FORCE` is not set, "messed up output" if `CLICOLOR_FORCE` is set :cry:
I couldn't find a way to check if CMake is running in a terminal or in a GUI.

drdanz:
Just for reference, this is related: https://bixense.com/clicolors/

drdanz:
Failures are unrelated.
Merging this

PR #1691, opened 2018-05-17
[WIP] Cleanup interfaces

Nicogene:
This PR:
-  Deprecates the interfaces that ends with "2".
- Moves all the methods of the 2 in the interfaces not numbered.
- Deprecates obsolete members of old interfaces.

List of Interfaces purged:
- `ICalibrateControl2` -> `ICalibrateControl`
- `ICalibrateControl2Raw` -> `ICalibrateControlRaw`
- `IFrameGrabberControl2` -> `IFrameGrabberControl`
....

To make yarp compiles extra changes have been done, and this PR probably breaks many things.

DO NOT MERGE


drdanz:
Merged, thanks.

PeterBowman:
I noticed that a lot of stuff in CBW2 could be removed after these changes: [example](https://github.com/robotology/yarp/blob/aa45b107d9f85aaf45a4a5edb21df4e5e99e4864/src/libYARP_dev/src/devices/ControlBoardWrapper/ControlBoardWrapper.cpp#L1546-L1560). Not sure if you are already considering a cleanup of this device, @Nicogene.

Nicogene:
> I noticed that a lot of stuff in CBW2 could be removed after these changes: example. Not sure if you are already considering a cleanup of this device, @Nicogene.

Yes, the cleanup you pointed out should be done one day, honestly I don't know when :sweat_smile: .
If you want to open a PR, it is more than welcome :+1: 
This should be safe but in general the CBW is quite delicate to change it

PeterBowman:
>  If you want to open a PR, it is more than welcome +1

See https://github.com/robotology/yarp/pull/1999.

PR #1700, opened 2018-05-24
move math operators outside yarp::math namespace

### LINKED ISSUE #1602: Operators should not be in the yarp::math namespace
`Vector` and `Matrix` operators are defined inside the `yarp::math` namespace.

https://github.com/robotology/yarp/blob/29e9108a52713dbe737d9a9c44a62da3adc3ce51/src/libYARP_math/include/yarp/math/Math.h#L16-L27

This means that, in order to perform an operation (for example `Vector b = a + s;`), the user has to add `using namespace yarp::math;` or to call the operator explicitly, i.e. (I think) something like this: `Vector b = yarp::math::operator+(a, s);`
The first option is considered bad practice, and the second option is just wrong.

I think we could just move them outside of the namespace, the only effect is if someone is using the second syntax, but I don't think anyone would be doing that...

damn1:
fixes #1602

### DIFF
--- a/doc/release/v3_0_0.md
+++ b/doc/release/v3_0_0.md
@@ -165,6 +165,8 @@ Important Changes
 * added new utility methods to Quaternion class.
 * method toRotationMatrix() has been renamed to toRotationMatrix4x4() to avoid
   confusion with method toRotationMatrix3x3().
+* operators for `yarp::math::<class>`es have been moved from the `yarp::math` to
+  the global namespace.
 
 ### Tools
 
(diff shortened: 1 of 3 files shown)


jgvictores:
Opinion on the open [proposal](https://github.com/robotology/yarp/issues/1602#issuecomment-380063350) of open issue #1602 ?

Nicogene:
@jgvictores we discussed it internally and for the moment we are not planning to move the operators in a different `namespace`.
In the [example](http://www.cplusplus.com/reference/utility/rel_ops/) you pointed out the operators are in the `namespace rel_ops` because these are templates and this will  introduce these overloads for all types not defining their own.

Nicogene:
Merged, thanks!

jgvictores:
The origin of the inclusion of the operators in the namespace, cited [here](https://github.com/robotology/yarp/issues/1602#issuecomment-379692519), was `follow the strict guidelines`.

Could the `PR type: Bugfix` label be removed?

PR #1702, opened 2018-05-25
FrameTransformClient: add consistency check in setTranform

barbalberto:
TF where parent and child frame have the same name are rejected as invalid.
A TF like that caused the client to run into a infinite loop.


barbalberto:
@aerydna @randaz81 Can you please check if there are other places where this check could be useful?
I don't know if there are other ways to insert a tranform into the database (from config file?) or other function where this check may be enforced.

drdanz:
Rebased and merged, thanks.

PR #1708, opened 2018-05-26
Friday night cleanup (Network, Time)

drdanz:
* Remove unneeded calls to Time::turboBoost() (already called in initMinimum())
* OS/Time: Deprecate turboBoost() (called automatically by Network)
* OS: Make Time a namespace (it was an useless instantiable class with just static methods)
* OS/Network: Remove Bottle and BottleImpl initialization (no longer necessary)
* OS/Network: Reset system timer resolution at the end of the execution. According to https://msdn.microsoft.com/en-us/library/vs/alm/dd757624(v=vs.85).aspx `timeBeginPeriod` (called in startTurboBoost`) affects a global Windows setting and should be matched with a call to  `timeEndPeriod` (called inendTurboBoost).

### DIFF
--- a/example/dev/simple_motor_client.cpp
+++ b/example/dev/simple_motor_client.cpp
@@ -85,8 +85,7 @@ int main(int argc, char *argv[])
     }
 
     Network yarp;
-	Time::turboBoost();
-    
+
     char name[1024];
     Value& v = options.find("robot");
     Value& part = options.find("part");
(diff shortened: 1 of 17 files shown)


pattacini:
⚠️ converting `Time` into a namespace could very likely break user code since there's typically no `using namespace yarp` directive in the user code.

pattacini:
`Time` is under `yarp::os`, sorry!
Then, everything is fine. Removing my comments.

barbalberto:
Nice!

PR #1711, opened 2018-05-28
move Companion out of YARP_OS

### LINKED ISSUE #1509: Move Companion out of YARP_OS
I suggest we make a separate library for that.

damn1:
Fixes #1509 

In order to remove any `libYARP_OS` dependences from the `Companion` some changes were necessary:
* Some `Companion` functions has been moved to `NetworkBase`
    * `disconnectInput`
    * `poll`
    * `sendMessage`
    * `wait` (the `wait()` has been splitted into two new functions `waitConnection()` and `waitPort()` to keep code clean and readable.

* `Companion::exists()` has been removed unifying its code with `NetworkBase::exists()`. A parameter has been added to manage the last check in the function. 

* `readString()` function has been deprecated in `NetworkBase` and its implementation has been moved from `Companion` to a common utility file in OS (`Terminal.h`).

* `libYARP_companion` library has been created which exposes the `main` function from the `yarpcompanion.h` file.

### DIFF
--- a/src/libYARP_OS/include/yarp/os/impl/Terminal.h
+++ b/src/libYARP_OS/include/yarp/os/impl/Terminal.h
@@ -1,0 +1,35 @@
+/*
+ * Copyright (C) 2006-2018 Istituto Italiano di Tecnologia (IIT)
+ * Copyright (C) 2006-2010 RobotCub Consortium
+ * All rights reserved.
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD-3-Clause license. See the accompanying LICENSE file for details.
+ */
+
+#ifndef YARP_OS_IMPL_TERMINAL_H
+#define YARP_OS_IMPL_TERMINAL_H
+
+#include <string>
+#include <yarp/os/api.h>
+
+namespace yarp {
+namespace os {
+namespace impl {
+namespace terminal {
+
+YARP_OS_impl_API
+bool EOFreached();
+
+YARP_OS_impl_API
+std::string getStdin();
+
+YARP_OS_impl_API
+std::string readString(bool *eof);
+
+} // terminal
+} // impl
+} // os
+} // yarp
+
+#endif // YARP_OS_IMPL_TERMINAL_H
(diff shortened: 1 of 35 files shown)


src/libYARP_OS/include/yarp/os/impl/Terminal.h
```diff
@@ -0,0 +1,186 @@
...
+            }
+
+        /* if no names matched, then return null. */
+        return ((char *)nullptr);
+    }
+    /* Attempt to complete on the contents of TEXT.  START and END show the
+   region of TEXT that contains the word to complete.  We can use the
+   entire line in case we want to do some simple parsing.  Return the
+   array of matches, or nullptr if there aren't any. */
+    static char ** my_completion (const char* text, int start, int end)
```
drdanz: This method generates a -Wunused-function warning, is this used anywhere?

src/libYARP_OS/include/yarp/os/impl/Terminal.h
```diff
@@ -0,0 +1,186 @@
...
+#ifdef WITH_LIBEDIT
+    #include <editline/readline.h>
+    #include <vector>
+    static std::vector<std::string> commands;
+    static yarp::os::Port* rpcHelpPort = nullptr;
+    static bool commandListInitialized = false;
+    static char* szLine = (char*)nullptr;
+    static bool readlineEOF=false;
+
+    static char* dupstr(char* s)
```
drdanz: These function should not be declared static in a .h file...
Also please move the implementation in a .cpp file

drdanz:
Please rebase over the latest devel, since there is a conflict...

src/libYARP_OS/include/yarp/os/impl/Terminal.h
damn1: Actually `dupstr()` is only used by `generate_commands()`, which is only used by `my_completion()`. I see it is assigned here https://github.com/robotology/yarp/blob/9c8cea6fa292864ff2ee4aacf3a98553836cdc8d/src/libYARP_OS/src/Companion.cpp#L2191 
when `WITH_LIBEDIT` is set, to a pointer to function of the Readline library (explained here:
http://www.delorie.com/gnu/docs/readline/rlman_47.html)

drdanz:
Merged, thanks.

PeterBowman:
@damn1 please see https://github.com/robotology/yarp/issues/1509#issuecomment-394099994.

drdanz:
Just a reminder to myself. By moving the companion out of YARP_OS, we were planning to remove the optional dependency to libedit. Unfortunately RFModule uses libedit as well, we should do some more changes in order to remove it.

Nicogene:
Also `yarpamanger-console` needs libedit

PR #1714, opened 2018-05-29
Generated licence update

### LINKED ISSUE #1576: License for generated files
Generated files currently don't have a license. Usually it is assumed that files without a license statement are "all right reserved", and it would mean that the files are not BSD, and therefore YARP_dev binary could not be BSD.

I think we should modify the idl generators so that they add a BSD boilerplate to those files, to make it clear that you can use them in your code.

CC @lornat75

aerydna:
fixes #1576


drdanz:
Merged, thanks.

PR #1716, opened 2018-05-30
Replace int with size_t in Bottle

Nicogene:
This PR replaces int with size_t in Bottle where necessary.

I have intentionally put the target future with yarp 3, eventually this PR can be postponed, it should not be a breaking change.

`demoRedBall` tested successfully.
Please reviecode.

### DIFF
--- a/tests/libYARP_OS/PortTest.cpp
+++ b/tests/libYARP_OS/PortTest.cpp
@@ -428,7 +428,7 @@ public:
             if (result!=nullptr) {
                 checkEqual(bot1.size(),result->size(),"size check");
                 YARP_INFO(Logger::get(),std::string("size is in fact ") +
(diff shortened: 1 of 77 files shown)


tests/libYARP_OS/PortTest.cpp
```diff
@@ -428,7 +428,7 @@ class PortTest : public UnitTest {
             if (result!=nullptr) {
                 checkEqual(bot1.size(),result->size(),"size check");
                 YARP_INFO(Logger::get(),std::string("size is in fact ") +
-                          NetType::toString(result->size()));
+                          NetType::toString((unsigned int) result->size()));
```
drdanz: Reminder: ADD `NetType::toString(size_t)`

drdanz:
Merged, thanks.

PR #1725, opened 2018-06-04
[WIP] Quaternion and other math-related classes moved from libYARP_Math to libYARP_sig

randaz81:
The declaration of following yarp classes have been moved from libYARP_math to libYARP_sig.

Quaternion
FrameTransform
Vec2D

The implementation of methods which do not use libYARP_math functionalities are defined in libYARP_sig, libYARP_math otherwise.


drdanz:
As discussed, this might not be the right solution, I'm closing this for the moment.

drdanz:
In case you are looking for the branch, I pushed the branch to robotology-playground and deleted the original one

PR #1739, opened 2018-06-10
Pull up module path tweaks to main config file

PeterBowman:
This is a follow up to #1733 for non-static YARP builds, which were prone to cause similar errors in downstreams (see #1736) due to internal dependencies (e.g. TinyXML) not being found.


drdanz:
@PeterBowman I think this change is no longer needed after #1748, please re-open if I'm wrong

Issue #1682, opened 2018-05-10
C++11-fy yarp singletons


Nicogene:
Here the list of all singleton classes inside `yarp` that should be c++11-fied: 
- [x] `yarp::os::Carriers`
- [x] `yarp::os::LogForwarder`
- [x] `yarp::manager::ErrorLogger`
- [x] `PlotManager` (`yarpscope`)

(Please update the list above if you find other singleton like them)

Nicogene:
#1685 c++11-fied all singleton remained in yarp, closing for now.

Nicogene:
#1699 c++11-fy more singletons

PR #1744, opened 2018-06-12
Drivers: c++11-fy singleton

damn1:
Related to #1682


Nicogene:
Is it related to a particular bug?

damn1:
Not one that is documented in this repo. I put the label just to give a type to the pull request but maybe cleanup is more suitable.

drdanz:
Merged, thanks.

PR #1753, opened 2018-06-15
Add noexcept in move operators and constructor

Nicogene:
Please review code.

see http://www.codingstandard.com/rule/12-5-4-declare-noexcept-the-move-constructor-and-move-assignment-operator/


drdanz:
I'm not 100% sure that changelog is not required for such a change, since it breaks the ABI... Well let's merge it anyway.

drdanz:
Merged, thanks.

PR #1761, opened 2018-06-19
Fix negative vocab

### LINKED ISSUE #1749: Negative VOCAB in Image.h
As noticed by @jgvictores (https://github.com/robotology/yarp/commit/185a9c36334860c5f655cf25b3766a7c680cf215#r29116913) there is a negative `VOCAB` in Image.h

https://github.com/robotology/yarp/blob/185a9c36334860c5f655cf25b3766a7c680cf215/src/libYARP_sig/include/yarp/sig/Image.h#L63-L65

@aerydna is this wanted or just a typo?

Nicogene:
This PR fixes #1749.

It will break probably the ABI compatibility but I didn't find another way to fix it.

Please review code.

### DIFF
--- a/doc/release/v3_0_1.md
+++ b/doc/release/v3_0_1.md
@@ -40,6 +40,10 @@ Bug Fixes
 
 * Fixed missing isDense parmeter.
 
+#### YARP_sig
+
+* Fixed negative vocab(#1749).
+
 ### GUIs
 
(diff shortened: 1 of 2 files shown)


drdanz:
I'd say this was a bug. i don't know if this was used anywhere. Let's merge and see if people complains. :rofl:

PR #1766, opened 2018-06-22
update yarp name help and add list of yarprun ports command

aerydna:
allows to list only the ports opened by yarprun executables


drdanz:
Merged, thanks.

PR #1768, opened 2018-06-22
PlatformTime.cpp: changed sleep_for() to delaySystem()

aerydna:
PlatformTime.cpp: changed sleep_for() to delaySystem()


drdanz:
Why do you want this? The implementation of `delaySystem()` is exactly the same...

PR #1769, opened 2018-06-25
Fix get Current/get Currents using stateExt and iCurrentControl

lrapetti:
Fix the implementation of `getCurrent` and `getCurrents` methods in the `RemoteControlBoard` using the `stateExt:o` port.

The implementation has been modified also in the `ControlBoardWrapper` in order to use the `ICurrentControl` interface methods.

### DIFF
--- a/doc/release/v3_0_1.md
+++ b/doc/release/v3_0_1.md
@@ -43,6 +43,10 @@ Bug Fixes
   are available.
 * Fixed SharedLibraryClassFactory::destroy(). Now it is const like create().
 
+#### YARP_dev
+
+* Fixed getCurrent/getCurrents. Now the `stateExt` port is used, and the methods are called through the `iCurrentControl` interface.
+
 #### YARP_math
 
 * Fixed a regression in the build system that prevented YARP from being
--- a/src/libYARP_dev/src/devices/ControlBoardWrapper/ControlBoardWrapper.cpp
+++ b/src/libYARP_dev/src/devices/ControlBoardWrapper/ControlBoardWrapper.cpp
@@ -3442,55 +3442,6 @@ bool ControlBoardWrapper::getAmpStatus(int j, int *v)
     return false;
 }
 
-bool ControlBoardWrapper::getCurrents(double *vals)
-{
-    double *currs = new double[device.maxNumOfJointsInDevices];
-    bool ret = true;
-    for(unsigned int d=0; d<device.subdevices.size(); d++)
-    {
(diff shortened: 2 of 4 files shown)


barbalberto:
Looks good to me. @randaz81 the `getCurrent` from `iAmplifier` and from `iCurrentControl` interface are equivalent, right?

Nicogene:
Please add a line in the changelog `v3_0_1.md` thanks!

lrapetti:
@Nicogene 
changelog updated https://github.com/robotology/yarp/pull/1769/commits/4891fc20aaa9f41ef9569551579ef30acdb98a3a

lrapetti:
>  could you please rebase the branch over the latest devel

@Nicogene 
Did you mean the latest master?

Nicogene:
>Did you mean the latest master?
Yes sorry it was a lapsus!

barbalberto:
I'd suggest a small change. The two functions `getCurrent` in the `amp` interface and in the `iCurrentControl` are equivalent. The low level device may or may not implement any of them.

Since we are improving the wrapper, we may use the `iCurrentControl` as you did and, if the interface is not implemented by the device, the wrapper can fallback to the `amp` interface and try to get the values from there.

Anyway great catch!! Well done

lrapetti:
> Since we are improving the wrapper, we may use the iCurrentControl as you did and, if the interface is not implemented by the device, the wrapper can fallback to the amp interface and tries to get the values from there.

@barbalberto 
I have add the use of `IAmplifier` in case the `ICurrentControl` is not found, so both interfaces are now supported.

barbalberto:
@lrapetti I'd like to ask you another small change.

Suppose the device has both `amp` and `current` interfaces implemented, and the `getCurrent` method returns false.

In this case the `getCurrent` method will be called twice, and the second time is useless because we already know the result will be false, because they collapse to the same code in the low level device.

The device is, supposely, getting the values from the hardware and it may takes some time to produce the answer.

I'd split the check in two parts
```
if(interface1)
    getCurrent
else if(interface2)
    getCurrent

if(ret)
   assign values    // just one for loop

```
This way the `getCurrent` is called only once.

lrapetti:
@barbalberto I have updated the commit as you suggested and rebased.
According to the changes I have changed the error condition initializing `ret=false`:
```
ret = false
if(interface1)
    ret = getCurrent
else if(interface2)
    ret = getCurrent

if(ret)
   assign values 
else
   error condition
```

barbalberto:
Great, many thanks. When test are succesfully we can merge it

PR #1778, opened 2018-07-02
Fix YARP_COMPILE_EXECUTABLES build

drdanz:
The `idl_tools` component is now required in order to use `yarp_idl_to_dir` and `yarp_add_idl`.
This component is included by default when COMPONENTS are not passed to `find_package(YARP)`
Moreover if the component is not included but either  `yarp_idl_to_dir` or `yarp_add_idl` are used, a deprecation warning is printed, and the component is included, making it backwards compatible.

Warning: requires latest YCM `devel`, perhaps we should release YCM 0.9 before releasing YARP 3.0.1. Since I needed YCM form devel, I also took the chance to refactor a little, and move YCM to the `extern` folder


drdanz:
@pattacini

drdanz:
I just noticed that this fixes only CMake run, the build is still failing

drdanz:
Should be fixed now
@traversaro Please note that I had to make some change to `multipleAnalogSensorsMsgs` in order to support building the devices without the executables.

drdanz:
And I was wrong :crying_cat_face: 
The static builds are broken again

drdanz:
ok, hopefully it should be fixed now...

PR #1779, opened 2018-07-03
Coverity fixes

damn1:
Fixes some problems reported by Coverity static analyzer

### DIFF
--- a/src/carriers/tcpros_carrier/yarpros.cpp
+++ b/src/carriers/tcpros_carrier/yarpros.cpp
@@ -338,7 +338,7 @@ int main(int argc, char *argv[]) {
                       lookup.portnum,
                       reply);
         printf("%s\n", reply.toString().c_str());
-        return ok?0:1;
+        return 0;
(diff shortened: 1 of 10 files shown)


src/yarpmotorgui/main.cpp
```diff
@@ -158,6 +157,11 @@ int main(int argc, char *argv[])
                     }
                 }
             }
+
+            if (desc_driver)
+            {
+                delete desc_driver;
+            }
```
drdanz: I'm not sure about what is happening here, can you explain why you are deleting the polydriver? Is it no longer needed?

drdanz:
@damn1 I removed the second commit, rebased and merged, we can discuss about the second commit (6558dc0aeb48da8f7ab5779e858204bf0e2c0c4e) later

src/yarpmotorgui/main.cpp
Nicogene: You are right, I missed it I thought the main terminated there.
BTW before deleting the pointer it I would call `desc_driver->close()`

PR #1781, opened 2018-07-04
OS/Type: Refactor using PIMPL, move implementation in .cpp, cleanup

drdanz:
Warning: small ABI change (removed private variables, changed copy assignment operator return value to non-const), API is the same.

The move constructor and move assignment operator are "new" therefore I'm adding them later on devel


src/libYARP_OS/src/Type.cpp
```diff
@@ -7,60 +7,185 @@
...
+    {
+        rhs.prop = nullptr;
     }
-}
 
-Type::~Type() {
-    if (prop) {
+    ~Private()
+    {
         delete prop;
```
Nicogene: I would add
```
if (prop)  {
    delete prop;
    prop = nullptr;
}
```
Or even better shall we move to `unique_ptr` ? :heart_eyes:

src/libYARP_OS/include/yarp/os/Type.h
```diff
@@ -9,84 +9,88 @@
...
+     *
+     * @param rhs the Type to copy
+     * @return this object
+     */
+    Type& operator=(const Type& rhs);
+
+/** @} */
+/** @{ */
+
+    std::string getName() const;
```
Nicogene: It would be good to add the documentation also for the other methods :wink:

src/libYARP_OS/src/Type.cpp
drdanz: :-1: for the extra if and assignment to `nullptr` in the destructor... they are completely useless, `delete` will not delete a `nullptr`, and after the destruction `prop` is no longer accessible, therefore no need to assign it.

:+1: for unique_ptr, but that's for another pr...

src/libYARP_OS/include/yarp/os/Type.h
drdanz: It would be good, if someone knew what they are supposed to do! :laughing:

PR #1789, opened 2018-07-10
OVR compilation fix with yarp 3.0

mebbaid:
`ovrheadset.cpp` used deprecated methods/objects, doesn't compile with new YARP, hence the modifications.

This PR and corresponding fork handles the `yarp::math` and `rateThread` related errors when compiling the `ovrheadset.cpp` . It, one the other hand, doesn't address the breaking changes in the Oculus SDK from verions 1.17 onwards, it assumes you are using the 1.16 version (or older).

### DIFF
--- a/src/devices/ovrheadset/OVRHeadset.cpp
+++ b/src/devices/ovrheadset/OVRHeadset.cpp
@@ -775,7 +775,7 @@ void yarp::dev::OVRHeadset::threadRelease()
         // Disable Performance Hud Mode before destroying the session,
         // or it will stay after the device is closed.
         int PerfHudMode = (int)ovrPerfHud_Off;
-        ovr_SetInt32(session, OVR_PERF_HUD_MODE, PerfHudMode);
+        ovr_SetInt(session, OVR_PERF_HUD_MODE, PerfHudMode);
 
         ovr_Destroy(session);
         session = 0;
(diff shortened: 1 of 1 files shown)


src/devices/ovrheadset/OVRHeadset.cpp
```diff
@@ -775,7 +775,7 @@ void yarp::dev::OVRHeadset::threadRelease()
         // Disable Performance Hud Mode before destroying the session,
         // or it will stay after the device is closed.
         int PerfHudMode = (int)ovrPerfHud_Off;
-        ovr_SetInt32(session, OVR_PERF_HUD_MODE, PerfHudMode);
+        ovr_SetInt(session, OVR_PERF_HUD_MODE, PerfHudMode);
```
drdanz: Those changes are not related to YARP 3, are those changed in the Oculus SDK? If they are, we should update also the `find_package` call...

mebbaid:
Yes, indeed, in fact I am using the 1.16 release before the breaking changes, however, for some reason also the 'Ovr_GetInt32' throws an undefined identifier error.

src/devices/ovrheadset/OVRHeadset.cpp
mebbaid: Some changes are related to yarp, the ones concerning rateThread and systemPeriodThread

drdanz:
Rebased and merged on master

PR #1790, opened 2018-07-11
Fix extern/ycm CMakeLists

claudiofantacci:
`YARP_MODULE_PATH` was used in place of `YCM_MODULE_PATH` and, as a consequence, CMake 3.8 ycm modules were not added in `CMAKE_MODULE_PATH`.


drdanz:
Moved to master and merged. Thanks.

Issue #1763, opened 2018-06-20
Fix calibrate() related warnings


drdanz:

drdanz:
Ping @randaz81

drdanz:
@randaz81 Fixed in `devel`, suppressed in `master`. Closing this.

PR #1791, opened 2018-07-11
Calib fix

randaz81:
IControlCalibration interface cleanup and directory refactoring.
Fixes issue #1763 https://github.com/robotology/yarp/issues/1763
Bump version 3.0.101.


randaz81:
As discussed with @traversaro and @drdanz some time ago, we chose to not apply this policy to specific parts (>80%) of yarp_dev library, in order to speed-up  development on robots. Additionally, the interface modified by this PR is implementation (robot) dependent, meaning that it is for internal use only. Users should never call these methods, they MUST use the remote calibration interface developed by @barbalberto. Hence, the only known issue is with the motioncontrols of icub-main, already fixed in a different PR.

randaz81:
Discussed f2f with @drdanz and @barbalberto. The PR has been approved.

PR #1792, opened 2018-07-11
Fix rplidar2 compilation

### LINKED ISSUE #1689: rplidar2 doesn't compile
**Describe the bug**
`rplidar2` device fail to compile:
```
/home/ngenesio/yarp/extern/rplidar/sdk/src/rplidar_driver.cpp:528:10: error: case value evaluates to -1, which cannot be narrowed to type 'unsigned long' [-Wc++11-narrowing]
    case rp::hal::Event::EVENT_TIMEOUT:

```

**To Reproduce**
Enable `ENABLE_yarpmod_rpLidar2`

**Configuration:**
 - OS: linux Kubuntu 16.04
 - yarp version: 2.3.73
-  compiler: clang

Nicogene:
This static cast is necessary for the compilation because the return value of "wait(..)" is unsigned but EVENT_TIMEOUT = -1. We will keep it until they will fix the bug in the sdk.(fixes #1689)

Moreover add NOT APPLE as dependency of the device since some bugs in the sdk make it impossible to compile on macOs(see https://travis-ci.org/robotology/yarp/jobs/402720473 for example)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -404,6 +404,9 @@ before_script:
   # OpenCV on travis is built with C++98 ABI.
   - if [ "$TRAVIS_OS_NAME" == "linux" ]; then if [ "$CC" == "clang" ]; then export YARP_CMAKE_OPTIONS="${YARP_CMAKE_OPTIONS} -DYARP_USE_OPENCV=OFF"; fi; fi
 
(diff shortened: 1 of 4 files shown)


drdanz:
Merged, thanks.

PR #1795, opened 2018-07-11
Fix typos

drdanz:
I did not catch them manually, I used `codespell` for this...
Please double-check to ensure that this patch is not changing something that should not be touched


drdanz:
Fixed `occur`, and rebased. Merging, since the tests passed.

PR #1796, opened 2018-07-11
OS/impl/UnitTest: Fix segfault when generated failure string is longer than 1000 characters

drdanz:
`sprintf` in a fixed size buffer was obviously generating crashes whenever the generated string was longer than 1000 characters.


drdanz:
For some reason this commit ended up in #1795 ...
Closing this

PR #1798, opened 2018-07-12
Fix multi namespace

Nicogene:
Multinamespace is a c++11 singleton(since #1699)  and need to be initialized
before the first port that is opened and it has to exist until
the last port is closed.

Please review code.
This PR is a firs step for the documentation of the `Network` object see #1670

### DIFF
--- a/doc/release/v3_0_1.md
+++ b/doc/release/v3_0_1.md
@@ -58,6 +58,7 @@ Bug Fixes
 * Fixed SharedLibraryClass::getContent(). Now a const and a non const version
   are available.
 * Fixed SharedLibraryClassFactory::destroy(). Now it is const like create().
+* Added initialization of `MultiNameSpace` in `Network::initMinimum`.
 
 #### YARP_dev
 
--- a/src/libYARP_OS/src/Network.cpp
+++ b/src/libYARP_OS/src/Network.cpp
@@ -867,12 +867,15 @@ void NetworkBase::initMinimum(yarp::os::yarpClockType clockType, yarp::os::Clock
         // make sure system is actually able to do things fast
         yarp::os::impl::Time::startTurboBoost();
 
-        __yarp_is_initialized++;
+        // MultiNameSpace is a c++11 singleton and need to be initialized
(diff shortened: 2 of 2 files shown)


src/libYARP_OS/src/Network.cpp
```diff
@@ -845,6 +845,10 @@ void NetworkBase::initMinimum(yarp::os::yarpClockType clockType, yarp::os::Clock
     if (__yarp_is_initialized == 0) {
         // Broken pipes need to be dealt with through other means
         yarp::os::impl::signal(SIGPIPE, SIG_IGN);
+        // Multinamespace is a c++11 singleton and need to be initialized
+        // before the first port that is opened and it has to exist until
+        // the last port is closed.
+        getNameSpace();
```
drdanz: I'm not sure where it used to be, but I believe that this should be *after* ACE and (I think) after all the other initialization (quiet, verbose and probably turboboost)

doc/release/v3_0_1.md
```diff
@@ -58,6 +58,7 @@ Bug Fixes
 * Fixed SharedLibraryClass::getContent(). Now a const and a non const version
   are available.
 * Fixed SharedLibraryClassFactory::destroy(). Now it is const like create().
+* Added initialization of `Multinamespace` in `Network::initMinimum`.
```
drdanz: `Multinamespace` -> `MultiNameSpace`

src/libYARP_OS/src/Network.cpp
```diff
@@ -845,6 +845,10 @@ void NetworkBase::initMinimum(yarp::os::yarpClockType clockType, yarp::os::Clock
     if (__yarp_is_initialized == 0) {
         // Broken pipes need to be dealt with through other means
         yarp::os::impl::signal(SIGPIPE, SIG_IGN);
+        // Multinamespace is a c++11 singleton and need to be initialized
```
drdanz: `Multinamespace` -> `MultiNameSpace`

drdanz:
Rebased and merged, thanks.

PR #1799, opened 2018-07-12
Coverity fixes

damn1:
New fixes, and I also resumed https://github.com/robotology/yarp/commit/6558dc0aeb48da8f7ab5779e858204bf0e2c0c4e after https://github.com/robotology/yarp/pull/1779#discussion_r200121789.

### DIFF
--- a/src/yarpmotorgui/main.cpp
+++ b/src/yarpmotorgui/main.cpp
@@ -140,8 +140,7 @@ int main(int argc, char *argv[])
             desc_driver_options.put("device", "robotDescriptionClient");
             desc_driver_options.put("local", descLocalName);
             desc_driver_options.put("remote", "/robotDescription");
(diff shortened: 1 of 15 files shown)


src/yarpmotorgui/main.cpp
```diff
@@ -158,6 +157,12 @@ int main(int argc, char *argv[])
                     }
                 }
             }
+
+            if (desc_driver)
+            {
+                desc_driver->close();
+                delete desc_driver;
+            }
```
Nicogene: `desc_driver=nullptr` ?

drdanz: Not really necessary, since the pointer goes out of scope right after the delete...

drdanz:
Rebased and merged, thanks.

PR #1801, opened 2018-07-13
Interface templates

randaz81:
* The following clsses have been refactored. They are no more template classes:
  ImplementAmplifierControl
  ImplementEncoders

* Class definitions have moved from:
  src/libYARP_dev/include/yarp/dev/ImplementControlBoardInterfaces.h
  to:
  src/libYARP_dev/include/yarp/dev/ImplementAmplifierControl.h
  src/libYARP_dev/include/yarp/dev/ImplementEncoders.h

* ControlBoardInterfaces.h has been split into multiple new files:
  IAmplifierControl.h
  IAxisInfo.h
  IControlDebug.h
  IControlLimits.h


randaz81:
One final cleanup step could be renaming all files:
`*Impl.h` 
into:
`Implement*.h`
(or viceversa). 
Currently, there is no file name standard.

This latter is not a breaking change: implement files are used only inside yarp_dev and are not directly included by user code.

barbalberto:
I do not have a srtong preference about names, maybe `*Impl.h'.
> This latter is not a breaking change: implement files are used only inside yarp_dev and are not directly included by user code.

Not so true, devices like `canBusMotionControl` do require those include files, but usually they include `ControlBoardInterfacesImpl.h` which, in turns, includes the other ones so should not be a problem.

randaz81:
Fixed #1763

doc/release/v3_1_0.md
```diff
@@ -17,9 +17,47 @@ New Features
...
+  src/libYARP_dev/src/ImplementMotorEncoders.cpp (from src/libYARP_dev/src/IMotorEncodersImpl.cpp)
+  src/libYARP_dev/src/ImplementPWMControl.cpp (from src/libYARP_dev/src/PWMControlImpl.cpp)
+  src/libYARP_dev/src/ImplementPidControl.cpp (from src/libYARP_dev/src/IPidControlImpl.cpp)
+  src/libYARP_dev/src/ImplementPositionControl.cpp (from src/libYARP_dev/src/IPositionControlImpl.cpp)
+  src/libYARP_dev/src/ImplementPositionDirect.cpp (from src/libYARP_dev/src/IPositionDirectImpl.cpp)
+  src/libYARP_dev/src/ImplementRemoteCalibrator.cpp (from src/libYARP_dev/src/RemoteCalibratorImpl.cpp)
+  src/libYARP_dev/src/ImplementRemoteVariables.cpp (from src/libYARP_dev/src/IRemoteVariablesImpl.cpp)
+  src/libYARP_dev/src/ImplementTorqueControl.cpp (from src/libYARP_dev/src/TorqueControlImpl.cpp)
+  src/libYARP_dev/src/ImplementVelocityControl.cpp (from src/libYARP_dev/src/IVelocityControlImpl.cpp)
+  src/libYARP_dev/src/ImplementVirtualAnalogSensor.cpp (from src/libYARP_dev/src/VirtualAnalogSensorImpl.cpp)
```
drdanz: @randaz81 For next time, please note that:
* the release notes are for the final user, there is no need to list all the `.cpp` files renamed, only the changes to the `.h` files are important, but the *internal* path is not... (i.e. `yarp/dev/ImplementControlLimits.h (from yarp/dev/IControlLimitsImpl.h)` is enough).
* Use 80 characters word wrap.
* Wrap file names, methods, etc. using the "``" character, so that they are rendered using the `monospaced font`
* The list of files in the `CMakeLists.txt` are sorted alphabetically, please keep it that way.

Issue #1684, opened 2018-05-10
Remove dynamic allocations from "impl" layer of motor control interfaces


valegagge:
In order to fix the issue #1660 , I allocated temporary buffers dinamiccally.
 
As suggest by @lornat75 , It is better remove dynamic allocations from "impl" layer of yarp despite of performance test results, for two reasons:
 - the "impl" layer could be used on embedded in the future
 - it is preferable to have dynamic allocations concentrated only at certain points of the code

Therefore I'll develop a simple memory manager to solve this issue; it could be used in other contex also.

drdanz:
If the issue is related to multi-threading, perhaps declaring the buffers [`thread_local`](http://en.cppreference.com/w/cpp/language/storage_duration) can help? Maybe not, I don't know the details of the issue... :stuck_out_tongue_closed_eyes:

PR #1804, opened 2018-07-16
Some fix in impl-layer of control board interfaces

valegagge:
In this branch I fixed some small bugs in the implementation layer of some control board interface.

I fixed following function:

- `ImplementControlMode::setControlModes`: where axes conversion was missing.

- `ImplementCurrentControl::getCurrentRanges(double*min, double *max)` and `ImplementCurrentControl::getCurrentRanges(j, double *min, double *max)`: where conversion from Sensor to Ampere was missing. I think this conversion is needed because the user can set the reference current with a value belonging to (min, max), like [documenation says.](http://www.yarp.it/classyarp_1_1dev_1_1ICurrentControl.html#af1ef99cf8a35950d437253711e17e9bc)

- `ImplementPositionControl::checkMotionDone(bool *flag)` where axes conversion was missing


src/libYARP_dev/src/ControlModeImpl.cpp
```diff
@@ -104,5 +104,12 @@ bool ImplementControlMode::setControlModes(const int n_joint, const int *joints,
 
 bool ImplementControlMode::setControlModes(int *modes)
 {
-    return raw->setControlModesRaw(modes);
+    int *modes_tmp=new int [nj];
```
Nicogene: I would use `std::array` instead of dynamic allocation, here and in the other places.
See http://www.modernescpp.com/index.php/std-array-dynamic-memory-no-thanks

src/libYARP_dev/src/IPositionControlImpl.cpp
```diff
@@ -160,7 +160,14 @@ bool ImplementPositionControl::checkMotionDone(const int n_joint, const int *joi
...
 bool ImplementPositionControl::checkMotionDone(bool *flag)
 {
-    return iPosition->checkMotionDoneRaw(flag);
+    bool flags_tmp[nj];
+    bool ret = iPosition->checkMotionDoneRaw(flags_tmp);
+    for(int i=0; i<nj; i++)
+    {
+        flag[i] = flags_tmp[castToMapper(helper)->toHw(i)];
+    }
+    delete []flags_tmp;
```
Nicogene: `flag_tmp` has not dynamically allocated, coverity is right, this `delete` has to be removed.

src/libYARP_dev/src/ControlModeImpl.cpp
Nicogene: Actually it is not possible to declare:
```c++
std::array<int, nj> modes_tmp;
```
because `nj` is not `constexpr`. :cry: 

But here we can use smart pointers and get rid of all `delete` :wink:
```
std::unique_ptr<int[]> modes_tmp(new int [nj]);
```

valegagge: Hi @Nicogene ,
this request would fix some bugs only.

Here is not the place to change the dynamic allocations, that will be modify with a fixed size buffer manager asap, but on devel branch. If you are interested in, please see https://github.com/valegagge/yarp/tree/fixedSizeBuffersManager.

The dynamic allocations had been a temporary solotion to fix issue (https://github.com/robotology/icub-support/issues/624) in order to continue make faster the firmware response to some user-request, parallelizing the request to all motion control devices. (see (https://github.com/robotology/yarp/pull/1668)

The fixed-size buffer-manager will be use instead of dynamic allocation as explained in: https://github.com/robotology/yarp/issues/1684.

So, here I would only to fix some wrong behaviour of these functions.
The important thing now is to understand if my changes fix bugs or not.

cheers 
Vale

src/libYARP_dev/src/IPositionControlImpl.cpp
valegagge: now flags_tmp is allocated dinamically: before it was not declared in standard way.
It was fault. sorry! see https://github.com/robotology/yarp/pull/1804/commits/8cf860aebc8e4cb30402b2501622937ee7928560

src/libYARP_dev/src/ControlModeImpl.cpp
Nicogene: Hi @valegagge ! I see, if these dynamic allocations has to die so it is not worthy to migrate to smart pointers :wink:

valegagge: yes! I'm testing fixedsizeBufferManager! :)
So we can merge this pr?

randaz81:
Beware: some files edited in this PR (which is applied to master branch) do not exist anymore in devel branch.

src/libYARP_dev/src/ControlModeImpl.cpp
Nicogene: I'd add some lines in the changelog, and I don't know if @drdanz is going to release yarp `3.0.1` soon. Shall we include it in this release or in `3.1.1` ??

PR #1808, opened 2018-07-19
Fix copyright headers

### LINKED ISSUE #950: Update GPL/LGPL boilerplates
According to the [GPL-2.0](https://www.gnu.org/licenses/old-licenses/gpl-2.0.html) and [LGPL-2.1](https://www.gnu.org/licenses/old-licenses/lgpl-2.1.html) documentation:

> To apply these terms, attach the following notices to the library. It is safest to attach them to the start of each source file to most effectively convey the exclusion of warranty; and each file should have at least the "copyright" line and a pointer to where the full notice is found.

We have always used a "summary" where the license and the copyright holders are listed, but I think that using the "official" [GPL](https://www.gnu.org/licenses/old-licenses/gpl-2.0.html#SEC4)/[LGPL](https://www.gnu.org/licenses/old-licenses/lgpl-2.1.html#SEC4) boilerplates is "safer", and therefore I suggest that we start changing them.

What do you think?

cc @lornat75

drdanz:
Everything that was not BSD is now LGPL2.1+ with these exceptions:

* `src/yarpscope` uses QCustomPlot that is GPL3+ and it is therefore GPL3+
* `src/yarpviz` uses qgv that is GPL3+ and it is therefore GPL3+
* `example/ContainerExample` uses yarpscope and it is therefore GPL3+
* `example/matrix/gsl_example.cpp` uses GSL and it is therefore GPL2+
* `src/carriers/depth_image_portmonitor` and `src/carriers/zfp_portmonitor` are BSD.

The license of following files is still unclear:

* `bindings/java/examples/matlab/simulink/sfun_time.c`
* `bindings/lua/argcargv.i`
* `cmake/FindGraphviz.cmake`
* `example/yayarpview/CImg.h`

All the other files *should* now be ok.

@lornat75 please review.

Fixes #950


jgvictores:
Re `bindings/java/examples/matlab/simulink` (for the record): As commented in the [README](https://github.com/robotology/yarp/blob/ec0e3d4c6e62236dc20dcb450be1cdfd33e1407e/bindings/java/examples/matlab/simulink/README.txt), `sfun_time.c` and `utility.mdl` were imported from an external project. The link to the original project is actually broken now (sorry for introducing this :ghost:).

The 2 files are only used to introduce some soft real-time in the Simulink process, because things go pretty crazy without it. Search provides [several alternatives](https://www.mathworks.com/matlabcentral/fileexchange/?utf8=%E2%9C%93&term=soft+real+time), but all options pitfall into the same copyright issue. : (

The example was nice, so I wouldn't recommend deleting it completely. Maybe an option would to be to remove the 2 files from YARP and recommend users to drop some soft real-time block in the project for it not to go nuts. The drawbacks would be that this is not immediate (references to the block should be removed from the Simulink project), and that the example would no longer be self-contained.

PR #1825, opened 2018-08-03
Coverity fixes

damn1:
New fixes

### DIFF
--- a/src/libYARP_OS/src/NameConfig.cpp
+++ b/src/libYARP_OS/src/NameConfig.cpp
@@ -267,15 +267,12 @@ std::string NameConfig::getHostName(bool prefer_loopback, const std::string& see
 
             if (ip.find(':')!=std::string::npos) continue;
 
-            bool would_be_loopback = false;
-            if (ip == "127.0.0.1" || ip == "127.1.0.1" ||
-                ip == "127.0.1.1") {
-                would_be_loopback = true;
-            }
-#ifdef YARP_HAS_ACE
-#ifdef ACE_ADDR_HAS_LOOPBACK_METHOD
-            would_be_loopback = ips[i].is_loopback();
-#endif
+#if defined YARP_HAS_ACE && defined ACE_ADDR_HAS_LOOPBACK_METHOD
+            bool would_be_loopback = ips[i].is_loopback();
+#else
+            bool would_be_loopback = (ip == "127.0.0.1" ||
+                                      ip == "127.1.0.1" ||
+                                      ip == "127.0.1.1");
 #endif
 
             // If we haven't any interface yet, take this one
--- a/src/libYARP_OS/src/PortReaderBufferBase.cpp
+++ b/src/libYARP_OS/src/PortReaderBufferBase.cpp
@@ -572,35 +572,24 @@ bool PortReaderBufferBase::acceptObjectBase(PortReader *obj,
             mPriv->consumeSema.wait();
         }
     }
-    bool ok = true;
-    if (ok) {
-        reader->setExternal(obj, wrapper);
 
-        mPriv->stateMutex.lock();
-        bool pruned = false;
-        if (mPriv->ct>0 && mPriv->prune) {
-            PortReaderPacket *readerPacket =
-                mPriv->dropContent();
-            pruned = (readerPacket!=nullptr);
-        }
-        //mPriv->configure(reader, false, true);
-        mPriv->pool.addActivePacket(reader);
-        mPriv->ct++;
-        mPriv->stateMutex.unlock();
-        if (!pruned) {
-            mPriv->contentSema.post();
-        }
-        //YARP_ERROR(Logger::get(), ">>>>>>>>>>>>>>>>> adding data");
-    } else {
-        mPriv->stateMutex.lock();
-        mPriv->pool.addInactivePacket(reader);
-        mPriv->stateMutex.unlock();
-        //YARP_ERROR(Logger::get(), ">>>>>>>>>>>>>>>>> skipping data");
+    reader->setExternal(obj, wrapper);
 
-        // important to give reader a shot anyway, allowing proper closing
-        YARP_DEBUG(Logger::get(), "giving PortReaderBuffer chance to close");
+    mPriv->stateMutex.lock();
+    bool pruned = false;
+    if (mPriv->ct>0 && mPriv->prune) {
+        PortReaderPacket *readerPacket =
+            mPriv->dropContent();
+        pruned = (readerPacket!=nullptr);
+    }
+    //mPriv->configure(reader, false, true);
+    mPriv->pool.addActivePacket(reader);
+    mPriv->ct++;
+    mPriv->stateMutex.unlock();
+    if (!pruned) {
         mPriv->contentSema.post();
     }
+    //YARP_ERROR(Logger::get(), ">>>>>>>>>>>>>>>>> adding data");
 
     return true;
 }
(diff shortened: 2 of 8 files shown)


src/libYARP_companion/src/Companion.cpp
```diff
@@ -1860,7 +1860,10 @@ int Companion::write(const char *name, int ntargets, char *targets[]) {
         std::string slash=NetworkBase::getDirectorySeparator();
         hist_file += slash;
         hist_file += "yarp_write";
-        yarp::os::mkdir(hist_file.c_str());
+        if (yarp::os::mkdir(hist_file.c_str()) != 0)
+        {
+            return 1;
```
traversaro: Most `return 1` in this file come with a ` fprintf(stderr, "error message\n");` that prints and error that explains the problem. I think it would be helpful to have it also here, printing a message with the path to the directory that it was not possible to create. A typical case in which this message would be helpful is if for some reason `getDataHome` is full or not writable.

src/libYARP_manager/src/localbroker.cpp
```diff
@@ -978,7 +977,10 @@ int LocalBroker::ExecuteCmd()
         close(pipe_child_to_parent[WRITE_TO_PIPE]);
         FILE* in_from_child = fdopen(pipe_child_to_parent[READ_FROM_PIPE],"r");
         int flags=fcntl(pipe_child_to_parent[READ_FROM_PIPE],F_GETFL,0);
-        fcntl(pipe_child_to_parent[READ_FROM_PIPE],F_SETFL,flags|O_NONBLOCK);
+        if (fcntl(pipe_child_to_parent[READ_FROM_PIPE],F_SETFL,flags|O_NONBLOCK) != 0)
+        {
+            return -1;
```
traversaro: Why `-1` as a return value? It does not seems to be used in the rest of the function anywhere, and is not documented in the function signature.

src/libYARP_companion/src/Companion.cpp
```diff
@@ -1450,8 +1450,8 @@ int Companion::cmdDetectRos(bool write) {
     }
 
     std::string uri = NetworkBase::getEnvironment("ROS_MASTER_URI");
-    if (uri=="") {
-        fprintf(stderr, "ROS_MASTER_URI environment variable not set.\n");
+    if (uri=="" || uri.length() > 512) {
+        fprintf(stderr, "ROS_MASTER_URI environment variable not set or too long.\n");
```
traversaro: It would be nice as a user to know that too long means "more than 512 characters", so I can eventually take care of it with some trick if in my case ROS_MASTER_URI is longer.

src/libYARP_manager/src/localbroker.cpp
damn1: Right I didn't check carefully. I see there are similar situations in the same function at https://github.com/robotology/yarp/blob/4bca59c4f41bff16e9b4d453257878ab263745be/src/libYARP_manager/src/localbroker.cpp#L879
and 
https://github.com/robotology/yarp/blob/4bca59c4f41bff16e9b4d453257878ab263745be/src/libYARP_manager/src/localbroker.cpp#L895
so I'll do the same here.

src/libYARP_OS/src/PortReaderBufferBase.cpp
```diff
@@ -572,35 +572,24 @@ bool PortReaderBufferBase::acceptObjectBase(PortReader *obj,
...
+    if (mPriv->ct>0 && mPriv->prune) {
+        PortReaderPacket *readerPacket =
+            mPriv->dropContent();
+        pruned = (readerPacket!=nullptr);
+    }
+    //mPriv->configure(reader, false, true);
+    mPriv->pool.addActivePacket(reader);
+    mPriv->ct++;
+    mPriv->stateMutex.unlock();
+    if (!pruned) {
```
Nicogene: I'm I am afraid that
```c++
bool ok = true;
if (ok) { ...
}
else { ...
}
```
one time had sense, was it inside a loop and `ok` were updated?
Obviously your fix is right, I'm just worried about the possible hidden bug.

damn1: We will never know why that piece of code is like that. The comment on the function 
```
/// Careful!  merge with ::read -- very similar code
/// Until merge, don't change one without looking at other :-(
```
made me think about a copy/paste and change where needed. In the `::read()` function that `ok` variable is set with a result of an operation, maybe here it was not the case but to keep the same structure of the code it was directly set to true.

Nicogene: You are right it has been completely copy-pasted, so I think we can safely remove this useless `if-else`

src/yarp-config/yarpcontextutils.cpp
```diff
@@ -862,8 +862,15 @@ int diffList(folderType fType, bool verbose)
...
-            if (recursiveDiff(installedPath + PATH_SEPARATOR + (*subDirIt).c_str(), userPath, tmp)>0)
-                std::cout << (*subDirIt) << std::endl;
+            try
+            {
+                if (recursiveDiff(installedPath + PATH_SEPARATOR + (*subDirIt).c_str(), userPath, tmp)>0)
+                    std::cout << (*subDirIt) << std::endl;
+            }
+            catch (...)
+            {
+                printf("Excpetion occurred during call to diffList() on path \"%s\"\n", userPath.c_str());
```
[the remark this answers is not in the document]
Nicogene: `diff_match_patch` throws the exception but I didn't find in which cases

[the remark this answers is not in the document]
damn1: In `recursiveDiff()` there is a call to `patch_toText()` and that's the function that can raise an exception. I decided to catch it here in order to print the error at the same level where the success result is printed.

src/libYARP_companion/src/Companion.cpp
```diff
@@ -1450,8 +1450,8 @@ int Companion::cmdDetectRos(bool write) {
     }
 
     std::string uri = NetworkBase::getEnvironment("ROS_MASTER_URI");
-    if (uri=="") {
-        fprintf(stderr, "ROS_MASTER_URI environment variable not set.\n");
+    if (uri=="" || uri.length() > 512) {
```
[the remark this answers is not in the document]
drdanz: Removed this commit

src/libYARP_OS/src/PortReaderBufferBase.cpp
drdanz: Still not 100% convinced that it is a good idea to merge this change, but anyway...

src/libYARP_OS/src/NameConfig.cpp
```diff
@@ -267,15 +267,12 @@ std::string NameConfig::getHostName(bool prefer_loopback, const std::string& see
...
-#ifdef YARP_HAS_ACE
-#ifdef ACE_ADDR_HAS_LOOPBACK_METHOD
-            would_be_loopback = ips[i].is_loopback();
-#endif
+#if defined YARP_HAS_ACE && defined ACE_ADDR_HAS_LOOPBACK_METHOD
+            bool would_be_loopback = ips[i].is_loopback();
+#else
+            bool would_be_loopback = (ip == "127.0.0.1" ||
+                                      ip == "127.1.0.1" ||
+                                      ip == "127.0.1.1") ? true : false;
```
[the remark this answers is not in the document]
drdanz: Fixed

PR #1838, opened 2018-08-28
Fix BufferedPort::write() after interrupt-resume

### LINKED ISSUE #1834: Issue with interrupt() and resume() on BufferedPorts
Hi All, 

There seems to be a problem with BufferedPorts when calling `BufferedPort::interrupt()`, then `BufferedPort::resume()`. After resuming, BufferedPorts appear to be unable to **send** any data. Any subsequent call to `::write()` fails, as `isClosed()` returns true, even after resuming. 

BufferedPorts used as receivers are unaffected by this issue, and resume operation as expected. 

Simple Ports are also unaffected, as both senders and receivers resume operation as expected. 

#174 and #190 describe similar issues, only with Ports, so the problem may be somewhat similar. Could it be a limitation of the BufferedPort ?

Below a simple test code for the receiver : 
```c++
int main(int argc, char *argv[])
{
    yarp::os::Network yarp;

    yarp::os::BufferedPort<yarp::os::Bottle> port;

    port.open("/in");
    int count = 0;
    bool interrupted_once = false;

    while (count < 10)
    {
        if(port.read())
        {
            count = port.lastRead()->get(0).asInt();
            std::cout << "Received: " <<  count << std::endl;
            ++count;
        }

        if(!interrupted_once)
        {
            port.interrupt();
            yarp::os::Time::delay(1.5);
            port.resume();
            interrupted_once = true;
        }

    }

    port.close();

    return 0;
}
```
And the sender: 
```c++
int main(int argc, char *argv[])
{
    yarp::os::Network yarp;

    yarp::os::BufferedPort<yarp::os::Bottle> port;

    port.open("/out");
    yarp::os::Network::connect("/out", "/in");
    int count = 0;

    while (count < 10)
    {
        yarp::os::Bottle &bot = port.prepare();
        bot.clear();
        bot.addInt(count);
        port.write(true);

        if(count == 5)
        {
            port.interrupt();
        }

        yarp::os::Time::delay(1);

        if(count == 5)
        {
            port.resume();
        }
        ++count;
    }

    port.close();

    return 0;
}
```
Launching the receiver before the sender, we get the following behaviour:
- receiver is interrupted and misses 1 message, but resumes properly, 
-  sender is interrupted at count = 5, then never sends anything after this point, even after resuming. 

**Configuration:**
 - OS: Ubuntu 16.04
 - yarp version: Tested with v2.3.72. I have yet to see if it affects v3.0+.
-  compiler: gcc 5.4.0

Let me know if I can provide any extra details. 

Thanks a lot for your help,

Nicogene:
This PR fixes #1834.
Moreover it add a test to check this fix.

`read()` worked after interrupt/resume because it checks `port.isOpen()`

Please review code.

### DIFF
--- a/doc/release/v3_1_1.md
+++ b/doc/release/v3_1_1.md
@@ -8,6 +8,12 @@ A (partial) list of bug fixed and issues resolved in this release can be found
 Bug Fixes
 ---------
 
+### Libraries
+
+#### YARP_OS
+
+* Fix `write()` in BufferedPort after interrupting-resuming(#1834).
+
 ### Bindings
 
 * Usage of methods that take in input a yarp::sig::Vector in bindings has been
--- a/doc/release/v3_1_1.md
+++ b/doc/release/v3_1_1.md
@@ -10,8 +10,10 @@ Bug Fixes
 
 ### Bindings
 
-* Usage of methods that take in input a yarp::sig::Vector in bindings has been fixed ( https://github.com/robotology/yarp/pull/1828 ).
-* Disable extended analog sensor interfaces in C# to allow compilation of these bindings ( https://github.com/robotology/yarp/pull/1830 ).
+* Usage of methods that take in input a yarp::sig::Vector in bindings has been
+  fixed(#1828).
+* Disable extended analog sensor interfaces in C# to allow compilation of these
+  bindings(#1830).
 
 
 Contributors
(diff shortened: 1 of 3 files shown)


Tobias-Fischer:
LGMT. Travis failed due to unrelated test.

Nicogene:
Push forced with no changes because Appveyor didn't started.

PR #1840, opened 2018-08-29
Fixed typo in the inline definition of yarp::sig::Vector

nunoguedelha:
This implements the changes:
- just insert the `yarp::sig::Vector` typedef into the header portion of the interface file `yarpMATLAB_wrap.cxx`, don't have it interpreted by the SWIG preprocessor, using `%{ ...%}` instead of `%inline`,
- under SWIG processing, define `Vector` as a template using the command:
```
%template(Vector) yarp::sig::VectorOf<double>;
```

Refer to discussion in https://github.com/robotology/yarp/pull/1828.


towardthesea:
Hi @nunoguedelha ,
I don't think this way, using `%{ ...%}` for `Vector`, works as I keep getting error of 
```TypeError: in method ..., argument ... of type 'yarp::sig::Vector const &'``` for testing with Python.

Moreover, checking the generated `yarp.py`, methods with `Vector` as arguments are translated into `yarp::sig::Vector`for example:
```
def getPose(self, *args):
        """
        getPose(ICartesianControl self, yarp::sig::Vector & x, yarp::sig::Vector & o, Stamp stamp=None) -> bool
        getPose(ICartesianControl self, yarp::sig::Vector & x, yarp::sig::Vector & o) -> bool
        getPose(ICartesianControl self, int const axis, yarp::sig::Vector & x, yarp::sig::Vector & o, Stamp stamp=None) -> bool
        getPose(ICartesianControl self, int const axis, yarp::sig::Vector & x, yarp::sig::Vector & o) -> bool
        """
        return _yarp.ICartesianControl_getPose(self, *args)
```
 while the class declaration is only `Vector`:
```
class Vector(VectorBase):
    """Proxy of C++ yarp::sig::VectorOf<(double)> class."""
...
```
For replicating my error, you can first compile python binding for `icub-main` then run one of the tests we designed, eg:
```
python test_ctrlLib.py
```
Cheers,
Phuong

towardthesea:
@Nicogene  @nunoguedelha  You can reuse most of [this test](https://github.com/robotology/icub-main/blob/devel/bindings/tests/test_iKin.py) but remove the lines below 80 and the line `import icub`. It can serve as the test for `cartesian-controller`

Cheers,
P

Nicogene:
Wait to merge, with @towardthesea we found the test that check this issue.
I will soon push to this PR the python/lua/ruby scripts

Nicogene:
This PR actually fixes #1826

nunoguedelha:
> @nunoguedelha I did the small change needed for fixing the issue and squashed in your original commit.
> As soon as the test passes I think we can merge, if you and @traversaro agree.

@Nicogene @towardthesea thanks a lot guys! Sorry I was disconnected from this while dealing with some other hot problems.

Issue #1802, opened 2018-07-13
Serializing maps with thrift idl does not work properly


diegoferigo:
### Describe the bug

Let's start with the following thrift:

```thrift

struct SmallContainer {
    1: string name
}

struct BigContainer {
    1: SmallContainer c1,
    2: SmallContainer c2
}

struct test {
    1: map <string, BigContainer> myMap
}
```

The autogenerated class has problems in serializing the `map`, obtaining a serialization that screws up parentheses.

With the help of @drdanz we figured out the problem is due to the nesting of the value of the map (`BigContainer` is a struct containing two structs).

### Possible solution

Digging in the code, here is what I found. Maps are handled by:

https://github.com/robotology/yarp/blob/716b87f9975b85bf459b325e956fb86a9b38ce7a/src/idls/thrift/src/t_yarp_generator.cc#L2565-L2577

And the relevant part of the `generate_serialize_field` is line 2445 of

https://github.com/robotology/yarp/blob/716b87f9975b85bf459b325e956fb86a9b38ce7a/src/idls/thrift/src/t_yarp_generator.cc#L2429-L2451

In this case `generate_serialize_struct` accepts a `force_nesting` flag which is not set for maps. A possible solution would be transforming 

https://github.com/robotology/yarp/blob/716b87f9975b85bf459b325e956fb86a9b38ce7a/src/idls/thrift/src/t_yarp_generator.cc#L2574

to

```cpp
generate_serialize_field(out, &vfield, "", "", true);
```

This seems working also with non-nested structs, but a more thorough testing would be needed. I'm not an expert on this, and I don't know if this would affect other use cases.

### More details

The autogenerated code is the following:

```cpp
bool test::write_myMap(const yarp::os::idl::WireWriter& writer) const {
  {
    if (!writer.writeMapBegin(BOTTLE_TAG_STRING, BOTTLE_TAG_LIST, static_cast<uint32_t>(test.size()))) return false;
    std::map<std::string, BigContainer> ::const_iterator _iter322;
    for (_iter322 = test.begin(); _iter322 != test.end(); ++_iter322)
    {
      if (!writer.writeListBegin(0,2)) return false;
      if (!writer.writeString(_iter322->first)) return false;
      if (!writer.write(_iter322->second)) return false; // Here is the error
      if (!writer.writeListEnd()) return false;
    }
    if (!writer.writeMapEnd()) return false;
  }
  return true;
}
```

The line that writes the map's value should be:

```cpp
[...]

PR #1848, opened 2018-09-05
Fix serialisation / deserialisation of maps in thrift idl

diegoferigo:
This PR proposes an hotfix for https://github.com/robotology/yarp/issues/1802. For the time being I enabled the flag for handling nested objects in the value of the map. I fear that if the key is a nested object (side case though) we might face similar issues.

I didn't do any ad-hoc testing but I used this fix already for a couple of weeks with complex thrifts without any problem.

cc @drdanz @traversaro


drdanz:
Rebased and merged, thanks.

PR #1852, opened 2018-09-11
Fix "Unable to create directory" error

drdanz:
Also cleanup some nonsense implementations.

### DIFF
--- a/src/libYARP_OS/src/Os.cpp
+++ b/src/libYARP_OS/src/Os.cpp
@@ -47,10 +47,14 @@ int yarp::os::mkdir_p(const char *p, int ignoreLevels)
 
     size_t index = fileName.rfind('/');
     if (index==std::string::npos) {
+#if defined(_WIN32)
         index = fileName.rfind('\\');
         if (index==std::string::npos) {
             return 1;
         }
+#else
+        return 1;
+#endif
     }
     std::string base = fileName.substr(0, index);
     if (yarp::os::stat((char*)base.c_str())<0) {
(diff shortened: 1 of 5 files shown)


src/libYARP_OS/src/Os.cpp
```diff
@@ -41,9 +41,37 @@ int yarp::os::mkdir(const char *p)
...
+    size_t index = fileName.rfind('/');
+    if (index==std::string::npos) {
+#if defined(_WIN32)
+        index = fileName.rfind('\\');
+        if (index==std::string::npos) {
+            return 1;
+        }
+#else
+        return 1;
+#endif
```
Nicogene: It would be nice to use `std::experimental::filesystem::path::preferred_separator` here and in other places, unfortunately is experimental and it is in c++17.

I would suggest to use  "our" `getDirectorySeparator` then I realized why you didn't use it, it is inside `Network.h`.
Maybe we should open an issue about it.

drdanz: 100% agree on this, Network.h is not the right place for this kind of system utilities.

I'd suggest a `yarp::conf::filesystem::path::preferred_separator` constexpr variable, deprecate `getDirectorySeparator` and use `preferred_separator` inside.

The double check (for `/` and `\`) should be performed anyway on windows (even though it should be replaced by [`make_preferred`](https://en.cppreference.com/w/cpp/experimental/fs/path/make_preferred) at some point) but all this part should be replaced by [`create_directory`](https://en.cppreference.com/w/cpp/experimental/fs/create_directory) at some point, so I think it's useless.

For now I'll just proceed and merge this.

PR #1867, opened 2018-09-21
Fix header inclusion (ImplementControlLimits.h)

PeterBowman:
This patch solves ~~two issues~~:

* Broken header inclusion in `ImplementControlLimits2.h`.
* ~~Missing `ImplementControlLimits2.h` in `ControlBoardInterfacesImpl.h`. This used to work prior to YARP 3.0, hence I'm identifying it as a regression. Moreover, I preferred to remove `ImplementControlMode.h` so that it's treated equally with everything else. Alternatively, use deprecation macros:~~
  ```c++
  #ifndef YARP_NO_DEPRECATED
  #include <yarp/dev/ImplementControlLimits2.h> // and so on
  #else
  #include <yarp/dev/ImplementControlLimits.h> // and so on
  #endif
  ```
  (I'm not sure whether it makes sense since those deprecated headers are being installed anyway. Dropping this idea in case it's worth investigating further.)
  **Edit:** dropped from the final patchset (https://github.com/PeterBowman/yarp/commit/401e02126b5f0eee4c370fed4a7627e16b0d7a6f).

### DIFF
--- a/doc/release/v3_1_1.md
+++ b/doc/release/v3_1_1.md
@@ -48,6 +48,7 @@ Bug Fixes
 * Added missing `YARP_dev_API` to `IRobotDescription`.
 * Made optional the view of `IFrameGrabberControls` in `RGBDSensorWrapper`
   (#1875).
+* Fixed header inclusion in `ImplementControlLimits2.h`.
 
 #### `YARP_companion`
 
--- a/src/libYARP_dev/include/yarp/dev/ImplementControlLimits2.h
+++ b/src/libYARP_dev/include/yarp/dev/ImplementControlLimits2.h
@@ -10,7 +10,7 @@
 #define YARP_DEV_ICONTROLLIMITS2IMPL_H
 
 
-#include <yarp/dev/IControlLimitsImpl.h>
+#include <yarp/dev/ImplementControlLimits.h>
 #include <yarp/dev/api.h>
 #include <yarp/os/Log.h>
 


PeterBowman:
I think he's mainly working on the devel branch: https://github.com/drdanz/yarp/commit/009314a87568f7f4e3e34fbb3e6141534b6d536a. Anyway, I could split the first commit, which solves a compilation error, into a separate PR.

drdanz:
I'm not touching dev (only OS) in my branch, so feel free to change whatever you need.
I'm not the right person to review this patch though, @barbalberto, @randaz81, or @Nicogene know this part of code better than me, and know what happened to the `2` interfaces.

PeterBowman:
> I'd remove the "2" headers since that all the `yarp::dev::Interface2` have been deprecated in YARP 3.

I wonder if the first commit (https://github.com/PeterBowman/yarp/commit/5408bcda50131708c162e487572388bab3a9c58a) could be cherry-picked into master so that it makes it into v3.1.1. Right now, I can include the "2" impl headers from YARP 2.x and YARP 3.x dependent code, but said bug forces me to introduce additional macro checks (instead of falling back to `ImplementControlLimits`). I'd rather keep those **impl** headers around until the "2" **interface** headers are finally removed, too. For the same reason, I think they should be included in `ControlBoardInterfacesImpl.h` for now.

PeterBowman:
Done, @Nicogene (I'm not sure what codacy-bot was previously complaining about).

Nicogene:
Merged, thanks!

PR #1881, opened 2018-10-01
Unit tests for checking license

drdanz:
Add unit tests for checking that the license files are correct.

Modify `.travis.yml` so that the tests will not even start if the licenses are not correct. More tests like this can be added later.


Nicogene:
Merged, thanks!

Issue #1893, opened 2018-10-11
Allow to specify a different name for the node name and the yarprun server name in the yarpmanager


S-Dafarra:
The problem is that usually the computer's name is long and not easy to understand, like ``icub@iiticublap123``. This does not hold for the ``yarprun`` server name, which can be called simply ``icub29``. Usually to overcome this problem we have either to change the machine name or to add a host in ``/etc/hosts``. This can clash with internal IT policies, while it may be not ideal to modify a system file. 

Wouldn't it to be possible to specify also the IP address in the ``cluster-config.xml``, something like
```xml
<node display=":0.0" address="10.0.0.29">icub29</node>
```
or
```xml
<node display=":0.0" address="iiticublap123">icub29</node>
```
?

cc @Nicogene @mbrunettini @hu-yue @nunoguedelha @DanielePucci

Nicogene:
Closed by #1894

PR #1894, opened 2018-10-12
Specify a different name for the node name and the yarprun server name in the yarpmanager

S-Dafarra:
Fix https://github.com/robotology/yarp/issues/1893

Maybe also the graphic should be modified, adding a column with the address. It would be even better if it could be modified from the GUI, but I don't really know what to touch.

### DIFF
--- a/doc/release/v3_2_0.md
+++ b/doc/release/v3_2_0.md
@@ -44,6 +44,16 @@ New Features
 * Added `navigation2DServer` device.
 * Added `localization2DServer` device.
 
+New Features
+------------
+
+### GUIs
+
+#### `yarpmanager`
+
+* Allow to specify an address for every host in the `cluster-config.xml` file. 
+  This allow to use host names that cannot be resolved automatically by the network.
+
 Contributors
 ------------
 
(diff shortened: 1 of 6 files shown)


S-Dafarra:
Following the suggestion of @Nicogene I also modified the widget which looks like the following:
image
The fields are not editable. This probably requires some more modifications to avoid editing when the ``yarprun`` is already running.

drdanz:
is this backwards compatible?

S-Dafarra:
If no address is specified, it takes the name as address (see for example the ``pc104`` line in the image above and line https://github.com/robotology/yarp/pull/1894/files#diff-f3936ba50445105e6e7e5a0f8e5fc82eR141) which should correspond to the original behavior.

S-Dafarra:
> If you successfully tested it on a real iCub cluster

Currently @hu-yue and @GiulioRomualdi are using it in the greeny cluster.

> The same thing could be also for the other entries, so if we want to make them editables (and the changes written to `cluster-config.xml`) I think we should do it in a separate PR.

Agreed.

> Only a side note, please rebase and add a line in the changelog

Done

Nicogene:
Merged, thanks!

PR #1897, opened 2018-10-15
check if port is interrupted before resuming

### LINKED ISSUE #1896: calling port.resume() on a non-interrupted port results in system lockup
**Describe the bug**
calling port.resume() on a non-interrupted port results in system hang

**To Reproduce**
```
yarp::os::Port p;
p.open("/myport:i");
p.interrupt();
p.resume();
p.resume(); //lockup here
yInfo() << "Program finished";
p.close();
return 0;
```

**Expected behavior**
I expect it should ignore the second ``resume()`` command

 - OS: ubuntu 16.04 (WSL)
 - yarp version: YARP version 3.1.0+103-20180807.100+gitcae3bea
-  compiler: gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.10)

**Additional context**

It looks like it gets stuck on the semaphores in ``PortCoreAdapter.cpp`` line 113 and 114. My fix is to add
```
if(!core.isInterrupted()) return;
```
on line 401 of ``Port.cpp``, which seems to work in my example. I'm not sure if it has any undesired side-effects.

I will add this fix as a pull request.

arrenglover:
Fixes #1896

### DIFF
--- a/tests/libYARP_OS/PortTest.cpp
+++ b/tests/libYARP_OS/PortTest.cpp
@@ -1582,6 +1582,16 @@ public:
         checkTrue(p.close(),"Closing the broken_device");
     }
 
+    void testPortResumeDeadlock() {
+        report(0, "testing lockup if resume is called when not interrupted");
+        Port p;
+        checkTrue(p.open("/test"), "Checking the open of the port");
+        p.interrupt();
+        p.resume();
+        p.resume(); //should not lockup here
+        p.close();
+    }
+
     virtual void runTests() override {
         NetworkBase::setLocalMode(true);
 
@@ -1648,6 +1658,7 @@ public:
                                                       "brokenDevice",
                                                       "yarp::dev::BrokenDevice"));
         testPrepareDeadlock();
+        testPortResumeDeadlock();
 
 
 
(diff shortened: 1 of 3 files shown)


tests/libYARP_OS/PortTest.cpp
```diff
@@ -1582,6 +1582,16 @@ class PortTest : public UnitTest {
...
     }
 
+    void testPortResumeDeadlock() {
+        report(0, "testing lockup if resume is called when not interrupted");
+        Port p;
+        p.open("/test");
+        p.interrupt();
+        p.resume();
+        p.resume(); //should not lockup here
+        checkTrue(p.close(), "port closed successfully");
```
Nicogene: `p.close()` return void, not a bool it should be like this.
```suggestion
        p.close();
```
if you want to check the compilation of tests in `ccmake` enable `YARP_COMPILE_TESTS`

Nicogene:
I did some cleanup in commit and commits' message rebasing over latest master.
We will merge it as soon as the tests pass :+1:

Nicogene:
Merged, thanks!

PR #1902, opened 2018-10-19
Clang tidy - part 2

drdanz:
* `modernize-use-using`
* `modernize-redundant-void-arg`
* `modernize-loop-convert`
* `modernize-use-emplace`


drdanz:
by the way, the `modernize-loop-convert` with `-fix` is incredible.

PR #1904, opened 2018-10-26
OS: Refactor a few private methods

drdanz:
Current implementation of the `chooseCarrier` method is really confusing.

It accepts a pointer to a `const std::string` and a pointer to a `const yarp::os::Bytes` that can be null, but the method is called wither with the first parameter and the second equal to `nullptr`, or vice versa, therefore I split the method in 2 separate methods, and changed them to use const references instead of pointers.

Anyway, the real problem is that these lines are *really* confusing and hard to understand:

https://github.com/drdanz/yarp/blob/096b771f31599fd06bc5a6f68817be9db199b398/src/libYARP_OS/src/Carriers.cpp#L62-L71

I replaced them with `substr` and recursion.

---

Edit: Added a second commit that does very similar changes in Network.cpp

---

Edit 2: Added a third commit for ResourceFinder

### DIFF
--- a/src/libYARP_OS/src/Carriers.cpp
+++ b/src/libYARP_OS/src/Carriers.cpp
@@ -30,7 +30,28 @@
 using namespace yarp::os::impl;
 using namespace yarp::os;
 
-
+namespace
+{
+static std::string bytes_to_string(const Bytes& header)
+{
+    std::string ret;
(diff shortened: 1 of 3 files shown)


drdanz:
I tested all the cases I could think about with custom strings, I don't think they require to be tested *after* the changes, since now standard STL algorithms are used...

PR #1905, opened 2018-10-27
OS: Pass terminating characters as const char instead of int

drdanz:
Change the signature of the following methods:

* `ConnectionReader::expectText()`
* `ConnectionWriter::appendString()`
* `InputStream::readLine()`

**WARNING**: Very small API break, but I sincerely hope that nobody ever used these methods with an int


drdanz:
> But why was it `int`? :fearful: 

I really have no idea, internally they were considered `char`

PR #1906, opened 2018-10-28
png support

randaz81:
It is now possible to save image files into .png format

### DIFF
--- a/src/libYARP_sig/include/yarp/sig/ImageFile.h
+++ b/src/libYARP_sig/include/yarp/sig/ImageFile.h
@@ -33,14 +33,14 @@ namespace yarp {
                     FORMAT_PNG
                 };
 
-            // PPM/PGM format
+            // read methods
             bool YARP_sig_API read(ImageOf<PixelRgb>& dest,   const std::string& src, image_fileformat format = FORMAT_ANY);
             bool YARP_sig_API read(ImageOf<PixelBgr>& dest,   const std::string& src, image_fileformat format = FORMAT_ANY);
             bool YARP_sig_API read(ImageOf<PixelRgba>& dest,  const std::string& src, image_fileformat format = FORMAT_ANY);
             bool YARP_sig_API read(ImageOf<PixelMono>& dest,  const std::string& src, image_fileformat format = FORMAT_ANY);
             bool YARP_sig_API read(ImageOf<PixelFloat>& dest, const std::string& src, image_fileformat format = FORMAT_ANY);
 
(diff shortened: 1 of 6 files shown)


src/libYARP_sig/src/ImageFile.cpp
```diff
@@ -250,6 +254,93 @@ static bool ImageReadMono(ImageOf<PixelMono> &img, const char *filename)
 // private write methods
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+#if YARP_HAS_PNG_C
+static bool SavePNG(char *src, const char *filename, size_t h, size_t w, size_t rowSize, png_byte color_type, png_byte bit_depth)
+{
```
Nicogene: `src` and `filename` are used but not checked to be `!=nullptr`, I'd like to add this check.

src/libYARP_sig/include/yarp/sig/ImageFile.h
```diff
@@ -29,7 +29,8 @@ namespace yarp {
                     FORMAT_PGM,
                     FORMAT_PPM,
                     FORMAT_JPG,
-                    FORMAT_NUMERIC
+                    FORMAT_NUMERIC,
+                    FORMAT_PNG
                 };
```
Nicogene: Is it missing ` bool YARP_sig_API write(const ImageOf<PixelRgb>& src, const std::string& dest, image_fileformat format = FORMAT_PNG);` ?

src/libYARP_sig/src/ImageFile.cpp
drdanz: Unfortunately this is not checked in the other similar methods as well. I'd start by adding the check here, and later fix the other methods

src/libYARP_sig/include/yarp/sig/ImageFile.h
drdanz: I believe this is correct, it is the comment `// PPM/PGM format` next to these methods that is misleading...

drdanz:
Also it would be very nice if you could add in the `.travis-yml` file
* `libpng-dev` (in the `apt` section)
* `libpng` (the homebrew section, but it requires #1909 to be merged)

src/libYARP_sig/src/ImageFile.cpp
randaz81: libpng is a C library. Its APIs do not allow neither the use of smart pointers, nor fstream.

PR #1913, opened 2018-11-06
Fix error in zfp CMakeList

barbalberto:
the command `target_include_directories` was used instead of `target_link_libraries`
[CI_SKIP]


Nicogene:
Merged, thanks!

PR #1914, opened 2018-11-06
Add HierachicalMutex

Nicogene:
Implementation taken from [C++ Concurrency in Action of Antony Williams](https://www.manning.com/books/c-plus-plus-concurrency-in-action).
It defines a hierarchy among mutexes in order to enforce the lock of
mutexes from the higher value of the lowest value of hierarchy.

An exception is thrown if this order is violated.
It is a proposal, if you think it might be interesting to include in yarp we can merge it.

### DIFF
--- a/src/libYARP_OS/include/yarp/os/HierarchicalMutex.h
+++ b/src/libYARP_OS/include/yarp/os/HierarchicalMutex.h
@@ -1,0 +1,73 @@
+/*
+ * Copyright (C) 2006-2018 Istituto Italiano di Tecnologia (IIT)
+ * All rights reserved.
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD-3-Clause license. See the accompanying LICENSE file for details.
+ */
+
+#ifndef YARP_OS_HIERARCHICALMUTEX_H
+#define YARP_OS_HIERARCHICALMUTEX_H
+
+#include <yarp/os/api.h>
+#include <mutex>
+
+namespace yarp {
+namespace os {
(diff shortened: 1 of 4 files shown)


src/libYARP_OS/include/yarp/os/HierarchicalMutex.h
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD-3-Clause license. See the accompanying LICENSE file for details.
+ */
+
+#ifndef YARP_OS_HIERARCHICALMUTEX_H
+#define YARP_OS_HIERARCHICALMUTEX_H
+
+#include <yarp/os/api.h>
+#include <mutex>
```
claudiofantacci: Is this include necessary here? It should be moved in the implementation file.

src/libYARP_OS/src/HierarchicalMutex.cpp
```diff
@@ -0,0 +1,85 @@
...
+    static thread_local unsigned long this_thread_hierarchy_value;
+
+    void check_for_hierarchy_violation()
+    {
+        if(this_thread_hierarchy_value <= hierarchy_value)
+        {
+            throw std::logic_error("mutex hierarchy violated");
+        }
+    }
+    void update_hierarchy_value()
```
claudiofantacci: For readability, I would separete method implementations with at least one empty line. Pleas ealso check other methods of this PR.

PR #1920, opened 2018-11-13
Fix imu bosh i2c

Nicogene:
Despite the serial, i2c does not put the ack in the first 2 bytes of the
response, so they are not to be discarded.

This device is not tested by CI


src/devices/imuBosch_BNO055/imuBosch_BNO055.cpp
```diff
@@ -98,6 +99,10 @@ bool BoschIMU::open(yarp::os::Searchable& config)
 
     readFunc = i2c_flag ? &BoschIMU::sendReadCommandI2c : &BoschIMU::sendReadCommandSer;
 
+    // In case of reading throught serial the first two bytes of the response are the ack, so
```
gsaponaro: typo: throught -> through

PR #1923, opened 2018-11-22
small fix in RGBDSensorWrapper

valegagge:
swapped the init strings of color and depth parameters in order to maintain their meaning


drdanz:
Right, this should be on `master`... @valegagge please move your branch on master (see robotology/QA#168).
Also it's probably worth a line in the release notes.

valegagge:
@drdanz Is it necessary update the change log also for this small change? :)

valegagge:
I think this is not a bug fix , but only a fix to make more readable the code. The code works fine even if without this change. For this reason, I used devel branch instead of master, but if you (@drdanz  and @Nicogene ) prefer, I can move the commit on master.

drdanz:
Merged, thanks!

valegagge:
Ok thanks

PR #1924, opened 2018-11-22
Fixed annoying warnings caused by Image.h

pattacini:
Since `Image.h` gets included within the user code, the type mismatch causes plenty of annoying warnings.

Don't know whether changing the returned type will cause other warnings in cascade though. If so, we could keep the returned type unchanged and then fix the line `38`.


Nicogene:
Merged, thanks!

PR #1930, opened 2018-12-06
Control board wrapper improvement

barbalberto:
TESTED ON ROBOT

This PR changes how the ControlBoardWrapper refreshes the data from the HW and send them via the output ports state:o and stateExt:o.

This is due to a corner case found, where the position data on the 2 ports were different.
Previous implementation was calling `getEncoders` for one port and `getEncodersTimed` for the other one.
In the low-level device one function was implemented and the other one was not.

Although it is legal for the wrapper to do so (and the bug is actually on the low-level device), it is quite a counter-intuitive and misleading behaviour for the wrapper, and it is better to have a more uniform one.
Since the wrapper during the attach phase enforces the requirement of the `IEncoderTimed` interface only, this is the only one used now.

Furthermore, in the old implemetnation, the `getEncoderXXX` was called up to three times for each run:
- once for state:o
- once for stateExt:o
- once for ROS topic

This was due to the fact that the ports are optional, at runtime the wrapper can be configured to use either YARP port or ROS topic only or both.

Proposed implementation calls getEncoder only once, then the results are stored and sent via ports in the optimal way. This can led to a performance improvement.

### DIFF
--- a/src/libYARP_dev/src/devices/ControlBoardWrapper/ControlBoardWrapper.cpp
+++ b/src/libYARP_dev/src/devices/ControlBoardWrapper/ControlBoardWrapper.cpp
@@ -15,6 +15,8 @@
 #include <yarp/os/Log.h>
 #include <yarp/os/LogStream.h>
 #include <sstream>
+#include <numeric>
+#include <algorithm>
 
 #include <cstring>         // for memset function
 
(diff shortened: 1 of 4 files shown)


barbalberto:
Successfully tested on robot and rebased on latest devel.
Noe this PR can be merged

.travis.yml
```diff
@@ -18,8 +18,8 @@ branches:
     - /^.*appveyor.*$/
 
 stages:
-  - check
-  - test
+#  - check
+#  - test
```
drdanz: Please remove this

barbalberto: Ops, I though I did it on my fork only ...
Anyway I see valgrind still complains, so I'll set WIP in this PR and try to see where is the problem

drdanz:
Merged, thanks.

PeterBowman:
It seems to me this `getTorques` call could vanish, too:

https://github.com/robotology/yarp/blob/bfca0db72d5592e349aa93445459109a6a7df4ba/src/libYARP_dev/src/devices/ControlBoardWrapper/ControlBoardWrapper.cpp#L934

PR #1932, opened 2018-12-07
Add opencv compatibility library

### LINKED ISSUE #1672: YARP does not compile with OpenCV 4
I pulled `opencv` and `yarpdataplayer` and `yarpdatadumper` doesn't compile:
```
/home/ngenesio/yarp/src/yarpdatadumper/main.cpp:418:30: error: use of undeclared identifier 'IplImage'
                int frameW=((IplImage*)itemEnd.obj->getPtr())->width;
                             ^
/home/ngenesio/yarp/src/yarpdatadumper/main.cpp:418:39: error: expected expression
                int frameW=((IplImage*)itemEnd.obj->getPtr())->width;
                                      ^
/home/ngenesio/yarp/src/yarpdatadumper/main.cpp:419:30: error: use of undeclared identifier 'IplImage'
                int frameH=((IplImage*)itemEnd.obj->getPtr())->height;
                             ^
/home/ngenesio/yarp/src/yarpdatadumper/main.cpp:419:39: error: expected expression
                int frameH=((IplImage*)itemEnd.obj->getPtr())->height;
                                      ^
/home/ngenesio/yarp/src/yarpdatadumper/main.cpp:428:52: error: use of undeclared identifier 'CV_FOURCC'
                videoWriter.open(videoFile.c_str(),CV_FOURCC('H','F','Y','U'),
                                                   ^
/home/ngenesio/yarp/src/yarpdatadumper/main.cpp:429:38: error: use of undeclared identifier 'cvSize'
                                 fps,cvSize(frameW,frameH),true);
                                     ^
/home/ngenesio/yarp/src/yarpdatadumper/main.cpp:457:49: error: use of undeclared identifier 'IplImage'
                    cv::Mat img=cv::cvarrToMat((IplImage*)item.obj->getPtr());
                                                ^
/home/ngenesio/yarp/src/yarpdatadumper/main.cpp:457:58: error: expected expression
                    cv::Mat img=cv::cvarrToMat((IplImage*)item.obj->getPtr());
                                                         ^
/home/ngenesio/yarp/src/yarpdatadumper/main.cpp:457:37: error: no member named 'cvarrToMat' in namespace 'cv'
                    cv::Mat img=cv::cvarrToMat((IplImage*)item.obj->getPtr());

```

my `opencv_version` is `4.0.0-pre`.
I'm afraid that in opencv 4 they will remove `IplImage` and `cvarrToMat`.
We should make `yarp::sig::Image` compatible with `cv::Mat` without taking account of `IplImage`.

Nicogene:
In late 2018 we have finally a easy way that allow to get a `cv::Mat` from a `yarp::sig::ImageOf` :tada: 

Since there is no way to retrieve the Pixel code from the `cv::Mat` and the default coding is BGR, this library assumes that when converting to yarp image, the cv::Mat has to be BGR, and the cv::Mat obtained from the reversed conversion is BGR.

Converting from OpenCV to Yarp no copies are involved.
In the other direction is quite tricky to understand when OpenCV performs a copy and when not :confused: 

@drdanz and me found a sort of rule behind their choice to copy or not, but maybe it is not important.
The argument of `toCvMat()` is a `&&` a.k.a a right-reference, this means that after this function call the integrity of the data of the input yarp image is no longer guaranteed. Since we can't actually guarantee it :sweat_smile:  

It can be used easily in cmake projects as `cv` component:

```cmake
find_package(YARP REQUIRED
                  COMPONENTS OS
                             conf
                             sig
                             dev
                             cv)
```

Example of usage:

```c++
ImageOf<PixelRgb> inImg;
// Read it from port
cv::Mat cvImg = toCvMat(std::move(inImg));
// Some OpenCV processing
auto yarpImg = fromCvMat<PixelRgb>(std::move(cvImg));
//  Write on a port
```
Last this PR deprecates `getIplImage` and `wrapIplImage` since they are now obsolete.

It is super WIP and all the suggestions of the CV guys are more than welcome.

It fixes #1672 .

It has been tested with different pixel codes (`PixelRgb`, `PixelRgba`, `PixelMono`, `PixelBgr`, `PixelBgra`), and it works :sunglasses: 

Please review code

### DIFF
--- a/src/libYARP_cv/CMakeLists.txt
+++ b/src/libYARP_cv/CMakeLists.txt
@@ -1,0 +1,41 @@
+# Copyright (C) 2006-2018 Istituto Italiano di Tecnologia (IIT)
+# All rights reserved.
+#
+# This software may be modified and distributed under the terms of the
+# BSD-3-Clause license. See the accompanying LICENSE file for details.
+
+project(YARP_cv)
+
+set(YARP_cv_HDRS include/yarp/cv/Cv.h
+                 include/yarp/cv/Cv-inl.h)
+
+source_group(TREE "${CMAKE_CURRENT_SOURCE_DIR}"
+             PREFIX "Header Files"
+             FILES ${YARP_cv_HDRS})
+
+add_library(YARP_cv INTERFACE)
+add_library(YARP::YARP_cv ALIAS YARP_cv)
+
+target_include_directories(YARP_cv INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+                                             $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
+
+target_link_libraries(YARP_cv INTERFACE YARP::YARP_sig)
+list(APPEND YARP_cv_PUBLIC_DEPS YARP_sig)
+
+target_include_directories(YARP_cv SYSTEM INTERFACE ${OPENCV_INCLUDE_DIR})
+target_link_libraries(YARP_cv INTERFACE ${OPENCV_INCLUDE_DIR})
+list(APPEND YARP_cv_PUBLIC_DEPS OpenCV)
+
+
+install(TARGETS YARP_cv
+        EXPORT YARP_cv
+        COMPONENT YARP_cv
+        PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/yarp/cv)
+
+install(DIRECTORY ${PROJECT_SOURCE_DIR}/include/yarp/cv
+        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/yarp)
+
+include(YarpInstallBasicPackageFiles)
+yarp_install_basic_package_files(YARP_cv
+                                 DEPENDENCIES ${YARP_cv_PUBLIC_DEPS}
+                                 PRIVATE_DEPENDENCIES ${YARP_cv_PRIVATE_DEPS})
--- a/src/libYARP_cv/include/yarp/cv/Cv-inl.h
+++ b/src/libYARP_cv/include/yarp/cv/Cv-inl.h
@@ -1,0 +1,103 @@
+/*
+ * Copyright (C) 2006-2018 Istituto Italiano di Tecnologia (IIT)
+ * All rights reserved.
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD-3-Clause license. See the accompanying LICENSE file for details.
+ */
+
+#ifndef YARP_CV_CV_INL_H
+#define YARP_CV_CV_INL_H
+
+#include<cassert>
+
+namespace yarp {
+namespace cv {
+
+template <typename T>
+struct type_code;
+
+template<> struct type_code<yarp::sig::PixelMono> : std::integral_constant<int, CV_8UC1> {};
+template<> struct type_code<yarp::sig::PixelMono16> : std::integral_constant<int, CV_16UC1> {};
+template<> struct type_code<yarp::sig::PixelMonoSigned> : std::integral_constant<int, CV_8SC1> {};
+template<> struct type_code<yarp::sig::PixelFloat> : std::integral_constant<int, CV_32FC1> {};
+template<> struct type_code<yarp::sig::PixelInt> : std::integral_constant<int, CV_32SC1> {};
+template<> struct type_code<yarp::sig::PixelBgr> : std::integral_constant<int, CV_8UC3> {};
+template<> struct type_code<yarp::sig::PixelRgb> : std::integral_constant<int, CV_8UC3> {};
+template<> struct type_code<yarp::sig::PixelRgbSigned> : std::integral_constant<int, CV_8SC3> {};
+template<> struct type_code<yarp::sig::PixelRgbFloat> : std::integral_constant<int, CV_32FC3> {};
+template<> struct type_code<yarp::sig::PixelRgbInt> : std::integral_constant<int, CV_32SC3> {};
+template<> struct type_code<yarp::sig::PixelHsv> : std::integral_constant<int, CV_8UC3> {};
+template<> struct type_code<yarp::sig::PixelHsvFloat> : std::integral_constant<int, CV_32FC3> {};
+template<> struct type_code<yarp::sig::PixelRgba> : std::integral_constant<int, CV_8UC4> {};
+template<> struct type_code<yarp::sig::PixelBgra> : std::integral_constant<int, CV_8UC4> {};
+
+// -1 : no conversion required.
+template <typename T>
+struct convert_code_to_cv;
+
+template<> struct convert_code_to_cv<yarp::sig::PixelMono> : std::integral_constant<int, -1> {};
+template<> struct convert_code_to_cv<yarp::sig::PixelMono16> : std::integral_constant<int, -1> {};
+template<> struct convert_code_to_cv<yarp::sig::PixelMonoSigned> : std::integral_constant<int, -1> {};
+template<> struct convert_code_to_cv<yarp::sig::PixelFloat> : std::integral_constant<int, -1> {};
+template<> struct convert_code_to_cv<yarp::sig::PixelInt> : std::integral_constant<int, -1> {};
+template<> struct convert_code_to_cv<yarp::sig::PixelBgr> : std::integral_constant<int, -1> {};
+template<> struct convert_code_to_cv<yarp::sig::PixelRgb> : std::integral_constant<int, CV_RGB2BGR> {};
+template<> struct convert_code_to_cv<yarp::sig::PixelRgbSigned> : std::integral_constant<int, CV_RGB2BGR> {};
+template<> struct convert_code_to_cv<yarp::sig::PixelRgbFloat> : std::integral_constant<int, CV_RGB2BGR> {};
+template<> struct convert_code_to_cv<yarp::sig::PixelRgbInt> : std::integral_constant<int, CV_RGB2BGR> {};
+template<> struct convert_code_to_cv<yarp::sig::PixelHsv> : std::integral_constant<int, CV_HSV2BGR> {};
+template<> struct convert_code_to_cv<yarp::sig::PixelHsvFloat> : std::integral_constant<int, CV_HSV2BGR> {};
+template<> struct convert_code_to_cv<yarp::sig::PixelRgba> : std::integral_constant<int, CV_RGBA2BGRA> {};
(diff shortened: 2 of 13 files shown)


src/libYARP_cv/include/yarp/cv/Cv-inl.h
```diff
@@ -0,0 +1,119 @@
...
+ * All rights reserved.
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD-3-Clause license. See the accompanying LICENSE file for details.
+ */
+
+#ifndef YARP_CV_CV_INL_H
+#define YARP_CV_CV_INL_H
+
+#include <yarp/os/LogStream.h>
```
traversaro: Leftover?

src/libYARP_cv/include/yarp/cv/Cv-inl.h
```diff
@@ -0,0 +1,119 @@
...
+template<> struct cvTypeCode<yarp::sig::PixelInt> : std::integral_constant<int, CV_32SC1> {};
+template<> struct cvTypeCode<yarp::sig::PixelBgr> : std::integral_constant<int, CV_8UC3> {};
+template<> struct cvTypeCode<yarp::sig::PixelRgb> : std::integral_constant<int, CV_8UC3> {};
+template<> struct cvTypeCode<yarp::sig::PixelRgbSigned> : std::integral_constant<int, CV_8SC3> {};
+template<> struct cvTypeCode<yarp::sig::PixelRgbFloat> : std::integral_constant<int, CV_32FC3> {};
+template<> struct cvTypeCode<yarp::sig::PixelRgbInt> : std::integral_constant<int, CV_32SC3> {};
+template<> struct cvTypeCode<yarp::sig::PixelHsv> : std::integral_constant<int, CV_8UC3> {};
+template<> struct cvTypeCode<yarp::sig::PixelHsvFloat> : std::integral_constant<int, CV_32FC3> {};
+template<> struct cvTypeCode<yarp::sig::PixelRgba> : std::integral_constant<int, CV_8UC4> {};
+template<> struct cvTypeCode<yarp::sig::PixelBgra> : std::integral_constant<int, CV_8UC4> {};
```
traversaro: Is this mapping between YARP PixelCode and OpenCV pixel codes documented somewhere?

traversaro:
> The argument of `toCvMat()` is a `&&` a.k.a a right-reference, this means that after this function call is not guarantee the integrity of the data of the input yarp image. Since we can't actually guarantee it :sweat_smile:  

I guess this is part of what `&&` means, but given that is a relatively unknown constructor among YARP users (that sometimes are not exactly C++ super-experts) I think this is a comment that it may be worth make in the function docs.

src/libYARP_cv/include/yarp/cv/Cv.h
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+#ifndef YARP_CV_CV_H
+#define YARP_CV_CV_H
+
+#include <yarp/sig/Image.h>
+
+#include <type_traits>
+#include <opencv/cv.h>
+#include <opencv2/imgproc/imgproc.hpp>
```
traversaro: I though all the headers in `opencv2` were actually deprecated. Is this inclusion necessary?

drdanz:
Regarding the BGR/BGRA discussion we had earlier, I think that, in order to avoid data loss, the user should receive a `Mat` that is of the same type as the default for that pixel type, i.e. BGR for types mapping `CV_8UC3` (RGB, BGR), BGRA  for types mapping `CV_8UC4` (RGBA, BGRA), and whatever is the default for all the other types.

The user can eventually perform the conversion he needs of the types is not the one he needs.



The same is valid for the CV to yarp conversion, i.e. we can assume that the type is the default for the type of the `Mat` (i.e. BGR for `CV_8UC3`, BGRA for `CV_8UC4`)

drdanz:
> The argument of toCvMat() is a && a.k.a a right-reference, this means that after this function call is not guarantee the integrity of the data of the input yarp image. Since we can't actually guarantee it 😅

The same should be valid for the `fromCvMat`

drdanz:
Actually, I think this is a good pattern for all the compatibility libraries, we should add the right reference overload everywhere, and deprecate the const reference ones. Also because the `getRawImage() const` method  should return a `const`...

src/libYARP_cv/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+
+source_group(TREE "${CMAKE_CURRENT_SOURCE_DIR}"
+             PREFIX "Header Files"
+             FILES ${YARP_cv_HDRS})
+
+add_library(YARP_cv INTERFACE)
+add_library(YARP::YARP_cv ALIAS YARP_cv)
+
+target_include_directories(YARP_cv INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+                                              $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
```
claudiofantacci: The lines here are not indente properly 😃

src/libYARP_cv/include/yarp/cv/Cv-inl.h
claudiofantacci: I would also add one minor detail.
The convention where OpenCV use `cv` as a prefix to its function is C-style that they are going to change in favour of namespace `cv`. I can't find the discussion about that anymore, but since `cvTypeCode` and `cvCvtCodeFrom` are already in `yarp::cv` namespace, I propose to remove the `cv` prefix to have much cleaner names 😄

src/libYARP_cv/include/yarp/cv/Cv.h
claudiofantacci: What do you mean with deprecated? They are not 😄 
What would be more clean is to `#include <opencv2/imgproc.hpp>` 👍

src/libYARP_cv/include/yarp/cv/Cv-inl.h
```diff
@@ -0,0 +1,119 @@
...
+
+template<typename T>
+yarp::sig::ImageOf<T> yarp::cv::fromCvMat(const ::cv::Mat& cvImage)
+{
+    yarp::sig::ImageOf<T> outImg;
+    if (!std::is_same<T, yarp::sig::PixelBgr>::value)
+    {
+        ::cv::cvtColor(cvImage, cvImage, cvCvtCodeFrom<T>::value);
+    }
+    outImg.resize(cvImage.cols, cvImage.rows); // is it necessary?
```
claudiofantacci: By my understanding, It should not be necessary.

src/libYARP_cv/include/yarp/cv/Cv.h
traversaro: Ok, I probably misunderstood some docs!

claudiofantacci:
Travis builds with `TRAVIS_NO_DEPRECATED = TRUE` fail compiling YARP Image implementation.

src/libYARP_cv/include/yarp/cv/Cv-inl.h
gsaponaro: The naming convention of modern OpenCV is summarized [here](https://docs.opencv.org/3.4.4/df/de5/tutorial_interoperability_with_OpenCV_1.html):

> All the OpenCV related stuff is put into the _cv_ namespace to avoid name conflicts with other libraries data structures and functions. [...]
> Because the functions are already in a namespace there is no need for them to contain the _cv_ prefix in their name. As such all the new C++ compatible functions don't have this and they follow the camel case naming rule. This means the first letter is small (unless it's a name, like _Canny_) and the subsequent words start with a capital letter (like _copyMakeBorder_).

pattacini:
I believe it would be convenient to merge this PR before holidays so that we (at least I) could employ the Christmas break to update the user code.

Nicogene:
Unfortunately we reached a sort of API impasse, maybe someone can help us to overcome it :wink: 

Here is the problem.

Since no copy is involved, the memory is shared between `cv::Mat` and the yarp Image, and conversions(data modification) can be involved.
For both conversion functions: 
- :x: The argument can't be `const &`. The compiler doesn't complain but if I pass a `const &` I expect that the data is not changed by the function, but it is actually changed.
- :x: Pass by `&&` has the problem that we don't have move constructor/operator for `yarp::sig::ImageOf`, and honestly I don't know if it is implemented in `opencv`(probably yes).
- :heavy_check_mark: pass by copy is inefficient, but it has no pitfalls .

What do you think about it?

@drdanz @traversaro @claudiofantacci @pattacini @vtikha

Nicogene:
Discussing with @claudiofantacci and @aerydna with decided to use `&` not const as argument.

claudiofantacci:
@Nicogene thanks for the changes. I think they are ok, however test are failing because there are some things to be fixed in YARP. Are you also fixing them (like the `yarpdataplayer`)?

Nicogene:
`yarpdataplayer` ported and tested :muscle: 

image

`YARP_NO_DEPRECATED` builds fixed.

As soon as the tests passes we can merge from my side.

traversaro:
It is always good to see good old @tanismar , even if just in a dataset. : )

Nicogene:
Tests are now passed, we can merge it as soon as this PR is approved

pattacini:
@Nicogene there's also `yarpdatadumper` that needs to be updated.

Nicogene:
For the `yarpdatadumper` there is this 2e2e40d3819307cd96d59bd28b7104065792e6e4 commit, but I didn't tested it actually. Is it enough that commit in your opinion?

pattacini:
Actually, there are references to IplImage* that I suppose need to be addressed.

pattacini:
Also, I would test the functionality carefully before merging

Nicogene:
I successfully dumped with `yarpdatadumper` the images streamed by the `yarpdataplayer`.

pattacini:
### Caveat

This PR warns the user that `wrapIplImage()` is deprecated in favor of `fromCvMat()`. That's not strictly true since the former function works on `IplImage*` and not on `cv::Mat`.

Then, the proper conversion to help get around this should be:
```c++
IplImage * ipl = ...;
cv::Mat m = cv::cvarrToMat(ipl); 
```

As I've heard `cvarrToMat()` is going to disappear in latest OpenCV releases, we would definitely need to take our time to upgrade the user code (still functional) relying on those legacy OpenCV C-like routines producing `IplImage*` objects (e.g. `cvCreateImage()`, `cvReleaseImage()`...) in favor of their modern OpenCV C++-like counterparts that instead deal directly with `cv::Mat` objects.

The sore point here is that this upgrade can be quite cumbersome, entailing a thorough testing stage.

PR #1933, opened 2018-12-10
recomputeCurrentNavigationPath

randaz81:
Added method `yarp::dev::INavigation2DControlActions::recomputeCurrentNavigationPath()`


src/libYARP_dev/include/yarp/dev/INavigation2D.h
```diff
@@ -127,6 +127,13 @@ class yarp::dev::INavigation2DControlActions
     */
     virtual bool resumeNavigation() = 0;
 
+    /**
+    * Forces the navigation system to recompute the path from the current robot position to the current goal.
+    * If no goal has been set, the command has no effect.
+    * @return true/false
+    */
+    virtual bool recomputeCurrentNavigationPath() = 0;
```
drdanz: Just for reference, this is a breaking change, any 3rd party code implementing `yarp::dev::INavigation2DControlActions` will no longer build after this change.

randaz81: yep. Actually, I'm pretty sure I'm the only implementer of `yarp::dev::INavigation2DControlActions`, since this interface exists since this very last `devel` cycle (not yet released on master)

drdanz: Sorry, I thought this was on `master` already...
Anyway I was in favour of this change anyway :wink:

PR #1945, opened 2019-01-11
Created wrapper of PortReader to provide port subscribing method.

### LINKED ISSUE #1927: Provide callback function to Port::setReader
**Is your feature request related to a problem? Please describe.**
When a class has several readers, a secondary processor class is required for new state callback.

**Describe the solution you'd like**
Simplify creating new state callback:
`
    auto portIn = std::unique_ptr<yarp::os::Port>(new yarp::os::Port);
    portIn->open("/foo");
    portIn->setReader([this](yarp::os::ConnectionReader& connection){
        // do something
    });
`
also providing port name would simplify:
`
    auto portIn = std::unique_ptr<yarp::os::Port>(new yarp::os::Port("/foo"));
    portIn->setReader([this](yarp::os::ConnectionReader& connection){
        // do something
    });
`

**Describe alternatives you've considered**
Extending port class with another class and using the new class to create ports.

jbbjarnason:
Created a wrapper class of PortReader which implements virtual read function, this implementation calls a callback if provided.
There is no implementation to unsubscribe, is that required ?

Fixes #1927

### DIFF
--- a/src/libYARP_OS/include/yarp/os/Port.h
+++ b/src/libYARP_OS/include/yarp/os/Port.h
@@ -131,6 +131,13 @@ public:
     void setAdminReader(PortReader& reader) override;
 
     /**
+     * Subscribe to port reader new state with given callback.
+     *
+     * @param callback can be provided for example in lambda format [](ConnectionReader& conn){}
+     */
+    void setReaderCallback(const ReaderDataProcessor::readCallback& callback);
+
+    /**
      * Set a creator for readers for port data.
      *
      * Every port that input is received from will be automatically
@@ -258,6 +265,7 @@ public:
 private:
     void* implementation;
     bool owned;
+    ReaderDataProcessor dataProcessor;
 
     void* needImplementation() const;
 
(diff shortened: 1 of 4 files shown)


src/libYARP_OS/include/yarp/os/Port.h
```diff
@@ -258,6 +265,7 @@ class YARP_OS_API Port : public UnbufferedContactable
 private:
     void* implementation;
     bool owned;
+    ReaderDataProcessor dataProcessor;
```
drdanz: This and the `ReaderDataProcessor` class should be hidden inside the implementation,

src/libYARP_OS/include/yarp/os/Port.h
```diff
@@ -130,6 +130,13 @@ class YARP_OS_API Port : public UnbufferedContactable
     // Documented in Contactable
     void setAdminReader(PortReader& reader) override;
 
+    /**
+     * Subscribe to port reader new state with given callback.
+     *
+     * @param callback can be provided for example in lambda format [](ConnectionReader& conn){}
```
drdanz: I think you should add some documentation explaining that setReader and setReaderCallback are mutually exclusive, and by calling one of the two methods the previous callback is removed.

PR #1950, opened 2019-01-16
yarpmanager: optimize refresh of connection status and various improvements.

Nicogene:
The existence of source and destination of a connection was requested twice. Each request was a communication with the nameserver.

A flag about the existence has been added.

I tested on my laptop and it works, also with udp connections when the source disappear.
it has to be tested deeply to avoid pitfalls.
please review code


src/libYARP_manager/src/manager.cpp
```diff
@@ -1021,7 +1021,9 @@ bool Manager::connect()
     //connector.init();
     CnnIterator cnn;
     for(cnn=connections.begin(); cnn!=connections.end(); cnn++) {
-        if( !connector.connect((*cnn).from(), (*cnn).to(),
+        if( !(*cnn).getFromExists() ||
+            !(*cnn).getFromExists() ||
```
Tobias-Fischer: Should that be `getToExists()`?

src/libYARP_manager/src/manager.cpp
```diff
@@ -1134,7 +1136,9 @@ bool Manager::connected()
     CnnIterator cnn;
     bool bConnected = true;
     for(cnn=connections.begin(); cnn!=connections.end(); cnn++)
-        if( !connector.connected((*cnn).from(), (*cnn).to(), (*cnn).carrier()) )
+        if( !(*cnn).getFromExists() ||
+            !(*cnn).getFromExists() ||
```
Tobias-Fischer: Should that be `getToExists()`?

lornat75:
thanks @Nicogene I recommend a strong testing on the setups, these things are very tricky!

src/libYARP_manager/src/manager.cpp
Nicogene: Good catch! Copy paste error :sweat_smile:

Nicogene:
Edit: Added a messagebox when trying to stop `yarpserver` while `yarprun` are still running

Nicogene:
Edit2: cluster restyled 

image

Nicogene:
This patch has been tested on `iCubErzelli01` setup, also with udp connections and it works, ready to be merged.

PR #1958, opened 2019-02-03
new Frame Transform interfaces + ovrdevice refactoring

aerydna:
~~This pr do a **lot** of different things. unfortunately all of them are strictly related so it cannot be splitted in different pr (or it's very painful doing it). also, splitting it in different pr makes very unclear the sense of some modifications.~~

the necessity of this pr emerged with the urge to treat devices that produces frames (poses in the 3D space), like ovrdevice and vicon device, in the standard yarp way (with the interface + wrapper + client combo, P2P paradigm).
Until now, a device that produces frames had to open a `frameTransformClient` inside and comunicate via the `frameTransformServer` to other clients via a **Client-Server-Client** architecture, different from the usual **P2P** comunication pattern.
while this has some advantages it also forces a network comunication via yarp ports even when **it is not needed** (so it is **not** possible to open the device locally and get his data via c++ calls). In order to solve this i developed two new interfaces that can be adapted to **both** paradigm.

`IFrameSource`: this particular interface can seems unusual cause it does not have any pubblic pure virtual method. This is because all the beaviour and the math involved in chaining the Frames toghether can be **reused** among all the devices and it is already implemented. instead it have a sort of **protected intra-class interface**, to allow a correct comunication between the parent class and the derived calsses, **enforcing** a lot of stuff and robustifying the architecture. in particular this is useful to build up a **callback mechanism** that allow the user to trigger a function call on every update of the device (even remotely).

pr modification list:

~~- introduces a dependency in yarp to a **new** library of robotology called `RobotologyTemplateLibrary`, which contains some utilities that should not stay in yarp but since the library is still an **unstable stub**, the dependency is coordinated by the new flag `YARP_ENABLE_UNSTABLE`. we cannot wait for the library to be stable, cause the development of this new library is strictly related to his possible usage. and it is only used in this pr at the moment, and it will likely used fisrt in yarp. with @traversaro we agreed to flag the new library as **stable** within **max 2 releases** of yarp and from then continue to develop the library with the usual deprecation policies. moving the dependency outside the `YARP_ENABLE_UNSTABLE` flag.~~ 
_the work on this became **way too** time expensive and some minor importance elements have been reduced_

~~- definition of a new global flag `YARP_EXPERIMENTAL`. useful to confine part of the code that uses the new library or that are in a **experimental** state.~~

- ~~with a tool offered by the new library,~~ two new interfaces has been developed. Actually they come from the IFrameTransform interface, splitted in two parts. one for devices which **only** produces frames, **IFrameSource**, ~~and one that rappresent the version 2 of the `IFrameTransform`~~, _and one for set some tf via function call from the user application (IFrameSet.. i know the name i ugly and misleading. go on, propose a new one!)._

- yarp::math::FrameTransform name issue. two new name to identify the frame structure has been added: `frameId `and `parentFrame`. this should substitute the old `dst_frame_id` and `src_frame_id` _which have been deprecated_. the new implementation solve the #1775 

- `ovrdevice `has been adapted to the new paradigm, deriving from `iFrameSource`, eliminating both the inclusion of the FrameTransformClient and the yarp port to publish the headset pose. ~~the new implementation is confined by `YARP_EXPERIMENTAL`, otherwise the old implementation is active~~. this allows to use it without the transform server, taking the data directly from him via c++ code or remotely in a p2p via the **network wrappers**

- `FrameReceiver `the client side network wrapper.

- `FrameBroadcaster `the produces side network wrapper.

these last three devices support the new **callback** system. so it is possible to have a user application triggered and **syncronized** with the `ovrdevice `thread.

26/03/2019 _the network wrappers comunication is made via **publisher** and **subscriber** so the distribution over multiple subscriber of the tf data is guaranteed by the topic mechanism. also the ros comunication can be archieved via topics because the message used by the network wrapper is `yarp::rosmsg::tf2_msgs::TFMessage`._

notes:
- everithing is still to be tested
~~- i think in the future, the FrameTransformClient should implement the new `IFrameManager `interface, which is Client-Server-client compliant. @randaz81 told me it is used a lot outside yarp so i leaved it untouched.~~ _at this point the FrameTransform client and server can be dismissed.._
~~- **no ROS** compatibility have been added in the `FrameBroadcaster `for now. it can be done in a future pr as it to be quick and easy.~~ _ROS compatibility added_
~~- a lot of interoperations with the other paradigm (FrameTransformClient/Server) are possible, though none of them have been developed at the moment. ~~ 
_no interoperation with the old paradigm is needed anymore._
~~- since there are probably **762** different ways to do the cmake parts, i probably picked the **wrong** one and i presume it have to be corrected. so if you have suggestions regarding the _"cmake-shining-compliant"_  way to accomplish what i've done don't be afraid of fullfill my lack of experience with cmake.~~ the cmake part is very small right now
~~- if it isn't clear the relation between all the element of the pr is the following: _RobotologyTemplateLibrary_ `Result `type is used in the ----> new interfaces to manage Frames which have been developed to change the comunication paradigm of the -----> `ovrdevice `and Others. viceversa the ovrdevice stands as an example of the new implementation.~~ _the pr is much more simpler_
- fixes the dic-iit/element_retargeting_from_human#72 and it is a preparation for the 73 and 74.
- there will be, for sure, someone who don't like the names of the new flags/interfaces/something and has better alternative. not a problem for me. Just decide  the names you like and i will change it.
- the changelog and ~~the method documentation~~ will be added/updated after the review process (i expect it to be **long**)
- sorry for possible english mistake

@GiulioRomualdi @DanielePucci @drdanz @Nicogene @randaz81 @traversaro @barbalberto

### DIFF
--- a/src/devices/FrameBroadcaster/CMakeLists.txt
+++ b/src/devices/FrameBroadcaster/CMakeLists.txt
@@ -1,0 +1,40 @@
+# Copyright (C) 2006-2019 Istituto Italiano di Tecnologia (IIT)
+# All rights reserved.
+#
+# This software may be modified and distributed under the terms of the
+# BSD-3-Clause license. See the accompanying LICENSE file for details.
+
+yarp_prepare_plugin(FrameBroadcaster
+                    CATEGORY device
+                    TYPE yarp::dev::FrameBroadcaster
+                    INCLUDE FrameBroadcaster.h)
+
+if(ENABLE_FrameBroadcaster)
+
+  set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+  yarp_add_plugin(yarp_FrameBroadcaster FrameBroadcaster.cpp
+                                        FrameBroadcaster.h)
+
+   target_link_libraries(yarp_FrameBroadcaster PRIVATE YARP::YARP_OS
+                                                       YARP::YARP_sig
+                                                       YARP::YARP_dev
+                                                       YARP::YARP_math
+                                                       YARP_rosmsg)
+  list(APPEND YARP_${YARP_PLUGIN_MASTER}_PRIVATE_DEPS YARP_OS
+                                                      YARP_sig
+                                                      YARP_dev
+                                                      YARP_math
+                                                      YARP_rosmsg)
+
+  yarp_install(TARGETS yarp_FrameBroadcaster
+               EXPORT YARP_${YARP_PLUGIN_MASTER}
+               COMPONENT ${YARP_PLUGIN_MASTER}
+               LIBRARY DESTINATION ${YARP_DYNAMIC_PLUGINS_INSTALL_DIR}
+               ARCHIVE DESTINATION ${YARP_STATIC_PLUGINS_INSTALL_DIR}
+               YARP_INI DESTINATION ${YARP_PLUGIN_MANIFESTS_INSTALL_DIR})
+
+  set(YARP_${YARP_PLUGIN_MASTER}_PRIVATE_DEPS ${YARP_${YARP_PLUGIN_MASTER}_PRIVATE_DEPS} PARENT_SCOPE)
+
+  set_property(TARGET yarp_FrameBroadcaster PROPERTY FOLDER "Plugins/Device")
+endif()
--- a/src/devices/FrameReceiver/CMakeLists.txt
+++ b/src/devices/FrameReceiver/CMakeLists.txt
@@ -1,0 +1,40 @@
+# Copyright (C) 2006-2019 Istituto Italiano di Tecnologia (IIT)
+# All rights reserved.
+#
+# This software may be modified and distributed under the terms of the
+# BSD-3-Clause license. See the accompanying LICENSE file for details.
+
+yarp_prepare_plugin(FrameReceiver
+                    CATEGORY device
+                    TYPE yarp::dev::FrameReceiver
+                    INCLUDE FrameReceiver.h)
+
+if(ENABLE_FrameReceiver)
+
+  set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+  yarp_add_plugin(yarp_FrameReceiver FrameReceiver.cpp
+                                     FrameReceiver.h)
+
+  target_link_libraries(yarp_FrameReceiver PRIVATE YARP::YARP_OS
+                                                   YARP::YARP_sig
+                                                   YARP::YARP_dev
+                                                   YARP::YARP_math
+                                                   YARP_rosmsg)
+  list(APPEND YARP_${YARP_PLUGIN_MASTER}_PRIVATE_DEPS YARP_OS
+                                                      YARP_sig
+                                                      YARP_dev
+                                                      YARP_math
+                                                      YARP_rosmsg)
+
+  yarp_install(TARGETS yarp_FrameReceiver
+               EXPORT YARP_${YARP_PLUGIN_MASTER}
+               COMPONENT ${YARP_PLUGIN_MASTER}
+               LIBRARY DESTINATION ${YARP_DYNAMIC_PLUGINS_INSTALL_DIR}
+               ARCHIVE DESTINATION ${YARP_STATIC_PLUGINS_INSTALL_DIR}
+               YARP_INI DESTINATION ${YARP_PLUGIN_MANIFESTS_INSTALL_DIR})
+
+  set(YARP_${YARP_PLUGIN_MASTER}_PRIVATE_DEPS ${YARP_${YARP_PLUGIN_MASTER}_PRIVATE_DEPS} PARENT_SCOPE)
+
+  set_property(TARGET yarp_FrameReceiver PROPERTY FOLDER "Plugins/Device")
+endif()
--- a/src/devices/ovrheadset/OVRHeadset.h
+++ b/src/devices/ovrheadset/OVRHeadset.h
@@ -94,18 +94,18 @@ public:
     virtual ~OVRHeadset();
 
     // yarp::dev::DeviceDriver methods
-    virtual bool open(yarp::os::Searchable& cfg);
-    virtual bool close();
+    bool open(yarp::os::Searchable& cfg) override;
+    bool close() override;
 
     // yarp::os::RateThread methods
-    virtual bool threadInit();
-    virtual void threadRelease();
-    virtual void run();
+    bool threadInit() override;
+    void threadRelease() override;
+    void run() override;
 
     //  yarp::dev::IService methods
-    virtual bool startService();
-    virtual bool updateService();
-    virtual bool stopService();
+    bool startService() override;
+    bool updateService() override;
+    bool stopService() override;
 
     // yarp::dev::IJoypadController methods
     bool getAxisCount(unsigned int& axis_count) override;
@@ -142,7 +142,6 @@ private:
     void fillHatStorage();
     void resetInput();
 
-    FlexImagePort* gui_ports{ nullptr };
     std::vector<guiParam> huds;
     InputCallback* displayPorts[2]{ nullptr, nullptr };
     ovrEyeRenderDesc EyeRenderDesc[2];
@@ -169,7 +168,7 @@ private:
         {&rightFrame, &ts.HandPoses[ovrHand_Right]},
         {&leftFrame, &ts.HandPoses[ovrHand_Left]  },
         {&headFrame, &headpose },
-        {&(headFrame + "_predicted"), &predicted_headpose }
+        {&headFramePredicted, &predicted_headpose }
     };
 
     yarp::os::Mutex                  inputStateMutex;
@@ -185,6 +184,7 @@ private:
     std::string rightFrame;
     std::string rootFrame;
     std::string headFrame;
+    std::string headFramePredicted;
 
     bool closed{ false };
     long long distortionFrameIndex{ 0 };
--- a/src/devices/ovrheadset/OVRHeadset.h
+++ b/src/devices/ovrheadset/OVRHeadset.h
@@ -25,6 +25,9 @@
 #include <yarp/dev/DeviceDriver.h>
 #include <yarp/dev/ServiceInterfaces.h>
 #include <yarp/dev/IJoypadController.h>
+
+#include <yarp/dev/IFrameSource.h>
+#include <mutex>
 #include <yarp/dev/IFrameTransform.h>
 #include <yarp/dev/PolyDriver.h>
 #include <yarp/sig/Image.h>
@@ -34,6 +37,7 @@
 #include <OVR_CAPI_GL.h>
 #include <map>
 #include <vector>
+
 
 /**
 * @ingroup dev_impl_other
@@ -78,7 +82,13 @@ class OVRHeadset : public yarp::dev::DeviceDriver,
                    public yarp::os::PeriodicThread,
                    public yarp::dev::IService,
                    public yarp::dev::IJoypadController
+                  ,public yarp::dev::ImplementIFrameSource
 {
+protected:
+    void updateFrameContainer(FrameEditor& frameContainer) override;
+    bool callbackPrepare() override;
+    bool callbackDismiss() override;
+
 public:
     explicit OVRHeadset();
     virtual ~OVRHeadset();
@@ -132,20 +142,6 @@ private:
     void fillHatStorage();
     void resetInput();
 
-
-    yarp::os::BufferedPort<yarp::os::Bottle>* orientationPort;
-    yarp::os::BufferedPort<yarp::os::Bottle>* positionPort;
-    yarp::os::BufferedPort<yarp::os::Bottle>* angularVelocityPort;
-    yarp::os::BufferedPort<yarp::os::Bottle>* linearVelocityPort;
-    yarp::os::BufferedPort<yarp::os::Bottle>* angularAccelerationPort;
-    yarp::os::BufferedPort<yarp::os::Bottle>* linearAccelerationPort;
-    yarp::os::BufferedPort<yarp::os::Bottle>* predictedOrientationPort;
-    yarp::os::BufferedPort<yarp::os::Bottle>* predictedPositionPort;
-    yarp::os::BufferedPort<yarp::os::Bottle>* predictedAngularVelocityPort;
-    yarp::os::BufferedPort<yarp::os::Bottle>* predictedLinearVelocityPort;
-    yarp::os::BufferedPort<yarp::os::Bottle>* predictedAngularAccelerationPort;
-    yarp::os::BufferedPort<yarp::os::Bottle>* predictedLinearAccelerationPort;
-    
     FlexImagePort* gui_ports{ nullptr };
     std::vector<guiParam> huds;
     InputCallback* displayPorts[2]{ nullptr, nullptr };
@@ -166,6 +162,16 @@ private:
     ovrPoseStatef predicted_headpose;
     unsigned int guiCount;
     bool         enableGui{ true };
+    bool threadIdValid;
+    std::mutex frameMutex;
+    std::vector<std::pair<std::string*, ovrPoseStatef*>> poses
+    {
+        {&rightFrame, &ts.HandPoses[ovrHand_Right]},
+        {&leftFrame, &ts.HandPoses[ovrHand_Left]  },
+        {&headFrame, &headpose },
+        {&(headFrame + "_predicted"), &predicted_headpose }
+    };
+
     yarp::os::Mutex                  inputStateMutex;
     ovrInputState                    inputState;
     bool                             inputStateError{ false };
@@ -175,12 +181,10 @@ private:
     std::map<int, int>               DButtonToHat;
     std::map<ovrResult, std::string> error_messages;
 
-    IFrameTransform* tfPublisher;
-    bool             relative;
-    std::string      left_frame;
-    std::string      right_frame;
-    std::string      root_frame;
-    PolyDriver       driver;
+    std::string leftFrame;
+    std::string rightFrame;
+    std::string rootFrame;
+    std::string headFrame;
 
     bool closed{ false };
     long long distortionFrameIndex{ 0 };
--- a/src/libYARP_dev/CMakeLists.txt
+++ b/src/libYARP_dev/CMakeLists.txt
@@ -105,7 +105,9 @@ set(YARP_dev_HDRS include/yarp/dev/all.h
                   include/yarp/dev/Wrapper.h)
 
 if(TARGET YARP::YARP_math)
-  list(APPEND YARP_dev_HDRS include/yarp/dev/IFrameTransform.h
+  list(APPEND YARP_dev_HDRS include/yarp/dev/IFrameSet.h
+                            include/yarp/dev/IFrameSource.h
+                            include/yarp/dev/IFrameTransform.h
                             include/yarp/dev/ILocalization2D.h
                             include/yarp/dev/IMap2D.h
                             include/yarp/dev/INavigation2D.h
@@ -163,6 +165,7 @@ set(YARP_dev_SRCS src/CanBusInterface.cpp
 if(TARGET YARP::YARP_math)
   list(APPEND YARP_dev_SRCS src/IFrameTransform.cpp
                             src/IMap2D.cpp
+                            src/ImplementIFrameSource.cpp
                             src/INavigation2D.cpp
                             src/MapGrid2D.cpp
                             src/Map2DArea.cpp)
--- a/src/libYARP_dev/include/yarp/dev/IFrameSource.h
+++ b/src/libYARP_dev/include/yarp/dev/IFrameSource.h
@@ -220,18 +220,18 @@ protected:
      this should be set to true and false in the derived class to to flag the frames present in the container as obsolete and request a new update.
      tipically the implementer will set it to false when before getting/receiving new transforms and set it to true at the end of updateFrameContainer()
     */
-    std::atomic_bool cacheValid;
+    std::atomic_bool cacheValid{false};
     
     /**
      the method to be called from the derived classes to call the callbacks.
     */
     bool callAllCallbacks()
     {
-        if (!callbacks.size())
+        if (callbacks.empty())
             return false;
 
         updateFrameContainer(frameContainer);
-        for (auto callback : callbacks)
+        for (const auto& callback : callbacks)
             if (callback.second)
                 callback.second((IFrameSource*)this);
         return true;
--- a/src/libYARP_dev/include/yarp/dev/IFrameSource.h
+++ b/src/libYARP_dev/include/yarp/dev/IFrameSource.h
@@ -34,8 +34,13 @@ public:
     class Result
     {
     public:
-        bool        valid{ false };
+#if (defined _WIN32 && _MSC_VER < 1910) || (!defined _WIN32 && __cplusplus < 201402L)
+        bool        valid;
+#else
+        bool        valid{false};
+#endif
         int         error;
+
         T           value;
         //adding a operator bool(){return valid;} is a very bad idea as it can cause confusion between the result of bool functions and their success
     };
@@ -186,10 +191,15 @@ protected:
         yarp::math::FrameTransform nullframe;
 
     public:
+#if (defined _WIN32 && _MSC_VER < 1910) || (!defined _WIN32 && __cplusplus < 201402L)
+        const std::map<std::string, yarp::math::FrameTransform>::const_iterator begin() const { return storage.begin(); }
+        const std::map<std::string, yarp::math::FrameTransform>::const_iterator end() const { return storage.end(); }
+        const size_t erase(const std::string& k) { return storage.erase(k); }
+#else
         const auto begin() const { return storage.begin(); }
         const auto end() const { return storage.end(); }
         const auto erase(const std::string& k) { return storage.erase(k); }
-    
+#endif
         bool insertUpdate(const yarp::math::FrameTransform& frame)
         {
             storage[frame.frameId] = frame;
@@ -224,6 +234,7 @@ protected:
         for (auto callback : callbacks)
             if (callback.second)
                 callback.second((IFrameSource*)this);
+        return true;
     }
     
     /**
@@ -265,8 +276,8 @@ public:
     virtual bool unsetCallback(const int& id) override;
     virtual Result<int> setCallback(std::function<bool(IFrameSource*)> cb) override;
     virtual std::string allFramesAsString() override;
-    virtual bool clearOlderFrames(const std::chrono::milliseconds& lifeMax);
-    virtual bool clearStaticFrames();
+    virtual bool clearOlderFrames(const std::chrono::milliseconds& lifeMax) override;
+    virtual bool clearStaticFrames() override;
     virtual Result<bool> canTransform(const std::string &target_frame, const std::string &source_frame) override;
     virtual bool frameExists(const std::string &frame_id) override;
     virtual std::unordered_set<std::string> getAllFrameIds() override;
--- a/src/libYARP_dev/include/yarp/dev/IFrameSource.h
+++ b/src/libYARP_dev/include/yarp/dev/IFrameSource.h
@@ -1,0 +1,282 @@
+/*
+ * Copyright (C) 2006-2019 Istituto Italiano di Tecnologia (IIT)
+ * All rights reserved.
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD-3-Clause license. See the accompanying LICENSE file for details.
+ */
+
+#ifndef YARP_DEV_IFRAMESOURCE_H
+#define YARP_DEV_IFRAMESOURCE_H
+
+#include <unordered_set>
+#include <map>
+#include <vector>
+#include <functional>
+#include <yarp/dev/api.h>
+#include <yarp/dev/IFrameSet.h> // <---useless include.. but without it IFrameSet doesn't export his symbol. to investigate
+#include <yarp/math/Quaternion.h>
+#include <yarp/math/FrameTransform.h>
+#include <thread>
+#include <atomic>
+
+namespace yarp {
+namespace dev {
+class IFrameSource;
+class ImplementIFrameSource;
+}
+}
+
+class YARP_dev_API yarp::dev::IFrameSource
+{
+public:
+    template <typename T>
+    class Result
+    {
+    public:
+        bool        valid{ false };
+        int         error;
+        T           value;
+        //adding a operator bool(){return valid;} is a very bad idea as it can cause confusion between the result of bool functions and their success
+    };
+
+    enum errors
+    {
+        OK = 0,
+        FRAME_NOT_FOUND,
+        TARGET_NOT_FOUND,
+        SOURCE_NOT_FOUND,
+        FRAMES_NOT_CONNECTED,
+        WRONG_INPUT_FORMAT,
+        TIMEOUT,
+        CALLBACK_ACTIVE,
+        UNDEFINED
+    };
+
+    /**
+    Disable a specific callback
+    * @param id the ID of the callback to be removed. received with setCallback()
+    * @return true if succeded false otherwise
+    */
+    virtual bool unsetCallback(const int& id) = 0;
+
+    /**
+    enable a callback system. the data producer will call all the callback at every data update
+    * @param cb the callback to be set
+    * @return Result<int> the ID of the callback (useful to unset it) wrapped by a result object containing the operation success or failure
+    */
+    virtual Result<int> setCallback(std::function<bool(IFrameSource*)> cb) = 0;
+    
+    /**
+    Creates a debug string containing the list of all registered frames.
+    * @return std::string
+    */
+    virtual std::string allFramesAsString() = 0;
+
+    /**
+    Removes all frames older than lifeMax. frame with timestamp 0.0 are considered static and will not be removed
+    * @return true if any frame has been removed, otherwise
+    */
+    virtual bool clearOlderFrames(const std::chrono::milliseconds& lifeMax) = 0;
+
+    /**
+    Removes all static frames.
+    * @return true if any frame has been removed, otherwise
+    */
+    virtual bool clearStaticFrames() = 0;
+
+    /**
+    Test if a transform exists.
+    * @param target_frame_id the name of target reference frame
+    * @param source_frame_id the name of source reference frame
+    * @return Result<bool>. if the target or the source frame are not present the result will be negative and the value inside should be ignored. if both the frame are correct but not connected 
+    *the result will be positive and the value negative. otherwise both positive
+    */
+    virtual Result<bool> canTransform(const std::string &target_frame, const std::string &source_frame) = 0;
+
+    /**
+    Check if a frame exists.
+    * @param frame_id the frame to be searched
+    * @return true if it exists, false otherwise
+    */
+    virtual bool frameExists(const std::string &frame_id) = 0;
+
+    /**
+    Gets a set containing all the frame_id.
+    * @return the set of frame ids
+    */
+    virtual std::unordered_set<std::string> getAllFrameIds() = 0;
+
+    /**
+    Gets a vector containing all the registered frames.
+    * @return the vector
+    */
+    virtual std::vector<yarp::math::FrameTransform> getAllFrames() = 0;
+
+    /**
+     Get the parent of a frame.
+    * @param frame_id the name of target reference frame
+    * @return a valid outcome will contain the parent frame. if the frame_id doesn't exist an invalid outcome should be returned
+    */
+    virtual Result<std::string> getParent(const std::string& frame_id) = 0;
+
+    /**
+     Get the transform between two frames.
+    * @param target_frame_id the name of target reference frame
+    * @param source_frame_id the name of source reference frame
+    * @param time point in which retrieve the data (if minor than 0 the last data will be used)
+    * @return a valid outcome will contain the frameTransform. if a frame doesn't exist or the frame are not connected an invalid outcome should be returned
+    */
+    virtual Result<yarp::math::FrameTransform> getTransform(const std::string &target_frame_id, const std::string &source_frame_id, double timestamp = -1) = 0;
+
+    /**
+    Transform a point into the target frame.
+    * @param target_frame_id the name of target reference frame
+    * @param source_frame_id the name of frame in which input_point is expressed
+    * @param input_point the input point (x y z)
+    * @param time point in which retrieve the data (if minor than 0 the last data will be used)
+    * @return the point transformed
+    */
+    virtual Result<yarp::sig::Vector> transformPoint(const std::string &target_frame_id, const std::string &source_frame_id, const yarp::sig::Vector &input_point, double timestamp = -1) = 0;
+
+    /**
+     Transform a Stamped Pose into the target frame.
+    * @param target_frame_id the name of target reference frame
+    * @param source_frame_id the name of frame in which input_pose is expressed
+    * @param input_pose the input quaternion (x y z r p y)
+    * @param time point in which retrieve the data (if minor than 0 the last data will be used)
+    * @return the pose transformed
+    */
+    virtual Result<yarp::sig::Vector> transformPose(const std::string &target_frame_id, const std::string &source_frame_id, const yarp::sig::Vector &input_pose, double timestamp = -1) = 0;
+
+    /**
+     Transform a quaternion into the target frame.
+    * @param target_frame_id the name of target reference frame
+    * @param source_frame_id the name of frame in which input_quaternion is expressed
+    * @param input_quaternion the input quaternion (x y z w)
+    * @param time point in which retrieve the data (if minor than 0 the last data will be used)
+    * @return the quaternion transformed
+    */
+    virtual Result<yarp::math::Quaternion> transformQuaternion(const std::string &target_frame_id, const std::string &source_frame_id, const yarp::math::Quaternion &input_quaternion, double timestamp = -1) = 0;
+
+    /**
+     Block until a transform from source_frame_id to target_frame_id is possible or it times out. returns the frame transform requested
+    * @param target_frame_id the name of target reference frame
+    * @param source_frame_id the name of source reference frame
+    * @param timeout (in seconds) to wait for
+    * @return the frameTransform
+    */
+    virtual Result<yarp::math::FrameTransform> waitForTransform(const std::string &target_frame_id, const std::string &source_frame_id, const double &timeout) = 0;
+
+};
+
+/**
+ the ImplementIFrameSource class provides a default implementation of all the method of IFrameSource interface wich can be reused by the derived classes, 
+ requiring them to only implement some protected method and filling up a container with their frames.
+*/
+class YARP_dev_API yarp::dev::ImplementIFrameSource : public yarp::dev::IFrameSource
+{
+private:
+    std::map<int, std::function<bool(IFrameSource*)>> callbacks;
+protected:
+    
+    class FrameEditor
+    {
+        std::map<std::string, yarp::math::FrameTransform> storage;
+        yarp::math::FrameTransform nullframe;
+
+    public:
+        const auto begin() const { return storage.begin(); }
+        const auto end() const { return storage.end(); }
+        const auto erase(const std::string& k) { return storage.erase(k); }
+    
+        bool insertUpdate(const yarp::math::FrameTransform& frame)
+        {
+            storage[frame.frameId] = frame;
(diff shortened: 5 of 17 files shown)


src/libYARP_dev/include/yarp/dev/IFrameSource.h
```diff
@@ -0,0 +1,200 @@
...
+#include <yarp/math/Quaternion.h>
+#include <yarp/math/FrameTransform.h>
+
+namespace yarp {
+    namespace dev {
+        class IFrameSource;
+    }
+}
+
+struct CustomFrameCompare
```
traversaro: If this is part of the public API, I suggest to move this inside a namespace.

src/libYARP_dev/include/yarp/dev/IFrameSource.h
```diff
@@ -0,0 +1,200 @@
...
+{
+    bool operator()(const yarp::math::FrameTransform& lhs, const yarp::math::FrameTransform& rhs)
```
traversaro: I may be wrong, but I guess that you need to mark as `inline` all the methods defined in the header, otherwise you can get an One Definition Rule (ODR) violation, see https://en.cppreference.com/w/cpp/language/definition . 
This comment applies also to the rest of the class in the headers.

src/libYARP_dev/include/yarp/dev/IFrameSource.h
```diff
@@ -0,0 +1,200 @@
...
+            return true;
+    }
+};
+
+class FrameEditor
+{
+    std::map<std::string, yarp::math::FrameTransform> storage;
+    yarp::math::FrameTransform nullframe;
+public:
+    const auto begin() const { return storage.begin(); }
```
traversaro: At least in the  definition of a class in the public API, I think it is better to avoid using `auto`, otherwise the programmer has no way of understanding the type returned by the method.

src/libYARP_dev/include/yarp/dev/IFrameSource.h
```diff
@@ -0,0 +1,200 @@
...
+        DIRECT,
+        INVERSE,
+        UNDIRECT,
+        DISCONNECTED
+    };
+    ConnectionType getConnectionType(const std::string& target_frame, const std::string& source_frame, std::string* commonAncestor = nullptr);
+
+    robotology::Result<yarp::math::FrameTransform> getChainedTransform(const std::string& target_frame_id, const std::string& source_frame_id);
+
+    virtual bool frameExistsRaw(const std::string &frame_id);
```
traversaro: Why `raw` in the name?

src/libYARP_dev/include/yarp/dev/IFrameSource.h
```diff
@@ -0,0 +1,200 @@
...
+
+    FrameEditor frameContainer;
```
traversaro: I guess it should be possible to expose a `IFrameSource` interface from a class that **does not store** its frame information in a `FrameEditor`, so it could make sense to avoid inserting the `FrameEditor` in the virtual interface, right?

src/libYARP_dev/include/yarp/dev/IFrameSource.h
```diff
@@ -0,0 +1,200 @@
...
+    * @param target_frame_id the name of target reference frame
+    * @param source_frame_id the name of source reference frame
+    * @return true/false
+    */
+    virtual bool frameExists(const std::string &frame_id);
+
+    /**
+    Gets a vector containing all the registered frames.
+    * @param ids the returned vector containing all frame ids
+    * @return true/false
```
traversaro: Documentation seems to be outdated w.r.t. to the methods.

src/libYARP_dev/include/yarp/dev/IFrameSource.h
```diff
@@ -0,0 +1,200 @@
...
+    */
+    virtual std::unordered_set<std::string> getAllFrameIds();
```
traversaro: For this and other methods, the old interface was: 
~~~
bool getAllFrameIds (std::vector< std::string > &ids) override
~~~
the old method permitted to get the the `frameIds` and the similar quantities without any dynamic memory allocation, while the new signature does not permit this. I suggest to also reintroduce the old signature to not constrain any user of the interface to have a dynamic memory allocation when using the interface.

src/devices/ovrheadset/OVRHeadset.h
```diff
@@ -132,58 +144,74 @@ class OVRHeadset : public yarp::dev::DeviceDriver,
...
+    GLFWwindow* window{ nullptr };
     ovrTrackingState ts;
     ovrPoseStatef headpose;
+    ovrPoseStatef predicted_headpose;
     unsigned int guiCount;
-    bool         enableGui;
-
+    bool         enableGui{ true };
+#ifdef YARP_EXPERIMENTAL
+    std::unique_ptr<std::mutex> frameMutex{ new std::mutex };
```
traversaro: Why not simply `std::mutex` instead of `std::unique_ptr<std::mutex>` ?

src/devices/ovrheadset/OVRHeadset.h
```diff
@@ -26,6 +26,7 @@
 #include <yarp/dev/ServiceInterfaces.h>
 #include <yarp/dev/IJoypadController.h>
 #include <yarp/dev/IFrameTransform.h>
+#include <yarp/dev/IFrameSource.h>
```
traversaro: Should this be guarded by `YARP_EXPERIMENTAL`?

traversaro:
Some additional comments: 
* As the `YARP_ENABLE_UNSTABLE` flag changes the public API, downstream users of YARP need a way to know if the `YARP` library that they find with `find_package(YARP ..)` was compiled with the `YARP_ENABLE_UNSTABLE` or not (to understand if they can use the `IFrameSource.h` file or not). This can be done by either adding a new CMake component to the package, or set a CMake variable in `YARPConfig.cmake` (but probably @drdanz was trying to deprecated those)

* As the `<robotology/Result.h>` header is include in public headers, and as the `RobotologyTemplateLibrary` CMake imported target is linked as `PUBLIC` to `YARP_dev`:
~~~
target_link_libraries(YARP_dev PUBLIC RobotologyTemplateLibrary)
~~~  
then in `YARPConfig.cmake` if  `YARP_ENABLE_UNSTABLE`  is set to `ON`, `find_dependency(RobotologyTemplateLibrary REQUIRED)` should be added, otherwise any downstream project consuming the `YARP::YARP_dev` library will actually have a linker error due to the not found `RobotologyTemplateLibrary` library not found. (Off topic: if the `RobotologyTemplateLibrary` target was exported, with a NAMESPACE [see https://github.com/open62541/open62541/issues/2415], at least the error would be during the CMake configuration). 

* Is the file ` src/devices/ovrheadset/calib.png` for what is used? The YARP repo is already quite massive, and if we can avoid inserting another ~120 Kb it can be useful. 

It may be just my ordinary fear of new stuff, but the new `YARP_ENABLE_UNSTABLE` stuff still scares me and I am not fully convinced by it, to be honest. The complexities include stuff like the first thing mentioned in this comment, or the fact that any additional binary option effectively doubles the number of configuration for which we should test YARP. I am not sure that the additional complexity of it is worth. Perhaps the simpler solution is just to try to get the testing and "iterations" of a new interface before a YARP release, and if it is not possible to do so, we explicitly mark them as unstable in the documentation.

CMakeLists.txt
```diff
@@ -60,7 +60,10 @@ message(STATUS "  CXX Compiler: ${CMAKE_CXX_COMPILER_ID} ${CMAKE_CXX_COMPILER_VE
 set_property(GLOBAL PROPERTY USE_FOLDERS 1)
 set_property(GLOBAL PROPERTY AUTOGEN_TARGETS_FOLDER "Autogen Targets")
 set_property(GLOBAL PROPERTY AUTOGEN_SOURCE_GROUP "Generated Files")
-
+option(YARP_ENABLE_UNSTABLE "enable unstable/experimental features")
```
traversaro: YARP CMake options are typically contained in `yarp/cmake/YarpOptions.cmake`

src/libYARP_dev/CMakeLists.txt
```diff
@@ -110,6 +114,10 @@ if (CREATE_LIB_MATH)
                             include/yarp/dev/Map2DLocation.h
                             include/yarp/dev/Map2DArea.h
                             include/yarp/dev/MapGrid2D.h)
+    if(RobotologyTemplateLibrary_FOUND)
```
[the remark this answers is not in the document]
aerydna: the find_package of robotologytemplatelibrary is inside an if(YARP_ENABLE_UNSTABLE). so it should be impossible for it to be found. and that logic is strictly related to RobotologyTemplateLibrary.

src/devices/ovrheadset/OVRHeadset.h
aerydna: sure! i forgot it

src/devices/ovrheadset/OVRHeadset.h
aerydna: because the mutex isn't alway involved. is involved only in the case where the user don't activate the callback (and so there are two threads). otherwise only the device thread is involved. in this case there's no need for the mutex

src/libYARP_dev/include/yarp/dev/IFrameManager.h
```diff
@@ -0,0 +1,60 @@
...
+    * @return true/false
+    */
+    virtual robotology::Result<void> clear() = 0;
+
+    /**
+    Register a transform between two frames.
+    * @param target_frame_id the name of target reference frame
+    * @param source_frame_id the name of source reference frame
+    * @param transform the transformation matrix from source_frame_id to target_frame_id
+    * @return true/false
```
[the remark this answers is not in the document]
traversaro: Sorry, I thought you were referring only to documentation that still needed to be written.

src/devices/ovrheadset/OVRHeadset.h
traversaro: What if the pointer to the device is passed to another device by the yarprobotinterface using the `IMultipleWrapper` method? In that case, any `IFrameSource` method of the class could be called by any arbitrary thread.

src/libYARP_dev/include/yarp/dev/IFrameSource.h
aerydna: and it is right. the return value of the method is strictly dependent on the return value of `std::map::begin()` and should remain so. it is also most likely that the return value will not be directly used by the user but passed to std method or called automatically by the compiler in the range-based for loops.


examples are `std::transform(myframeeditor.begin(), … etc)` or `for (const auto& f : myframeeditor){}`

aerydna: what the programmer have to understand is that begin method is a propagation of std::map::begin method. no other information are needed

src/libYARP_dev/include/yarp/dev/IFrameSource.h
aerydna: the "raw" version of the method are internal only and are not guarded. BTW see the "naming" note in the notes paragraph of the pr

traversaro: Not sure which "naming" note you are referring to. If it is "there will be, for sure, someone who don't like the names of the new flags/interfaces/something and has better alternative. not a problem for me. Just decide the names you like and i will change it." it is not that I do not like it, I just did not understand why `Raw` was added, as typically it is used in "***Raw" interfaces.

src/libYARP_dev/include/yarp/dev/IFrameSource.h
aerydna: mmm.. let me think.. actually the user calls the **already implemented** method of the virtual interface, which comunicate with the full-concrete layer via the intra-class protected interface using `FrameEditor` as a container. it does not make sense to remove the FrameEditor unless someone want to reimplement **all** of the interface method. i presume that to be a case less likely to happen respect to implement only the three protected method so i think it makes sense to not removing the FrameEditor. also because if you remove the frame editor you would force every device to implement all the method.
it's not the derived class that store its frame in the FrameEditor. It is the implemented side of the parent virtual class (IFrameSource) that use it to store the frames. The IFrameSource is not thinked  as pure interface though it can also be used as it was.

src/libYARP_dev/include/yarp/dev/IFrameSource.h
aerydna: yes, see the previous comment or the notes section of the pr

src/devices/ovrheadset/OVRHeadset.h
aerydna: i see your point. i'll think a solution for it.

aerydna:
> Some additional comments:
As the YARP_ENABLE_UNSTABLE flag changes the public API, downstream users of YARP need a way to know if the YARP library that they find with find_package(YARP ..) was compiled with the YARP_ENABLE_UNSTABLE or not (to understand if they can use the IFrameSource.h file or not). This can be done by either adding a new CMake component to the package, or set a CMake variable in YARPConfig.cmake (but probably @drdanz was trying to deprecated those)
As the <robotology/Result.h> header is include in public headers, and as the RobotologyTemplateLibrary CMake imported target is linked as PUBLIC to YARP_dev:
target_link_libraries(YARP_dev PUBLIC RobotologyTemplateLibrary)
then in YARPConfig.cmake if YARP_ENABLE_UNSTABLE is set to ON, find_dependency(RobotologyTemplateLibrary REQUIRED) should be added, otherwise any downstream project consuming the YARP::YARP_dev library will actually have a linker error due to the not found RobotologyTemplateLibrary library not found. (Off topic: if the RobotologyTemplateLibrary target was exported, with a NAMESPACE [see https://github.com/open62541/open62541/issues/2415], at least the error would be during the CMake configuration).

i totally agree.

> Is the file src/devices/ovrheadset/calib.png for what is used? The YARP repo is already quite massive, and if we can avoid inserting another ~120 Kb it can be useful.

it is for the next feature of the ovrdevice so it will end up there anyway. it will be visualized inside the oculus when in regulation mode. if you it bother you i can remove it and reinsert it in the next pr i'm working on. Or we can simply leave it there

aerydna:
> It may be just my ordinary fear of new stuff, but the new YARP_ENABLE_UNSTABLE stuff still scares me and I am not fully convinced by it, to be honest. The complexities include stuff like the first thing mentioned in this comment, or the fact that any additional binary option effectively doubles the number of configuration for which we should test YARP. I am not sure that the additional complexity of it is worth. Perhaps the simpler solution is just to try to get the testing and "iterations" of a new interface before a YARP release, and if it is not possible to do so, we explicitly mark them as unstable in the documentation.

- the instability is not only an instability of yarp itself but is dependent to an instability of an external library.
- i think the unstable part **should not** be tested because it should be possible to break it in anyway
- we can put the new class under a experimental namespace. the problem is only with existent class we want to improve with experimental feature.
- if you (or someone else) are too scared we can simply mark the library stable as it is, remove all the unstable stuff and wait the next release of yarp to release the breaking changes to the ovrdevice. my fear is that it is time consuming and i personally would fork the features i need (devices or interfaces) putting them in external repository in order to start using them immediatly in other projects then subsituting with the yarp version once it come out. this would increase the work overload

- another option is for me to develop things outside yarp (forking existent devices if needed) and when ready include them in yarp. this would increase the work overload though

- this option was imho the best way to contribute to yarp quickly remaining robust (with the modification you suggested).

src/libYARP_dev/include/yarp/dev/IFrameSource.h
traversaro: > the old method had also a dynamic memory allocation as the frame count where subject to change

As long as the capacity of the passed container is always greater than the maximum size, there is no dynamic memory allocation.  

>  we can simply make these container class member and return const references to them

How do you enforce proper mutual exclusion in this case?

src/devices/ovrheadset/OVRHeadset.h
aerydna: ok, a solution could be to enforce the usage of the callback if a user activate it and force in this case a single thread usage of the device (which will call multiple callback, one per user).

aerydna:
@traversaro, maybe the `experimental `namespace is the best way to handle this (it was also suggested by @drdanz ). the thing we can do  for existent classes (like oculus) is to fork them under the experimental namespace. with it, for a user is clear that he is using experimental features, subject to changes. what do you think?

traversaro:
> it is for the next feature of the ovrdevice so it will end up there anyway. it will be visualized inside the oculus when in regulation mode. if you it bother you i can remove it and reinsert it in the next pr i'm working on. Or we can simply leave it there

It is a bit difficult to discuss if it is ok or not to merge it or not if it is not used in the current PR. What happens if for example during the discussion for the new feature it turns out that adding the `.png` is not desirable or necessary? I would personally just add it in the new PR that actually just uses it, but I guess it is up to @drdanz to decide on that. 

> if you (or someone else) are too scared we can simply mark the library stable as it is, remove all the unstable stuff and wait the next release of yarp to release the breaking changes to the ovrdevice. my fear is that it is time consuming and i personally would fork the features i need (devices or interfaces) putting them in external repository in order to start using them immediatly in other projects then subsituting with the yarp version once it come out. this would increase the work overload

Ack. Consider that here there is a trade-off between the additional work necessary for you, and the additional complexity and associated problems (and resulting work) that may affect any YARP's user. 
However, I guess this is again something for which @drdanz's opinion would be determinant.

aerydna:
> Ack. Consider that here there is a trade-off between the additional work necessary for you, and the additional complexity and associated problems (and resulting work) that may affect any YARP's user.
However, I guess this is again something for which @drdanz's opinion would be determinant.

it's not mandatory that is a additional work for **me**. i can work and correct things outside of yarp and then maybe someone else have to reintegrate or integrate them into it. maybe they'll end up to not be integrated at all while being useful things. i understand your point though and i suggest to discuss this f2f

src/libYARP_dev/include/yarp/dev/IFrameSource.h
aerydna: there's no way for the user to know the size currently and for the mutual exclusion we can making it thread-local.

traversaro: > there's no way for the user to know the size currently

In practice on a given closed system you can measure the maximum number of frames, and call `.reserve` with a value greater than that. More practically, what typically happens is that the vector is resized a few time, but as the capacity is not change when the vector if shrinked, so there actually are dynamic memory allocation only when the number of frames increased (not always to be honest, as typically the strategy used in `std::vector` on `resize` to increase its buffer is always to double its capacity if a reallocation is needed, regardless of the actual size requested. 

> perhaps we can making it thread-local

Personal opinion: I do not think that forcing any implementation of the `getAllFrameIds` method to have a thread local `std::vector<std::string>` buffer (or even just the additional complexity of doing it once) just to avoid the signature with the out variable passed as a reference is worth it. 

>  after a quick search i found this article https://web.archive.org/web/20130930101140/http://cpp-next.com/archive/2009/08/want-speed-pass-by-value
didn't read it completely but seems to definitely promote the pass by value practice for best performances and for some caveats it propose different solution from the classical output function argument way. i'll read it with attention asap.

That article  assumes that the in any case the `std::vector<std::string>` needs to be created ex-novo every time by the caller, and it is just arguing that for **non-virtual methods defined in the same compilation unit of caller** the compiler is able to ensure that the `std::vector<std::string>` is created only once. Our case is quite different: we are discussing **virtual  methods** (so the compiler at compile time has **no way** of  knowing which method will be actually called) and we are  discussing how to avoid any kind of memory allocation, i.e. that the `std::vector<std::string>` is  created even only once. 

> A good idea for get rid of this question once for all would be to do some test. i think i will do it.

Are you discussing **speed tests** or **dynamic memory allocation** tests? I  am actually arguing to have an interface that does not force dynamic memory allocation, the increased speed would be just a byproduct from my point of view.

> A thing that we can try is also to move the implementation of these method in the header in order to move the compilation of these function on the user side, allowing for further optimization in case of inner-loops calls.

As this is a virtual call that the users will call by just having a pointer to `IFrameSource`, no matter what, the method cannot be inlined by the compiler, because the compiler has no idea of which method is actually called.

src/libYARP_dev/include/yarp/dev/IFrameSource.h
traversaro: Personal opinion: for consistency with the rest of the YARP interfaces, I would leave the IFrameSource as a pure virtual class, and move the default implementation that rely on the three protected methods to a separate class that inherits from `IFrameSource`, like a `IFrameSourceHelper` or a better name.

src/libYARP_dev/include/yarp/dev/IFrameSource.h
traversaro: To be honest I just realized that on an increase in the number of the frames there always be a dynamic memory allocation if the additional frame id does not fit in the sso buffer. However not having dynamic memory allocation if the frame numbers do not change seems already quite convenient to me.

pattacini:
> The PR is very large and touches multiple aspects, making it difficult to review.

I tend to agree with @randaz81.
Further, I think it'd be worth discussing offline the need for an external library.

src/libYARP_dev/include/yarp/dev/IFrameSource.h
aerydna: > Personal opinion: I do not think that forcing any implementation of the getAllFrameIds method to have a thread local std::vector<std::string> buffer (or even just the additional complexity of doing it once) just to avoid the signature with the out variable passed as a reference is worth it.

i would personally get rid of this kind of signature cause it bothers me a lot. none of the framework i use has this kind of signature. this should mean something. anyway **if we find a way to have the same behaviour** with a signature semantically appropriate why don't use every c++11/14 construct to do it?

> Our case is quite different: we are discussing virtual methods (so the compiler at compile time has no way of knowing which method will be actually called)

this is true.. as it is true that is not mandatory at all for those method to be virtual. i had this doubt also for other reasons. removing the "virtualness" of the method would make the IFrameSource not an interface anymore, but a simple partially virtual public superclass. i don't know, i have to think about it

> Are you discussing speed tests or dynamic memory allocation tests? I am actually arguing to have an interface that does not force dynamic memory allocation, the increased speed would be just a byproduct from my point of view.

ok this was not clear for me, thank you for pointing it out. if we add the external reference to be valorized, the interface should do a check on the capacity of the vector and return fail if inappropriate right? in this case i should also add a method to know the frame count. but what to do with device such as the vicon, where new marker can pop in or out in any moment?

traversaro: >  anyway **if we find a way to have the same behaviour** with a signature semantically appropriate why don't use every c++11/14 construct to do it?

Not sure if I got what you mean here. 

> ok this was not clear for me, thank you for pointing it out. if we add the external reference to be valorized, the interface should do a check on the capacity of the vector and return fail if inappropriate right? in this case i should also add a method to know the frame count. but what to do with device such as the vicon, where new marker can pop in or out in any moment?

Consistently with the rest of YARP, I do not think we need to ensure there are no dynamic memory allocations at all, but I just think we should avoid to have an interface that forces the users to have a quite huge number of dynamic memory allocations.

aerydna:
actually, **from a user perspective** it follows the usual yarp devices architecture. with an interface, a device driver and two network wrapper (server side and client side).

aerydna:
with @lornat75 we agreed to discuss the new library introduction in a yarp council, if possible this very week.

src/devices/ovrheadset/OVRHeadset.h
aerydna: it's easier.. i was thinking a way to make the lock and unlock not mandatory.. if i can't find any way to do it i will always use the mutex

aerydna:
iframesource architecture

classical yarp device architecture

randaz81:
Thank you @aerydna. However, I'm still a bit confused about the frameReceiver class (maybe something is missing in the graph?)
The original implementation of the tfServer/tfClient mechanism in Yarp has the two following goals:
- to communicate with modules belonging to the ROS framework, which all communicate via /tf topic (_[...cit...]as i said earlier no ros interoperation is implemented at the moment[...]_ -> No problem, we can think about this later)
- to distribute the transformations between multiple YARP modules (no YARP-ROS interoperation here)

**In this latter case,** let's assume the standard scenario in which _module 1_ computes the A->B transform, _module 2_ computes the B->C transform. Now _module 3_ needs to retrieve the A->C transform. 

I cannot see in your diagram how _module 3_ can connect to multiple Framesources, given that:
1- we are not using a centralized server (this is the current implementation in YARP, with its pros and cons).
2- we are not using a publishers/subscriber paradigm (this would be the best option in my opinion, but I have some concerns about yarp implementation).
3- we are not using multiple yarp ports to perform the connections. (this is the worst implementation, with the number of connections 'all-to-all' exponentionally growing. PS: please note that this logically equivalent to case 2, only the implemetnation is different).

Maybe this PR does not tackle this aspect yet and deals only with 1-1 scenarios?
Thanks for the clarifications.

barbalberto:
I agree that we need a **f2f** chat, starting from _what is the problem you are trying to solve_ with this PR because it is still not clear to me, hence reviewing the code is quite hard.

From my understanding the point is that you may need a `transformClient` that does not open a port. Although it look a strange requirement in the first place (why shall I hide my tranform from the rest of the world?), I doubt all those changes are really required.

About the "transformClient/Server do not follow the usual device + client/wrapper paradigm" I believe it is perfectly fine because they _**are not actual devices**_ . The `ovrHeadset` is a device, the `transformClient` is not. It is simply an utility class, more like the networkClock is. 

Other two comment I have to say:
1) It is generally better talk to people before doign such a big change. Find at least an agreement on the general idea before coding. imho PR are place to discuss about implementation details, not about the idea behind; that shall be discussed beforehand.

2) Bringing "experimental" stuff together with a device architecture refactoring create a pollutiong layer over a PR already complicated by is own. I believe adopting ideas in the experimental repo is another matter that has to be thoughtfully discussed in a YARP council

aerydna:
> About the "transformClient/Server do not follow the usual device + client/wrapper paradigm" I believe it is perfectly fine because they _**are not actual devices**_ . The `ovrHeadset` is a device, the `transformClient` is not. It is simply an utility class, more like the networkClock is.

exactly, bu they are used as network wrapper by the device like oculus and vicon. they do not have to be **actual device**, but they should be **actual wrappers**, and they are not. and for me they are fine, but we also need actual wrappers for devices that produces frames in the space.

> From my understanding the point is that you may need a `transformClient` that does not open a port. Although it look a strange requirement in the first place (why shall I hide my tranform from the rest of the world?), I doubt all those changes are really required.

actually i need that the devices should not open ports, like almost all the devices in yarp (except for ovrdevice), that doesn' t take care of the network publishing part and should be attached to a network wrapper to do it. ovrdevice, frametransformclient and server are inconsistent with the device architecture of yarp.

aerydna:
> Maybe this PR does not tackle this aspect yet and deals only with 1-1 scenarios?

exactly.

aerydna:
also, for those who say this pr to be too big, it actually contains the usual starter-pack for new kind of devices which include:

- an interface, or a set of interfaces(IFrameTransform)
- a device driver (here it would be ovrdevice)
- a client network wrapper
- a server network wrapper

since i wanted to develop the new interface changing the calssic yarp signature which is considered a bad practice (cplusplus common guidelines F.20) i developped it with a helpers (the Result class) and i think this helper to be putted in a dedicated library, common to Robotology and not only in yarp.

in my opinion all this things are not divisible.
it's not good to include a new library without including anything that use it. ----> in my case an interface use it and i think is not good to introduce an interface without including anything that derives it -----> in my case is the ovrdevice, which removes the old method do "export" outside his data -----> which lead to the development of a new network wrapper.

a criticism that i can accept about all that is the one moved to me personally by @drdanz saying that the commit structure is messy and difficult to review. and i found this to be right. The content of the pr, in my opinion is not splittable

src/libYARP_dev/include/yarp/dev/IFrameManager.h
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (C) 2006-2018 Istituto Italiano di Tecnologia (IIT)
```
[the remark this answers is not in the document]
aerydna: done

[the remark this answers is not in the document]
drdanz: Fixed in master and devel, please rebase and update your branch

aerydna:
i have 
- fixed the pr a lot making possible to for different thread to use both mode (callback and polling) simultaneously.
- did some of the architecture/oversight fix suggested/noticed by @traversaro.
- fixed all the cmake parts following @drdanz and @traversaro suggestion
- updated/added the documentation
- reordered all the commits.
- rebased on devel
waiting for feedbacks.

src/libYARP_dev/CMakeLists.txt
```diff
@@ -160,6 +163,10 @@ if(CREATE_LIB_MATH)
                             src/INavigation2D.cpp
                             src/MapGrid2D.cpp
                             src/Map2DArea.cpp)
+  if(RobotologyTemplateLibrary_FOUND)
+  message("asdasdasd")
```
traversaro: leftover

drdanz:
I still cannot see how this can be discussed as a single PR, we need to discuss this in person

src/libYARP_dev/CMakeLists.txt
aerydna: right

aerydna:
> I still cannot see how this can be discussed as a single PR, we need to discuss this in person
as i said earlier:
> in my opinion all this things are not divisible.
it's not good to include a new library without including anything that use it. ----> in my case an interface use it and i think is not good to introduce an interface without including anything that derives it -----> in my case is the ovrdevice, which removes the old method do "export" outside his data -----> which lead to the development of a new network wrapper.

so i cannot see how this can be splitted i multiple pr. as it is now the commits are also well separated. if you separe this in multiple pr the prs would be strictly dependent and it make no sense in my opinion.

drdanz:
Merged the first commit in devel.
Rebased and squashed the license fix and cleanup in the right places.

src/devices/FrameBroadcaster/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+                                                      YARP_dev
+                                                      YARP_math)
+
+  yarp_install(TARGETS yarp_FrameBroadcaster
+               EXPORT YARP_${YARP_PLUGIN_MASTER}
+               COMPONENT ${YARP_PLUGIN_MASTER}
+               LIBRARY DESTINATION ${YARP_DYNAMIC_PLUGINS_INSTALL_DIR}
+               ARCHIVE DESTINATION ${YARP_STATIC_PLUGINS_INSTALL_DIR})
+
+  yarp_install(FILES FrameBroadcaster.ini
```
drdanz: Please, update both the new `CMakeLists.txt` files, and remove the `.ini` files according to the latest changes (see #1964)

src/devices/ovrheadset/CMakeLists.txt
```diff
@@ -13,6 +13,12 @@ yarp_prepare_plugin(ovrheadset
 if(NOT SKIP_ovrheadset)
   set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
+  if(YARP_EXPERIMENTAL_API)
+    option(OVR_DEVICE_EXERIMENTAL "experimental version of OvrDevice" OFF)
```
drdanz: I'd rather move this option definition together with the `YARP_EXPERIMENTAL_API` option.
For coherence, and for grouping with cmake-gui, this option should start with `YARP_`.
Also you should be using `cmake_dependent_option`, so that the option is automatically hidden and set to false when you disable `YARP_EXPERIMENTAL_API`

cmake/YarpOptions.cmake
```diff
@@ -239,6 +239,15 @@ endif()
 yarp_deprecated_option(YARP_FILTER_API) # Since YARP 2.3.68.1
 yarp_deprecated_option(YARP_CLEAN_API) # Since YARP 2.3.68.1
 
+
+#########################################################################
+# Enable/disable wanted, experimental, and unwanted warnings
+
+option(YARP_EXPERIMENTAL_API "enable unstable/experimental features")
+if(YARP_EXPERIMENTAL_API)
+  add_definitions(-DYARP_EXPERIMENTAL)
```
drdanz: I'd rather keep the option name and the definition synchronized, therefore I'd change this to 
```diff
- add_definitions(-DYARP_EXPERIMENTAL)
+ add_definitions(-DYARP_EXPERIMENTAL_API)
```

src/devices/ovrheadset/CMakeLists.txt
```diff
@@ -13,6 +13,12 @@ yarp_prepare_plugin(ovrheadset
...
+  endif()
+  if(OVR_DEVICE_EXERIMENTAL)
+    add_definitions(-DOVR_EXPERIMENTAL)
```
drdanz: Same as the other comment, I'd use the same name for the option and for the definition

src/devices/ovrheadset/CMakeLists.txt
drdanz: (Also the value of this option should be printed in the output of `cmake`)

src/libYARP_math/include/yarp/math/FrameTransform.h
```diff
@@ -17,119 +17,124 @@
...
 {
-    namespace math
+namespace math
+{
+
+class YARP_math_API FrameTransform
+{
+public:
+
+    std::string parentFrame;
```
randaz81: The use of four data members is terribly confusing and its compatibility with ROS is suspicious
We are now confusing different concepts.
A frame transform defines a transformation from a source to a destination frame.
A frame can have a name and a parent. (NB: In ROS frames have child frames, not parent frames)
see: http://wiki.ros.org/tf/Overview/Data%20Types

aerydna: ok, i forgot to add some comments there. The two new member are supposed to replace the old ones which have to be deprecated (in the new implementation they are not used). and this avoid the confusion that created #1775.

regarding the parent/child i knew the difference with ROS and find very wrong to express the child frameID and not the parent. to explain the reason why just follow the following statement:

- the preferable structure for expressing position and orientation between entities, if we have to 
choose between a Tree and graph is a **Tree** in order to avoid conflict
- in a tree a child can have only one parent and a parent can have multiple child
- the **ID** of somethig is supposed to be unique and rappresent the key of the data. so the frameID should be unique in our structure
- with a frameID unique enforcement (via map structure) we can assure the presence of only one parent per frame thus enforcing the concistence of the tree.

conclusion: the tf in this PR are more suitable to be organized in a proper structure.

at the moment i find the ROS way to rappresent that structure wrong. **maybe (very probable) i am missing something important. so please enlighten me.**

after saying that, if we want to stay consistent with ROS despite it to be wrong(possibly) i have no problem. I only want to confront this design and understand why it is not better.

aerydna: on the other way if you want a consistent structure you have to enforce the unique-ness of the childFrame mamber which is counterintuitive

randaz81: Unfortunately, I cannot test ROS compatibility, because your branch is based on a broken YARP. Please rebase or integrate at least aee548794d6b ee013be596d4c63f.

aerydna: i've rebased it on devel on 18/03

src/devices/FrameBroadcaster/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+               COMPONENT ${YARP_PLUGIN_MASTER}
+               DESTINATION ${YARP_PLUGIN_MANIFESTS_INSTALL_DIR})
+
+  set(YARP_${YARP_PLUGIN_MASTER}_PRIVATE_DEPS ${YARP_${YARP_PLUGIN_MASTER}_PRIVATE_DEPS} PARENT_SCOPE)
+
+  set_property(TARGET yarp_FrameBroadcaster PROPERTY FOLDER "Plugins/Device/Fake")
```
drdanz: I don't think these are "fake" devices, therefore they should be in "Plugins/Device"

src/devices/FrameBroadcaster/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+
+  set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+  yarp_add_plugin(yarp_FrameBroadcaster FrameBroadcaster.cpp
+                                        FrameBroadcaster.h)
+
+   target_link_libraries(yarp_FrameBroadcaster PRIVATE YARP::YARP_OS
+                                                     YARP::YARP_sig
+                                                     YARP::YARP_dev
+                                                     YARP::YARP_math)
```
drdanz: Please fix the alignment here

src/devices/FrameReceiver/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+
+  set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+  yarp_add_plugin(yarp_FrameReceiver FrameReceiver.cpp
+                                        FrameReceiver.h)
+
+  target_link_libraries(yarp_FrameReceiver PRIVATE YARP::YARP_OS
+                                                     YARP::YARP_sig
+                                                     YARP::YARP_dev
+                                                     YARP::YARP_math)
```
drdanz: and here

src/devices/FrameReceiver/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+               LIBRARY DESTINATION ${YARP_DYNAMIC_PLUGINS_INSTALL_DIR}
+               ARCHIVE DESTINATION ${YARP_STATIC_PLUGINS_INSTALL_DIR})
+
+  yarp_install(FILES FrameReceiver.ini
+               COMPONENT ${YARP_PLUGIN_MASTER}
+               DESTINATION ${YARP_PLUGIN_MANIFESTS_INSTALL_DIR})
+
+  set(YARP_${YARP_PLUGIN_MASTER}_PRIVATE_DEPS ${YARP_${YARP_PLUGIN_MASTER}_PRIVATE_DEPS} PARENT_SCOPE)
+
+  set_property(TARGET yarp_FrameReceiver PROPERTY FOLDER "Plugins/Device/Fake")
```
drdanz: Same -> Folder "Plugins/Device"

src/libYARP_dev/CMakeLists.txt
```diff
@@ -109,7 +109,9 @@ if (CREATE_LIB_MATH)
                             include/yarp/dev/INavigation2D.h
                             include/yarp/dev/Map2DLocation.h
                             include/yarp/dev/Map2DArea.h
-                            include/yarp/dev/MapGrid2D.h)
+                            include/yarp/dev/MapGrid2D.h
+                            include/yarp/dev/IFrameSource.h
+                            include/yarp/dev/IFrameSet.h)
```
drdanz: This list should be in alphabetical order

src/libYARP_dev/CMakeLists.txt
```diff
@@ -159,7 +161,8 @@ if(CREATE_LIB_MATH)
                             src/IMap2D.cpp
                             src/INavigation2D.cpp
                             src/MapGrid2D.cpp
-                            src/Map2DArea.cpp)
+                            src/Map2DArea.cpp
+                            src/ImplementIFrameSource.cpp)
```
drdanz: Same here

src/devices/FrameBroadcaster/CMakeLists.txt
drdanz: Please, update both the new `CMakeLists.txt` files, and remove the `.ini` files according to the latest changes (see #1964)

src/devices/FrameReceiver/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+                                                      YARP_dev
+                                                      YARP_math)
+
+  yarp_install(TARGETS yarp_FrameReceiver
+               EXPORT YARP_${YARP_PLUGIN_MASTER}
+               COMPONENT ${YARP_PLUGIN_MASTER}
+               LIBRARY DESTINATION ${YARP_DYNAMIC_PLUGINS_INSTALL_DIR}
+               ARCHIVE DESTINATION ${YARP_STATIC_PLUGINS_INSTALL_DIR})
+
+  yarp_install(FILES FrameReceiver.ini
```
drdanz: Please, update both the new `CMakeLists.txt` files, and remove the `.ini` files according to the latest changes (see #1964)

aerydna:
lot of improvement have been done.. the new architecture can now replace the old in almost every aspect. some of the changes have been discussed f2f with the yarp council (example: removal of the external library) and some other are a (mine) proposal for solving some issues and concern moved by  @randaz81 and @barbalberto in a (still unfinished) f2f meeting.

src/devices/FrameBroadcaster/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+                                                      YARP_sig
+                                                      YARP_dev
+                                                      YARP_math
+                                                      YARP_rosmsg)
+
+  yarp_install(TARGETS yarp_FrameBroadcaster
+               EXPORT YARP_${YARP_PLUGIN_MASTER}
+               COMPONENT ${YARP_PLUGIN_MASTER}
+               LIBRARY DESTINATION ${YARP_DYNAMIC_PLUGINS_INSTALL_DIR}
+               ARCHIVE DESTINATION ${YARP_STATIC_PLUGINS_INSTALL_DIR})
```
drdanz: `YARP_INI DESTINATION ${YARP_PLUGIN_MANIFESTS_INSTALL_DIR}` is still missing here

src/devices/FrameReceiver/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+                                                      YARP_sig
+                                                      YARP_dev
+                                                      YARP_math
+                                                      YARP_rosmsg)
+
+  yarp_install(TARGETS yarp_FrameReceiver
+               EXPORT YARP_${YARP_PLUGIN_MASTER}
+               COMPONENT ${YARP_PLUGIN_MASTER}
+               LIBRARY DESTINATION ${YARP_DYNAMIC_PLUGINS_INSTALL_DIR}
+               ARCHIVE DESTINATION ${YARP_STATIC_PLUGINS_INSTALL_DIR})
```
drdanz: and here

src/devices/FrameBroadcaster/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+                    INCLUDE FrameBroadcaster.h)
+
+if(ENABLE_FrameBroadcaster)
+
+  set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+  yarp_add_plugin(yarp_FrameBroadcaster FrameBroadcaster.cpp
+                                        FrameBroadcaster.h)
+
+   target_link_libraries(yarp_FrameBroadcaster PRIVATE YARP::YARP_OS
```
drdanz: wrong alignment :wink:

src/devices/FrameBroadcaster/FrameBroadcaster.ini
```diff
@@ -0,0 +1,4 @@
+[plugin FrameBroadcaster]
```
drdanz: This file should not exist

src/devices/FrameBroadcaster/CMakeLists.txt
aerydna: fixed

src/devices/FrameBroadcaster/FrameBroadcaster.ini
aerydna: fixed

src/devices/FrameReceiver/CMakeLists.txt
aerydna: fixed

src/devices/FrameBroadcaster/CMakeLists.txt
aerydna: fixed

aerydna:
i've updated the previous tests with the new architecture which stands also as an example for using it. i've made a lot of improvement which makes the system to be used ALSO i a way very similar to the previous architecture (as can be seen from the tests.). the f2f discussion took 2 intensive meetings till now and it is still unfinished. we have to review all the update i made in the past weeks.
@barbalberto @randaz81 @drdanz

aerydna:
the pr has been closed following some technical disagreement

PR #1976, opened 2019-03-15
Use ACE targets + other small CMake improvements

drdanz:


PeterBowman:
@drdanz [this project](https://travis-ci.org/roboticslab-uc3m/tools/jobs/508577171) fails to compile current YARP devel (scroll all the way down) because of some executables (yarp, yarprun and yarpmanager-console, perhaps more) still exploiting old FindACE's variables instead of the new imported targets. These variables resolve to `<var>-NOTFOUND` and thus are reported by CMake in the error log. Note that we set `-DSKIP_ACE=ON` **and** libace-dev is not available in the Travis environment, whereas robotology builds actually install this library ([ref](https://travis-ci.org/robotology/yarp/jobs/508257594#L2610)) before configuring YARP.

PR #1977, opened 2019-03-15
include what you use fix on Types.h

aerydna:
simple as the title said. at the moment the bug is blocking the compilation on visual studio 2019


drdanz:
Rebased and merged, thanks.

PR #1985, opened 2019-03-26
Fix sig::image assignment operators

drdanz:
The move assignment operator causes a memory leak after #1980, moreover the extra move constructor is redundant.

The copy operator does not check for `this`. Same for the `copy()` method.


Nicogene:
> The move assignment operator causes a memory leak after #1980, moreover the extra move constructor is redundant.

Shall we open a issue ?
Unfortunately that line(that does not convince me at all) is necessary to avoid the segfault of camCalib.

drdanz:
There is a bug somewhere else, that line is not the fix.
Feel free to open an issue.

PR #1987, opened 2019-03-29
Timer fix

barbalberto:
This fix an issue in case a 'stop' is called on a Timer that has already expired and stopped


Nicogene:
This bug doesn't affect maste?

aerydna:
`This bug doesn't affect maste?`

it does actually

drdanz:
Rebased onto master, and changed base for merge

drdanz:
Merged, thanks.

PR #1999, opened 2019-04-10
Remove dead code in CBW (old motor interface leftovers)

PeterBowman:
Follows up https://github.com/robotology/yarp/pull/1691. Should be harmful, this patchset does nothing more than:

```c++
if (cond1)
{ /* use new interface methods */ }
else
{
    if (cond1) // *** REMOVED, we never reach this branch ***
    { /* use old interface methods */ }
    else
    { /* report error */ }
}
```

Also, some redundant/misleading/outdated comments have been deleted.


PeterBowman:
Not adding a changelog since this is sheer cleanup work. Perhaps the *master* branch should be addressed?

drdanz:
> Perhaps the master branch should be addressed instead?

Sounds reasonable... Rebased onto master

drdanz:
Merged, thanks.

Nicogene:
Thanks @PeterBowman for addressing this :+1:

PR #2003, opened 2019-04-11
Correctly handle return codes in IAnalogSensor server/client device pair

### LINKED ISSUE #1981: Return codes ignored in IAnalogSensor server/client devices
The [`yarp::dev::AnalogWrapper`](http://www.yarp.it/classyarp_1_1dev_1_1AnalogWrapper.html) & [`yarp::dev::AnalogSensorClient`](http://www.yarp.it/classyarp_1_1dev_1_1AnalogSensorClient.html) server/client pair does not exploit the return codes of the [`yarp::dev::IAnalogSensor`](http://www.yarp.it/classyarp_1_1dev_1_1IAnalogSensor.html) interface, leading to subtle bugs. Affects most methods: `getState`, `calibrateSensor` and `calibrateChannel`, plus overloads.

* On the server side, the reply bottle is never populated ([ref](https://github.com/robotology/yarp/blob/52616edf4673a55d11c94e5f6a2e9ed8810799d0/src/libYARP_dev/src/devices/AnalogWrapper/AnalogWrapper.cpp#L52-L117)); even more, the wrapper doesn't care about the return code (see [enumeration](http://www.yarp.it/classyarp_1_1dev_1_1IAnalogSensor.html#ab421a7572f7740278122dcdf65db7898)) of the callee/subdevice.

* On the client side, the response bottle is indeed evaluated (e.g. [ref](https://github.com/robotology/yarp/blob/52616edf4673a55d11c94e5f6a2e9ed8810799d0/src/libYARP_dev/src/devices/AnalogSensorClient/AnalogSensorClient.cpp#L273)), but then cast into a bool ([ref](https://github.com/robotology/yarp/blob/8af2069b1c8b85fea877770aafbbb6196c5165a8/src/libYARP_dev/include/yarp/dev/ControlBoardHelpers.h#L26-L37)). Note that `response.get(0).isVocab()` will **always** equal `false`, therefore `CHECK_FAIL` returns `true` and so does the interface method.

The subtle nature of this bug arises in the following scenario:

```c++
if (iAnalogSensor->calibrateSensor() != yarp::dev::IAnalogSensor::AS_OK)
{ /* report failure */ }
```

The client implementation of `calibrateSensor()` (among others) casts a `true` boolean value to an integer, which results in `1`, which corresponds to `yarp::dev::IAnalogSensor::AS_ERROR` (see docs).

As a workaround, either avoid such sanity checks, or instantiate the sensor device locally, i.e. avoid the server/client pair.

Tested on Ubuntu 16.04 (Xenial) + YARP v2.3.72.1, affects current YARP devel.

PeterBowman:
Closes #1981. Also, remote calls to `calibrateSensor(const yarp::sig::Vector& value)` are now handled by the server device.

### DIFF
--- a/src/libYARP_dev/src/devices/AnalogSensorClient/AnalogSensorClient.cpp
+++ b/src/libYARP_dev/src/devices/AnalogSensorClient/AnalogSensorClient.cpp
@@ -10,6 +10,7 @@
 #include "AnalogSensorClient.h"
 #include <yarp/os/Log.h>
 #include <yarp/os/LogStream.h>
+#include <yarp/os/Value.h>
 
 /*! \file AnalogSensorClient.cpp implementation of an analog sensor client class*/
 
@@ -17,6 +18,14 @@ using namespace yarp::dev;
 using namespace yarp::os;
 using namespace yarp::sig;
 
+namespace
+{
+    inline int checkResponse(bool ok, yarp::os::Bottle& response)
+    {
+        const yarp::os::Value & v = response.get(0);
+        return ok && v.isInt32() ? v.asInt32() : IAnalogSensor::AS_ERROR;
+    }
+}
 
 inline void InputPortProcessor::resetStat()
 {
@@ -258,7 +267,7 @@ int yarp::dev::AnalogSensorClient::calibrateSensor()
     cmd.addVocab(VOCAB_IANALOG);
     cmd.addVocab(VOCAB_CALIBRATE);
     bool ok = rpcPort.write(cmd, response);
-    return CHECK_FAIL(ok, response);
+    return checkResponse(ok, response);
 }
 
 int yarp::dev::AnalogSensorClient::calibrateSensor(const yarp::sig::Vector& value)
@@ -270,7 +279,7 @@ int yarp::dev::AnalogSensorClient::calibrateSensor(const yarp::sig::Vector& valu
     for (int i = 0; i < this->getChannels(); i++)
          l.addFloat64(value[i]);
     bool ok = rpcPort.write(cmd, response);
-    return CHECK_FAIL(ok, response);
+    return checkResponse(ok, response);
 }
 
 int yarp::dev::AnalogSensorClient::calibrateChannel(int ch)
@@ -280,7 +289,7 @@ int yarp::dev::AnalogSensorClient::calibrateChannel(int ch)
     cmd.addVocab(VOCAB_CALIBRATE_CHANNEL);
     cmd.addInt32(ch);
     bool ok = rpcPort.write(cmd, response);
-    return CHECK_FAIL(ok, response);
+    return checkResponse(ok, response);
 }
 
 int yarp::dev::AnalogSensorClient::calibrateChannel(int ch, double value)
@@ -291,7 +300,7 @@ int yarp::dev::AnalogSensorClient::calibrateChannel(int ch, double value)
     cmd.addInt32(ch);
     cmd.addFloat64(value);
     bool ok = rpcPort.write(cmd, response);
-    return CHECK_FAIL(ok, response);
+    return checkResponse(ok, response);
 }
 
 Stamp yarp::dev::AnalogSensorClient::getLastInputStamp()
--- a/src/libYARP_dev/src/devices/AnalogSensorClient/AnalogSensorClient.cpp
+++ b/src/libYARP_dev/src/devices/AnalogSensorClient/AnalogSensorClient.cpp
@@ -20,7 +20,7 @@ using namespace yarp::sig;
 
 namespace
 {
-    inline int checkResponse(bool ok, yarp::os::Bottle& response)
+    inline int checkResponse(bool ok, const yarp::os::Bottle& response)
     {
         const yarp::os::Value & v = response.get(0);
         return ok && v.isInt32() ? v.asInt32() : IAnalogSensor::AS_ERROR;
--- a/src/libYARP_dev/src/devices/AnalogWrapper/AnalogWrapper.cpp
+++ b/src/libYARP_dev/src/devices/AnalogWrapper/AnalogWrapper.cpp
@@ -55,35 +55,36 @@ bool AnalogServerHandler::_handleIAnalog(yarp::os::Bottle &cmd, yarp::os::Bottle
       return false;
 
     const size_t msgsize=cmd.size();
-    int ret = IAnalogSensor::AS_ERROR;
+    int ret=IAnalogSensor::AS_ERROR;
 
     int code=cmd.get(1).asVocab();
     switch (code)
     {
     case VOCAB_CALIBRATE:
       if (msgsize==2)
-        ret = is->calibrateSensor();
+        ret=is->calibrateSensor();
       else if (msgsize>2)
       {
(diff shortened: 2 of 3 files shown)


PeterBowman:
I'd rather implement `AnalogSensorClient::getState()` ([ref](https://github.com/robotology/yarp/blob/a211f01990c5de5845f37cc12aea2e09e53febd9/src/libYARP_dev/src/devices/AnalogSensorClient/AnalogSensorClient.cpp#L244-L248)), but this probably should be done in devel since *IControlCalibration.h* is missing the state vocab macro ([ref](https://github.com/robotology/yarp/blob/a211f01990c5de5845f37cc12aea2e09e53febd9/src/libYARP_dev/include/yarp/dev/IControlCalibration.h#L142-L151)).

src/libYARP_dev/src/devices/AnalogSensorClient/AnalogSensorClient.cpp
```diff
@@ -10,13 +10,22 @@
...
 
 /*! \file AnalogSensorClient.cpp implementation of an analog sensor client class*/
 
 using namespace yarp::dev;
 using namespace yarp::os;
 using namespace yarp::sig;
 
+namespace
+{
+    inline int checkResponse(bool ok, yarp::os::Bottle& response)
```
Nicogene: The signature should be: `inline int checkResponse(bool ok, const yarp::os::Bottle& response)`

src/libYARP_dev/src/devices/AnalogWrapper/AnalogWrapper.cpp
```diff
@@ -54,38 +54,44 @@ bool AnalogServerHandler::_handleIAnalog(yarp::os::Bottle &cmd, yarp::os::Bottle
...
     switch (code)
     {
     case VOCAB_CALIBRATE:
       if (msgsize==2)
-        is->calibrateSensor();
+        ret = is->calibrateSensor();
       else
       {
-        //read Vector of values and pass to is->calibrate();
+        Vector v(msgsize - 2);
```
Nicogene: What if  `msgsize` is 0 or 1? This initialization seems risky
And I would name that `2`, for code readability.

src/libYARP_dev/src/devices/AnalogWrapper/AnalogWrapper.cpp
```diff
@@ -54,38 +54,44 @@ bool AnalogServerHandler::_handleIAnalog(yarp::os::Bottle &cmd, yarp::os::Bottle
     if (is==nullptr)
       return false;
 
-    int msgsize=cmd.size();
+    const int msgsize=cmd.size();
```
Nicogene: I'd use `size_t` or `auto` here as data type

src/libYARP_dev/src/devices/AnalogSensorClient/AnalogSensorClient.cpp
PeterBowman: Right, I'll change this. I copy-pasted this signature from 

https://github.com/robotology/yarp/blob/a211f01990c5de5845f37cc12aea2e09e53febd9/src/libYARP_dev/include/yarp/dev/ControlBoardHelpers.h#L26-L37

PeterBowman: Done at https://github.com/robotology/yarp/pull/2003/commits/a4909c55c1aedf11575f53fb2bb54f1c41ae7878.

src/libYARP_dev/src/devices/AnalogWrapper/AnalogWrapper.cpp
PeterBowman: Done at https://github.com/robotology/yarp/pull/2003/commits/a4909c55c1aedf11575f53fb2bb54f1c41ae7878.

src/libYARP_dev/src/devices/AnalogWrapper/AnalogWrapper.cpp
PeterBowman: > What if `msgsize` is 0 or 1? This initialization seems risky

Done at https://github.com/robotology/yarp/pull/2003/commits/a4909c55c1aedf11575f53fb2bb54f1c41ae7878.

> And I would name that `2`, for code readability.

I'm sorry, what do you mean?

Nicogene: > I'm sorry, what do you mean? 

I'd use a variable like e.g. `size_t offset = 2;` and then ` Vector v(msgsize - offset);`.
Not sure if it is actually an offset, using a variable make the code more readable, but it is a personal taste

drdanz:
If i'm not wrong, this PR changes the protocol between the AnalogWrapper and AnalogClient, therefore I'd rather merge this in devel...

src/libYARP_dev/src/devices/AnalogWrapper/AnalogWrapper.cpp
PeterBowman: Done at https://github.com/robotology/yarp/pull/2003/commits/776bc3feaac113e11266fb8fddf03ca77e6b40e3.

PeterBowman:
> If i'm not wrong, this PR changes the protocol between the AnalogWrapper and AnalogClient, therefore I'd rather merge this in devel...

Changed base branch to devel.

PeterBowman:
Rebased.

> I'd rather implement `AnalogSensorClient::getState()` ([ref](https://github.com/robotology/yarp/blob/a211f01990c5de5845f37cc12aea2e09e53febd9/src/libYARP_dev/src/devices/AnalogSensorClient/AnalogSensorClient.cpp#L244-L248)), but this probably should be done in devel since _IControlCalibration.h_ is missing the state vocab macro ([ref](https://github.com/robotology/yarp/blob/a211f01990c5de5845f37cc12aea2e09e53febd9/src/libYARP_dev/include/yarp/dev/IControlCalibration.h#L142-L151)).

Just in case: where would this new vocab live, and how should it be named? There are more questions such as: should IAnalogSensor.h implement its own vocab definitions, currently hosted by IControlCalibration.h (which is transitively included by analog sensor devices - never a good thing)?

Nicogene:
> Just in case: where would this new vocab live, and how should it be named?

There is not an "official" place to keep the vocabs, usually the "specialized" vocabs are defined inside the header of the interface that use them, and then more general ones like `VOCAB_GET`, `VOCAB_SET` etc are in `include/yarp/dev/GenericVocabs.h`.

Unfortunately there is not a naming convention for vocabs, my personal preference is:

`VOCAB_<interface_name>_<action>`
e.g. for `IControlMode`: 
`VOCAB_CM_IDLE `

> There are more questions such as: should IAnalogSensor.h implement its own vocab definitions, currently hosted by IControlCalibration.h (which is transitively included by analog sensor devices - never a good thing)?

As I said, each interface should implement its own vocabs, and keep the generic vocabs used by more interfaces in `GenericVocabs.h`.

Nicogene:
I trust @barbalberto review, merging thanks!

PR #2007, opened 2019-04-23
Add initilizer_list constructor in Property

Nicogene:
This PR add the `initializer_list` constructor for the property.

Now it is possible to easily initilialize the Property from the construction.

E.g. : 
```c++
    Property p {{"one", Value(1)},
                {"two", Value(2.0)},
                {"string", Value("foo")}
                };

```

It is an idea that came up with @aerydna 
Please review code


Nicogene:
PS: I would implement something similar also for the Bottle, with a `initilizer_list` of `Value`s

drdanz:
Great idea, thanks

Nicogene:
The test added leads to `segfault` on windows, but only VS2017 :man_shrugging:

Very strange

Issue #2009, opened 2019-04-24
realsense2 device doesn't work with D435i camera(two RGB)


PeterBowman:
I'm trying to start the `realsense2` device with the following configuration (`realsense2.ini`):

```ini
device       RGBDSensorWrapper
subdevice    realsense2
name         /depthCamera

[SETTINGS]
depthResolution (640 480)
rgbResolution   (640 480)
framerate       30
enableEmitter   true

[HW_DESCRIPTION]
clipPlanes (0.2 10.0)
```

The launch command is `yarpdev --from realsense2.ini`, it crashes with a `[ERROR]realsense2Driver: Setting param rgbResolution failed... quitting. `. See [attached output file](https://github.com/robotology/yarp/files/3114075/rs.txt).

I was suspicious about this line:

`[INFO]realsense2Driver: Device consists of 3 sensors `

According to the following code excerpt, the device reports one depth sensor present **and two RGB cameras**, one of which (the second one, which is highly relevant) seems faulty since barely no options are present. The driver device loops over these sensors and assigns depth and color handles accordingly:

https://github.com/robotology/yarp/blob/32ba9f4d4bf7ec8d70fc586a148b491d4e51dab9/src/devices/realsense2/realsense2Driver.cpp#L515-L528

The RGB handle is linked with the last (=second) RGB camera. If I tweak this code so that to a pointer to the first camera is assigned to `m_color_sensor` (e.g. make the loop end at `m_sensors.size() - 1`), the device initializes correctly.

**Configuration:**
- Intel RealSense D435i
- OS: Ubuntu 16.04 (Xenial)
- yarp version: 3.1.1
- compiler: GCC 7

PeterBowman:
BTW I noticed that, regardless of the depth resolution set in the .ini file, the resolution of the depth frames always matches the chosen RGB resolution.

Nicogene:
Hi @PeterBowman 

First of all I'd suggest you to change the name of the `ini` file, it is exactly the same name of the `ini` file used by `yarpdev` to load the plugin, it should be ok, but I would not risk to make them clash, rename it in something like `realsense2Conf.ini`



>I was suspicious about this line:
[INFO]realsense2Driver: Device consists of 3 sensors 

Here in the lab we have only D435 and D415, we don't have the D435i to test it. With the D435 the output is:

```
 Name                 : Intel RealSense D435
  Serial Number        : 829212070409
  Firmware Version     : 05.10.06.00
[...]

PR #2010, opened 2019-04-29
Correctly detect color sensor on D435i

PeterBowman:
Fixes https://github.com/robotology/yarp/issues/2009, please test on other RealSense models (@Nicogene).

Known `rs2::sensor` subclasses include, at time of writing: `rs2::depth_sensor` (checked against in current code), `rs2::pose_sensor`, `rs2::wheel_odometer` and `rs2::roi_sensor` (see [*rs_sensor.hpp*](https://github.com/IntelRealSense/librealsense/blob/9bdbfd98440b6516f8c864fbdfd6a9778b2c5715/include/librealsense2/hpp/rs_sensor.hpp)). I noticed that the color sensor in D435i is perceived by the SDK as a `rs2::roi_sensor` (`sensor.is<rs2::roi_sensor>() == true`). Since this might not be portable to other models, the solution described in https://github.com/IntelRealSense/librealsense/issues/1163#issuecomment-364733894 has been implemented here.

### DIFF
--- a/src/devices/realsense2/realsense2Driver.cpp
+++ b/src/devices/realsense2/realsense2Driver.cpp
@@ -69,8 +69,10 @@ static void print_supported_options(const rs2::sensor& sensor)
 
     if (sensor.is<rs2::depth_sensor>())
         yInfo() << "Depth sensor supports the following options:\n";
-    else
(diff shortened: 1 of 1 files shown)


Nicogene:
Merged, thanks!

PR #2015, opened 2019-05-07
Add const counterpart of CanBuffer::getPointer()

PeterBowman:
Users were forced to `const_cast<>` a `const yarp::dev::CanBuffer` instance in order to call `getPointer()`.


PeterBowman:
Perhaps this change could be dangerous because of the non-const access to the `yarp::dev::CanMessage **data` member. I leave this open for discussion.

drdanz:
This is dangerous indeed. I suggest to add a const method that returns a const CanMessage instead

PeterBowman:
> I suggest to add a const method that returns a const CanMessage instead

The new signature is `const CanMessage *const *getPointer() const`.

drdanz:
Merged, thanks.

PR #2031, opened 2019-05-29
Import YCM 0.10.3 and Catch 2.8.0

drdanz:


PeterBowman:
@drdanz:

* please see https://github.com/robotology/ycm/issues/256
* I can't compile bindings on YARP devel, CMake is unable to find the `swig_add_library()` command; YARP is not instructed to even look into that faulty UseSWIG module:

https://github.com/robotology/yarp/blob/e2c339713e554d487f3c99616de817ddd0022269/extern/ycm/CMakeLists.txt#L15-L17

Issue #1982, opened 2019-03-20
Method yarp::dev::IPositionControl::setRefSpeeds produce zero reference speeds


xEnVrE:
**Describe the bug**                                                                                                                                                                           

A call to the method  `yarp::dev::IPositionControl::setRefSpeeds(const int n_joint, const int* joints, const double* spds)` produce **zero** reference speeds when the actual trajectory is executed after calling `yarp::dev::IPositionControl::positionMove()`.                                                                                                                              
                                                                                                                                                                                               
**To Reproduce**                                                                                                                                                                               

The following code produce the described behavior. The behavior was discovered using the robot `iCubGenova01` while trying to move the right hand finger joints.
                                                                                                                                                                                               
`imod` is of type ` yarp::dev::IControlMode*`                                                                                                                                                  
                                                                                                                                                                                               
`ipos` is of type `yarp::dev::IPositionControl*`                                                                                                                                               
                                                                                                                                                                                               
`joints` is of type `yarp::sig::VectorOf<int>` and contains the indexes of the joints                                                                                                          
                                                                                                                                                                                               
`ref_speeds` is of type `yarp::sig::Vector` and contains the desired reference speeds                                                                                                          
                                                                                                                                                                                               
`des_joints` is of type `yarp::sig::Vector` and contains the desired joints positions                                                                                                          
```c++                                                                                                                                                                                            
for (std::size_t i = 0; i < joints.size(); i++)
    imod->setControlMode(joints[i], VOCAB_CM_POSITION);

ipos->setRefSpeeds(joints.size(), joints.getFirst(), ref_speeds.data()); 

ipos->positionMove(joints.size(), joints.getFirst(), des_joints.data());
```                                                                                                                                                                                            
                                                                                                                                                                                               
The wrong behavior is clearly visible by looking at the `yarpmotorgui` while executing the code. The reference velocity sliders of all the involved joints go to zero and the label change color from green to orange.                                                                                                                                                                       
                                                                                                                                                                                               
**Please note** that the single joint version `yarp::dev::IPositionControl::setRefSpeed`, instead, **is working properly.**                                                                        
                                                                                                                                                                                               
**Configuration:**

I report the current configuration used on the `iCubGenova01` laptop:                                                                                                                                                       
- OS: Ubuntu 16.04                                                                                                                                                                            
- yarp version: v. 3.1.0, [link to commit](https://github.com/robotology/yarp/tree/a38cc505cd3d8e370a141b3313da59adc18db004)
[...]

PR #2033, opened 2019-05-31
Fix issue #1982

barbalberto:
This fix has to be merged in devel because the bug is present only in devel and not in master.
This fixes also another error in `getTemperatures` function which was not doing the joint remapping.

Please @randaz81 can you double check?


drdanz:
Merged, thanks.

PR #2042, opened 2019-06-05
`yarprobotinterface` will not run anymore deprecated DTD format v1.0

randaz81:
Parser for yarprobotinterface configuration files v1.0 is now deprecated. Only v3.0 is supported.


drdanz:
Merged, thanks.

PR #2051, opened 2019-06-20
Fix portaudio examples

drdanz:
@randaz81 please review.


barbalberto:
Actually I think it'll be better if examples will be updated to use the new devices `portaudioRecorder` and `portaudioPlayer` instead of restoring deprecated one.
@randaz81 what do you think?

drdanz:
It's probably a good idea, but I don't know what should be upgraded. Also @randaz81 told me that the `audioPlayerWrapper` and `audioRecorderWrapper` should be used instead...

PR #2053, opened 2019-06-25
Improved the yarp matrix function toString, which was previously calling the c++ string method substr, which caused this function to slow down for large matrices.

kothiga:
Improved the yarp matrix function toString, which was previously calling the c++ string method substr, which caused this function to slow down for large matrices.

Originally submitted as issue #2043.

Made small changes from the previous pull request to properly pass all unit tests.


drdanz:
Merged, thanks.

PR #2055, opened 2019-07-07
yarpdatadumper to save PNG images

pattacini:
This PR aims to extend `yarpdatadumper` so that:
- Images can be saved also in the PNG format; for that, just use the option `--type image_png`
- The format of the time stamps is provided from within the `info.log` file as in the following example:
    ```
    Type: Bottle;
    Stamp: [rx; | tx; | tx+rx;]
    [local-timestamp] /yarp-port-name [connected]
    [local-timestamp] /yarp-port-name [disconnected]
    ```

Preliminary tests show that the low-level YARP services devoted to deal with storing the PNG images are correctly called, even though I'm getting some PNG header errors on my machine. This behavior might be linked to my installation though, hence further verifications are required.

@vtikha I've modified the `yarpdataplayer` to let it read the new format of `info.log`. Please, have a look at it. Moreover, the new information on which stamps are logged could be employed later on to get rid of the command line option `withExtraTimeCol`.

@drdanz which is the release file I'm supposed to edit to keep track of these additions?

### DIFF
--- a/src/yarpdatadumper/main.cpp
+++ b/src/yarpdatadumper/main.cpp
@@ -15,7 +15,6 @@
 #include <yarp/os/RFModule.h>
 #include <yarp/os/Stamp.h>
 #include <yarp/sig/all.h>
-#include <yarp/cv/Cv.h>
 
 #include <iostream>
 #include <iomanip>
@@ -27,13 +26,17 @@
 
 #ifdef ADD_VIDEO
     #include <opencv2/opencv.hpp>
+    #include <yarp/cv/Cv.h>
 #endif
 
 
 using namespace std;
 using namespace yarp::os;
 using namespace yarp::sig;
-using namespace yarp::cv;
+
+#ifdef ADD_VIDEO
+    using namespace yarp::cv;
+#endif
 
 /**************************************************************************/
 enum class DumpType { bottle, image };
@@ -138,6 +141,7 @@ public:
         return (fName.str()+" ["+Vocab::decode(code)+"]");
     }
 
+#ifdef ADD_VIDEO
     const cv::Mat &getImage()
     {
         int code=p->getPixelCode();
@@ -155,6 +159,7 @@ public:
         }
         return img;
     }
+#endif
 };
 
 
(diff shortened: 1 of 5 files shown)


src/yarpdatadumper/main.cpp
```diff
@@ -130,7 +138,23 @@ class DumpImage : public DumpObj
         return (fName.str()+" ["+Vocab::decode(code)+"]");
     }
 
-    void *getPtr() override { return p->getRawImage(); }
+    const cv::Mat &getImage()
```
Nicogene: If I understood correctly `yarpdatadumper` should be usable also without `OpenCV`, in this sense this function should be included in `#ifdef ADD_VIDEO #endif` guards.

pattacini: You're absolutely right! Thanks for pointing it out 👍

pattacini:
Talked f2f to @drdanz who suggested using a `devel.md` for publishing info on these additions.

src/yarpdatadumper/main.cpp
```diff
@@ -88,6 +93,7 @@ class DumpImage : public DumpObj
 {
 private:
     Image *p;
+    cv::Mat img;
```
Nicogene: Probably also this declaration should stay inside `#ifdef ADD_VIDEO #endif`

pattacini:
@vvasco tested the PR on her Linux machine and the low-level YARP service responsible for storing the PNG files gave the same error I got earlier on my Windows:
```
[ERROR][write_png_file] Error during writing header
libpng warning: Invalid bit depth in IHDR
libpng error: Invalid IHDR data
```
Anything weird with:
https://github.com/robotology/yarp/blob/3c157dc43783506b71fb03fe65ad9b201886aaaa/src/libYARP_sig/src/ImageFile.cpp#L460
❓

pattacini:
Tests relative to the presence of stamps info were successful
Thanks @vvasco 🎖

pattacini:
`yarpdataplayer` is now capable of reading new as well as old `info.log` formats.

PR #2061, opened 2019-07-15
Fix mjpeg while using BGR images

### LINKED ISSUE #2060: Mjpeg carrier doesn't work with BGR images
**Describe the bug**

The `mjpeg` carrier does not perform the correct color conversion when connecting ports with different pixel formats.

**To Reproduce**

Connect a `BufferedPort<ImageOf <PixelBgr> >` to a `BufferedPort<ImageOf <PixelRgb> >` via the `mjpeg` carrier and send an image: the result is a received image with red and blue channels inverted.

**Expected behavior**

The correct color channels conversion should be performed to adapt to the pixel formats of the ports.

**Screenshots**

On the left a `BufferedPort<ImageOf<PixelBgr> >` sending an image (input image is red) to yarpview via `tcp` carrier, on the right the same port sending the same image via `mjpeg` carrier (should not be blue).

image

**Configuration**
 - OS: Ubuntu 18
 - yarp version: 3.2.0+20190710.3+git291925cd2
-  compiler: g++

Nicogene:
This PR fixes #2060.

`MONO`, `RGB` and `BGR` have been successfully tested

### DIFF
--- a/doc/release/master.md
+++ b/doc/release/master.md
@@ -1,0 +1,37 @@
+Important Changes
+-----------------
+
+New Features
+------------
+
+### Build System
+
+### Libraries
+
+#### `YARP_OS`
+
+#### `YARP_sig`
+
+#### `YARP_dev`
+
+#### `YARP_cv`
+
+
+### Devices
+
+### Tools
+
+#### `yarpdatadumper`
+
+### GUIs
+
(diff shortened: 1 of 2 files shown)


drdanz:
Merged, thanks.

Issue #2052, opened 2019-06-21
yarpdataplayer speed not consistent with real data


vvasco:
**Describe the bug**
`yarpdataplayer` seems to reproduce data faster than the real data.
This seems related to #2049.

**To Reproduce**
- Record a dataset using `yarpdatadumper --rxTime --txTime`;
- Replay the dataset through `yarpdataplayer --withExtraTimeCol 2`;
- Record the same dataset from `yarpdataplayer` using `yarpdatadumper --rxTime --txTime`;
- Compare the receiver times.

**Expected behavior**
I expect `yarpdataplayer` to use the recorded times for playing the data, while it seems to reproduce data faster.

**Screenshots**
I recorded a dataset with the options `--rxTime --txTime` and re-dumped the data from the dataplayer with the same options. When plotting the receiver times of both datasets, it seems that `yarpdataplayer` (red line) reproduces the data faster than the real data (blue line), while I would expect times to be equal:

![delay-player](https://user-images.githubusercontent.com/9716288/59907491-7f220d80-940b-11e9-940e-41dd740b4c57.jpg)

The delay accumulates and at the end of my dataset (which is ~`20mins` long), it reaches ~`158s`. 
`yarpdataplayer` seems to apply a warping factor of ~`0.8` to time, which seems consistent across different datasets. 

**Configuration (please complete the following information):**
 - OS: `Ubuntu 18.04.2 LTS`
 - yarp version: v3.1.0 commit `96123316f36edd00db68b46b2a6240d1d5ab11ba`

**Additional context**
I spotted the problem because I was trying to synchronize a command (black dashed line in the plot) using the recorded time, and it was given  ~`20s` later than expected.

pattacini:
Hi @vvasco 

Could you attach here in a zip containing the minimum dataset with which we can reproduce the problem?

vvasco:
Hi @pattacini 

Unfortunately I can't because the dataset contains sensitive data.
But I can provide @vtikha with it.

vtikha:
@vvasco & @lrapetti First of all I apologise for the delay but was not able to work on this any sooner. I managed to replicate the problem with your dataset and instructions. There was clearly an issue with the virtual virtual time of the `yarpdataplayer` and I have been working on a solution. 

**Behaviour**

I have used @vvasco 20 minutes dataset to:

- Replay the dataset through yarpdataplayer --withExtraTimeCol 2; (also --withExtraTimeCol 1)

[...]

Issue #2049, opened 2019-06-10
yarpdataplayer timestamps not consistent


lrapetti:
**Describe the bug**
The same dataset dumped with `yarpdatadumper` played multiple time seems to be played with different timing, and in particular this time seems to be related with the computational work that the machine is doing. 

**Expected behavior**
I would expect to be able to replay offline the same dataset with the real/same timestamp.

**Screenshots**
In particular I'm running [this dataset](https://drive.google.com/open?id=1Kvy-zZQpdnBeixCQFzh8lFQ2oTDvCDnV) , and the data are meant to be read from a consumer device ([human state provider](https://github.com/robotology/human-dynamics-estimation/tree/master/devices/HumanStateProvider)).
As you can see from the data, they were collected with both tx and rx time, so I have tried to play the data using one or the other, but the same behaviour seems to be present.
From the following screenshots you can see the different duration of the same dataset by checking the `Time Taken` coloumn in the `yarpdataplayer` interface (the same duration was double-checked with a clock):
- `yarpdataplayer --withExtraTimeCol 1`
  - just playing the data
    ![Screen Shot 2019-06-10 at 16 54 34](https://user-images.githubusercontent.com/35487806/59204713-7cb8eb80-8ba1-11e9-8536-7996fc1b4738.png)
  - data + consumer device
    ![Screen Shot 2019-06-10 at 16 57 42](https://user-images.githubusercontent.com/35487806/59204654-5eeb8680-8ba1-11e9-887b-d11e953b8016.png)
- `yarpdataplayer --withExtraTimeCol 2`
  - just playing the data
    ![Screen Shot 2019-06-10 at 16 58 58](https://user-images.githubusercontent.com/35487806/59204737-8b070780-8ba1-11e9-8bae-b108e0f20c40.png)
  - data + consumer device
    ![Screen Shot 2019-06-10 at 17 00 18](https://user-images.githubusercontent.com/35487806/59204742-922e1580-8ba1-11e9-986c-2fc187ae8763.png).

**Configuration:**
 - OS: `MacOS 10.13.6`
 - yarp version: `3.1.1+10-20190514.4+git96bad8480`

@kouroshD

DanielePucci:
@drdanz @Nicogene may probably help us with this?

Nicogene:
It seems a clock drifting, I will give a look to this. 👍 
But maybe @vtikha , that is the author of `yarpdataplayer`, has some hints about what should be the problem here.

[...]

PR #2074, opened 2019-08-05
adding timing changes to yarpdataplayer

vtikha:
This PR fixes the issues where timestamps were not consistent as reported in #2049 #2052.

### DIFF
--- a/src/yarpdataplayer/include/worker.h
+++ b/src/yarpdataplayer/include/worker.h
@@ -134,21 +134,21 @@ public:
     bool                    stepfromCmd;
 
     using Moment = std::chrono::time_point<std::chrono::high_resolution_clock>;
-    
-    void Initialize();
-    
-    void Tick();
-    
+
+    void initialize();
+
+    void tick();
+
     float diff_seconds() const { return dtSeconds; }
     float framesPerSecond() const { return fps; }
-    
+
     QMainWindow* wnd;
 
     /**
      * Master thread class
      */
     MasterThread(Utilities *utilities, int numPart, QMainWindow *gui, QObject *parent = NULL);
-    
+
     bool threadInit() override;
     /**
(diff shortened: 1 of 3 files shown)


src/yarpdataplayer/include/worker.h
```diff
@@ -130,15 +130,23 @@ class MasterThread : public QObject,  public yarp::os::PeriodicThread
     int                     numThreads;
     double                  timePassed, initTime, virtualTime;
     bool                    stepfromCmd;
+    
+    typedef std::chrono::time_point<std::chrono::steady_clock> Moment;
+    
+    void Initialize();
+    
+    void Tick();
```
Nicogene: I personally don't like function names upper case, but it is a personal taste, feel free to ignore it 😉 
Moreover it is not a public interface, then not a big deal.

src/yarpdataplayer/include/worker.h
```diff
@@ -130,15 +130,23 @@ class MasterThread : public QObject,  public yarp::os::PeriodicThread
     int                     numThreads;
     double                  timePassed, initTime, virtualTime;
     bool                    stepfromCmd;
+    
+    typedef std::chrono::time_point<std::chrono::steady_clock> Moment;
```
Nicogene: I'd prefer `using Moment = std::chrono::time_point<std::chrono::steady_clock>;`

src/yarpdataplayer/include/worker.h
vtikha: Correct, changed them, thanks

src/yarpdataplayer/include/worker.h
vtikha: changed thank you

Nicogene:
Merged, thanks!

PR #2076, opened 2019-08-08
Enable colored logger output on Windows 10

PeterBowman:
Colored output by `yarp::os::Log`'s utilities is achieved via setting the `YARP_COLORED_OUTPUT` environment variable on UNIX systems. This patch enables said feature for the Windows 10 console.

Tested on 17134 (1803). Should work since 10586 (1511), although code authors were forced to enable colors programatically from 14393 (1607) onwards according to https://github.com/symfony/symfony/issues/19520#issuecomment-238815318.

Adapted from RL/UC3M code, see https://github.com/roboticslab-uc3m/color-debug/issues/5 for full rationale.

Sample test code:

```c++
cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
project(yarp-logger-test LANGUAGES CXX)
find_package(YARP 3.1 REQUIRED COMPONENTS OS)
file(WRITE ${CMAKE_BINARY_DIR}/main.cpp
"
	#include <yarp/os/LogStream.h>
	int main()
	{
		yTrace() << \"yTrace()\";
		yInfo() << \"yInfo()\";
		yDebug() << \"yDebug()\";
		yWarning() << \"yWarning()\";
		yError() << \"yError()\";
		return 0;
	}
")
add_executable(${PROJECT_NAME} ${CMAKE_BINARY_DIR}/main.cpp)
target_link_libraries(${PROJECT_NAME} PRIVATE YARP::YARP_OS)
```

Output:

logger

### DIFF
--- a/src/libYARP_os/src/yarp/os/impl/windows/WinConsole.c
+++ b/src/libYARP_os/src/yarp/os/impl/windows/WinConsole.c
@@ -1,0 +1,21 @@
+#include <stdbool.h> // C99+
+#include <windows.h>
+
+void yarp_logger_enable_vt_colors(void)
+{
+    static bool enabled = false;
+
+    if (!enabled)
+    {
+        DWORD handleMode = 0;
+        HANDLE hStdout = GetStdHandle(STD_OUTPUT_HANDLE);
+
+        if (hStdout != INVALID_HANDLE_VALUE && GetConsoleMode(hStdout, &handleMode))
+        {
+            handleMode |= ENABLE_VIRTUAL_TERMINAL_PROCESSING;
+            SetConsoleMode(hStdout, handleMode);
+        }
+
+        enabled = true;
+    }
+}
--- a/cmake/YarpSystemCheck.cmake
+++ b/cmake/YarpSystemCheck.cmake
@@ -187,6 +187,13 @@ if(WIN32)
     add_definitions(-D_WINSOCK_DEPRECATED_NO_WARNINGS)
     add_definitions(-D_WINSOCK_DEPRECATED_NO_WARNINGS)
 
+    set(YARP_HAS_WIN_VT_SUPPORT 0)
+
+    # Check whether colored output is available on Windows console.
+    if(NOT CMAKE_SYSTEM_VERSION VERSION_LESS 10.0.10586)
+      set(YARP_HAS_WIN_VT_SUPPORT 1)
+    endif()
+
     # Traditionally, we add "d" postfix to debug libraries
     set(CMAKE_DEBUG_POSTFIX "d")
   endif()
--- a/src/libYARP_os/src/CMakeLists.txt
+++ b/src/libYARP_os/src/CMakeLists.txt
@@ -359,6 +359,8 @@ set(YARP_os_IMPL_POSIX_SRCS yarp/os/impl/posix/TcpAcceptor.cpp
 set(YARP_os_IMPL_MACOS_HDRS yarp/os/impl/macos/MacOSAPI.h)
 set(YARP_os_IMPL_MACOS_SRCS yarp/os/impl/macos/MacOSAPI.mm)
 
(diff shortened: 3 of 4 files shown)


drdanz:
Thanks for the patch, and sorry for the delay!
I have a couple of comments:
* The patch enables colors at build time, but what happens if you build yarp on one system that supports it, and run it on an older system (e.g. binary packages)? Perhaps the check should be at runtime?
* Since the `yarp_logger_enable_vt_colors` function is called only in `Log.cpp` I'd rather keep everything in the same compilation unit (same .cpp file), in an `extern "C"` block (if that's actually required), and possibly make it `inline`. Having it in a separate compilation unit would add an extra function call for every log line which is definitely not optimal for a logging function.

PeterBowman:
> The patch enables colors at build time, but what happens if you build yarp on one system that supports it, and run it on an older system (e.g. binary packages)? Perhaps the check should be at runtime?

Good catch, I need to investigate this.

> Since the `yarp_logger_enable_vt_colors` function is called only in `Log.cpp` I'd rather keep everything in the same compilation unit (same .cpp file), in an `extern "C"` block (if that's actually required), and possibly make it `inline`. Having it in a separate compilation unit would add an extra function call for every log line which is definitely not optimal for a logging function.

Done, but please note [`inline`](https://en.cppreference.com/w/cpp/language/inline) is only a hint to the compiler, which is free to ignore it (also, I think it can choose to inline functions not marked as such).

drdanz:
> I think it can choose to inline functions not marked as such

That's right, but the compiler cannot choose to inline functions that are defined in a separate compilation unit (because they are simply not defined anywhere during the compilation phase, but only available as a function in the linking phase), therefore this will always lead to a function call... I saw some talk about tools to do this kind of optimization, but it looked quite complicated stuff...

drdanz:
The patch looks good now, let me know if you find out something about the first point.
Also the CI was broken, but should be fixed now, please rebase your branch to restart the builds

drdanz:
> > The patch enables colors at build time, but what happens if you build yarp on one system that supports it, and run it on an older system (e.g. binary packages)? Perhaps the check should be at runtime?
> 
> Good catch, I need to investigate this.

@PeterBowman Did you have any chance to investigate this? I'd like to merge this PR, but I'm a little bit worried about this point...

PeterBowman:
> @PeterBowman Did you have any chance to investigate this? I'd like to merge this PR, but I'm a little bit worried about this point...

Sorry, I'll give it a look this weekend and confirm whether I can definitely find a solution.

drdanz:
If you can't confirm this, perhaps we can add some CMake option `YARP_EXPERIMENTAL_LOGGER_COLOR` on windows, and leave it disabled by default...

src/libYARP_os/src/yarp/os/impl/windows/WinConsole.c
```diff
@@ -0,0 +1,21 @@
...
+    static bool enabled = false;
+
+    if (!enabled)
+    {
+        DWORD handleMode = 0;
+        HANDLE hStdout = GetStdHandle(STD_OUTPUT_HANDLE);
+
+        if (hStdout != INVALID_HANDLE_VALUE && GetConsoleMode(hStdout, &handleMode))
+        {
+            handleMode |= ENABLE_VIRTUAL_TERMINAL_PROCESSING;
```
PeterBowman: This file successfully compiles on Windows7+VS2005 when replacing `ENABLE_VIRTUAL_TERMINAL_PROCESSING` with `0x0004` ([docs](https://docs.microsoft.com/en-us/windows/console/setconsolemode)). Then, `SetConsoleMode` simply returns `0` on failure.

drdanz: Cool, in this case I think we can go on and merge this!

drdanz:
Merged, thanks.

PR #2096, opened 2019-09-23
dev: Move files in proper .cpp/.h

drdanz:
Move a few interfaces in the `.h` file where they belong.
These interfaces need to be included at least once in the library (because they are declared `YARP_dev_API`, therefore the relative `.cpp` header is also included.

Moreover a few of these files are useless, therefore:

### dev

* The `<yarp/dev/GenericSensorInterfaces.h>` include file is deprecated in favour of
  `<yarp/dev/IGenericSensor.h>`.
* The `<yarp/dev/SerialInterfaces.h>` include file is deprecated in favour of
  `<yarp/dev/ISerialDevice.h>`.
* The `<yarp/dev/Wrapper.h>` include file is deprecated in favour of
  `<yarp/dev/IWrapper.h>` and `<yarp/dev/IMultipleWrapper.h>`.

### DIFF
--- a/doc/release/devel/GenericSensorInterfaces.md
+++ b/doc/release/devel/GenericSensorInterfaces.md
@@ -1,0 +1,7 @@
+GenericSensorInterfaces {#devel}
+-----------------------
+
+### dev
+
+* The `<yarp/dev/GenericSensorInterfaces.h>` include file is deprecated in favour of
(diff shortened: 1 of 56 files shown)


drdanz:
Travis is not working today. Merging anyway

PR #2109, opened 2019-10-15
Add the possibility to align the frames of the realsense camera also to the depth frame

Nicogene:
This PR add the possibility to align the frames of the realsense camera to the depth frame.
Before it was only possible to align the depth image over the rgb frame.

For doing so I add a new parameter called `alignmentFrame` and deprecated the old `needAlignment`.

The default behaviour has been maintaned, if not specified `alignmentFrame` is `RGB`.
The allowed values are `RGB`, `Depth` and `None`.

To be tested.

Please review code.

### DIFF
--- a/doc/release/devel/alignRGBD.md
+++ b/doc/release/devel/alignRGBD.md
@@ -1,0 +1,10 @@
+alignRGBD {#devel}
+---------
+
+### devices
+
+#### realsense2
+
+* Deprecated the `needAlignment` parameter in favour of `alignmentFrame`.
+  The default behaviour has been maintaned, if not specified `alignmentFrame`
+  is `RGB`. The allowed values are `RGB`, `Depth` and `None`.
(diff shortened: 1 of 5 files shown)


Nicogene:
> Isn't the current behavior None when the option is unspecified?

Before by default the two frames were aligned to the RGB frames. We decided that because it is useful having the frames aligned in cases such as extract the image of the closest blob or similar applications.

Nicogene:
I tested successfully this new feature and from my side it can be merged.

Unfortunately what the realsense does when aligning the rgb frame over the depth is not what @pattacini needed.

Here is the result of this alignment:

## `D415`
image


## `D435`
image

The difference of behavior between the two models is due to the differences between the fov of the two sensors(rbg and IR).

|   | `D415`  | `D435`   |  
|:-:|:-:|:-:|
| RGB-FOV  | 55.4591217041016 43.0581703186035  |  54.7643051147461 42.4490661621094 |  
| Depth-FOV | 54.2412223815918 42.0246467590332 |  79.3044281005859 63.7311859130859 |

Nicogene:
The fail in Appveyor was unrelated

Issue #2107, opened 2019-10-10
REGEX REPLACE sintax error causing invalid sequence 


CarlottaSartore:
**Issue Description**

In the file _cmake/YarpIDL.cmake_ line 171 (**devel branch**) 

https://github.com/robotology/yarp/blob/a656496b49d3c4afafef10e8a647abfb39a1e686/cmake/YarpIDL.cmake#L170-L172

causes the installation of a _YarpIDL.cmake_ file  that arises the following error when trying to compile other repositories that depend on Yarp package 

`when parsing string ^.+/([^\/]+)$ Invalid escape sequence \/ `

Solved by modifying the cmake file, hence substituting the aforementioned line with; 

```
      if(NOT _cpp_placement_ns)
        string(REGEX REPLACE "^.+/([^/]+)$" "\\1" _dest_file "${_dest_file}")
      endif()
```
and reinstall the header. 

CC @nunoguedelha

drdanz:
@CarlottaSartore Thanks for debugging and taking time to open this issue.
Would you like to try and create a pull request to fix this issue?

drdanz:
Fixed by #2110

PR #2110, opened 2019-10-16
fixed regex replace syntax error

CarlottaSartore:
Fixed syntax error of  issue  https://github.com/robotology/yarp/issues/2107


drdanz:
Merged, thanks.

PR #2105, opened 2019-10-09
os: Change layout for src/include file and rename library YARP_os

drdanz:
The rename follows the other libraries convention

There is no way to deprecate CMake targets yet (see https://gitlab.kitware.com/cmake/cmake/issues/17824), therefore we print a warning if the `OS` component is used, and a warnings every time `api.h` is included (unfortunately this does not work with `-isystem`).

PR #2112, opened 2019-10-29
Fix YarpConfig

Nicogene:
It should fix the issue introduced by #2105 .

It is a quite easy fix, the answer is @drdanz why did you put `_yarp_component_path_is_deprecated` ?
Do you forgot to define the macro? Or it was a typo?


drdanz:
Merged, thanks.

PR #2115, opened 2019-10-30
Change torque control input parameters "stiction_up" and "stiction_down" units.

nunoguedelha:
This is a dependency of https://github.com/robotology/icub-firmware/pull/101.

### DIFF
--- a/cmake/YarpIDL.cmake
+++ b/cmake/YarpIDL.cmake
@@ -168,7 +168,7 @@ function(_YARP_IDL_TO_DIR_GENERATE _family _file _name _index_file_name _output_
       set(_dest_dir "${_cpp_placement_dir}")
       set(_dest_file "${_gen_file}")
       if(NOT _cpp_placement_ns)
-        string(REGEX REPLACE "^.+/([^\/]+)$" "\\1" _dest_file "${_dest_file}")
+        string(REGEX REPLACE "^.+/([^/]+)$" "\\1" _dest_file "${_dest_file}")
       endif()
       set(_dest_file "${_dest_dir}/${_dest_file}")
     else()
--- a/src/libYARP_dev/src/ControlBoardHelper.cpp
+++ b/src/libYARP_dev/src/ControlBoardHelper.cpp
@@ -818,9 +818,9 @@ void ControlBoardHelper::convert_pid_to_user(const yarp::dev::PidControlTypeEnum
     out_usr.kd = out_usr.kd / output_conversion_units_user2raw;
     out_usr.max_output = out_usr.max_output / output_conversion_units_user2raw;
     out_usr.max_int = out_usr.max_int / output_conversion_units_user2raw;
-    out_usr.stiction_up_val = out_usr.stiction_up_val / output_conversion_units_user2raw;
-    out_usr.stiction_down_val = out_usr.stiction_down_val / output_conversion_units_user2raw;
-    out_usr.offset = out_usr.offset / output_conversion_units_user2raw;
+    out_usr.stiction_up_val = out_usr.stiction_up_val / mPriv->newtonsToSensors[k_usr];
+    out_usr.stiction_down_val = out_usr.stiction_down_val / mPriv->newtonsToSensors[k_usr];
+    out_usr.offset = out_usr.offset / mPriv->newtonsToSensors[k_usr];
 }
 
 Pid ControlBoardHelper::convert_pid_to_machine(const yarp::dev::PidControlTypeEnum& pidtype, const Pid &in_usr, int j_usr)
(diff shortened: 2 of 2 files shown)


nunoguedelha:
This branch has been tested on iCubGenova04:
- on `yarpmotorgui` in torque control mode with current output, on individual joints (legs, torso)
- running the Simulink model https://github.com/robotology/whole-body-controllers/blob/master/controllers/fixed-base-joints-control/jointsControl.mdl with the robot on the pole.

pattacini:
I didn't dig into the PR but I can confirm that `stiction_up` and `stiction_down` are used by fingers tuner in their present form.

drdanz:
@nunoguedelha Are you going to update this PR? Otherwise I'm going to close this...

nunoguedelha:
> @nunoguedelha Are you going to update this PR? Otherwise I'm going to close this...

Ciao @drdanz , yes sure, I'm doing the rebase in the next two days. I'm having a meeting with @traversaro regarding the conflicts between how the PR uses some parameters and their current use in `master` branch.

nunoguedelha:
> This PR cannot be accepted in its current form. This is a major breaking change: the measurement units of these parameters cannot be changed only because have been tested on a single robot with a modified firmware.
> If you need to modify the control scheme, I recommend using new parameters with new names, leaving untouched the semantic of the existing ones (there may exist code that specifically relies on it)

Absolutely @randaz81 , the plan was to test on other robots (purple iCub and R1), that's why I left the PR as a "Draft". But I ran out of time and never had bandwidth again to handle this because of other projects.

nunoguedelha:
We had a meeting @traversaro @pattacini and myself for discussing on this PR. Please find the minutes below...

## Meeting Minutes 06/03/2020

Participants: @nunoguedelha @traversaro @pattacini 

## Discussion Points & Conclusions

> 1. What is the definition of and how are the "stiction_up" and "stiction_down" parameters used in the **torque control improvement** feature related to this PR on one side, and the **fingers tuner** feature mentioned by Ugo on the other side.

We agreed that although the "stiction_up" and "stiction_down" parameters are not used the same way by the two features, the paramters represent the same physical properties: the static friction in both directions of rotation of a given joint, that needs to be overcome in order to set the joint in motion.

> 2. Currently the stiction parameters have "machine units" equivalent to raw PWM values, while Newton Meters would be more convinient to the **torque control improvement** feature, which motivated the change in this PR. Could we share the same units to be used by both features, and in that intent, could we consider changing those units in a future development, to Newton Meters, in case we drop this PR?

Ugo agreed that such change would be compatible with the **fingers tuner** feature. We agreed that it would be more convenient to have real physical units (Newton Meters), more suited to a controller API, since more independent from other settings, like typically the `Ktau`.
It would be more convenient, in general, to have the controllers configuration interface "closer" to the core component implementing those controllers and using that configuration.

> 3. In case the **fingers tuner** feature is also applied to the main joints of iCub (legs, torso, etc), how would that feature coexist with the **torque control improvement** from the point of view of handling the configuration parameters?

Multiple parameters sets can be selected via user defined tags (`POS_PID_DEFAULT`, `TRQ_PID_TUNED`, etc) for configuring specific control modes (velocity, torque, position, etc), which could be easily extended to a calibration mode, in order to avoid collisions between the two features.

> 4. Depending on the previous 2 points, is it necessary to create new parameters for the torque control improvement feature, or can we find a solution for sharing the already defined parameters?

We agreed on a solution that would be the less disruptive to the present architecture and the integration/release process. We can summarize it in the following action points ...

## Action Points

### Stiction parameters

1. Drop the present Pull Request #2115, no changes will be done on YARP. -> Nuno
2. The conversions done in YARP on the stiction parameters shall be undone by the **torque control improvement** feature in the icub-firmware, such that the feature sees the "stiction_up" and "stiction_down" as having the desired units: Newton Meters. -> Nuno
3. Open an issue for later doing a cleanup removing this workaround and implementing the direct transmission of the controllers/tuners configuration parameters, unchanged, to the respective target components (icub-firmware,high-level tuner software). -> Nuno or Ugo.
4. Open a ticket on iCub3 about the cogging issue. -> Silvio

### Feed-forward and PID parameters signs

1. Revert the sign changes of the parameters in the `robots-configuration/.../motorControl/` files (PID, Feed-forward, etc). -> Nuno
2. In `icub-firmware` (EMS firmware, controller code), use the absolute value of these parameters to configure the controllers. -> Nuno
3. Check if `Gearbox_M2J` sign is enough for adjusting the joint rotation direction, and the signs of the multiple feedback variables (e.g. joint position & veloicity), in case a cable or an encoder are mounted in the wrong way. -> Nuno
4. Document and present to Silvio and Ugo the new way how the motor input control variable (PWM, desired current, desired velocity) is mapped to:
    - a desired direction of rotation of the motor.
    - a desired sign of the motor/joint kinematic/dynamic feedback quantities.

nunoguedelha:
@traversaro @pattacini please feel free to correct anything I would have missed or anything unclear.
thanks!

drdanz:
> Drop the present Pull Request #2115, no changes will be done on YARP. -> Nuno

@nunoguedelha This means that we can close this, right? Please reopen if I got this wrong. Thanks

PR #2122, opened 2019-11-14
yarpdatadumper can handle dir name w/ special chars

pattacini:
This PR addresses #2116 so that special chars (`:`, `*`, `|`...) in the name of the dirs created by `yarpdatadumper` get replaced with the underscore. The chars `/`, '\' can be still used as it will allow for creating nested dirs.


pattacini:
@drdanz done 👍

drdanz:
Thanks @pattacini

PR #2129, opened 2019-11-28
Update dependencies

drdanz:
* CMake 3.12 or later is now required
* Optional dependency on YCM now requires version 0.10.4 or later.
* C++14 is now required
* CMake files are now installed in lib/cmake also on windows


drdanz:
For reference, the failures on appveyor are unrelated, this PR built successfull on appveyor and on travis before rebasing it, see https://ci.appveyor.com/project/robotology/yarp/builds/29187086 and https://travis-ci.org/robotology/yarp/builds/618518815

PR #2137, opened 2019-12-03
os/Log: Move implementation for colored output on Win32 in Log.cpp

drdanz:
@PeterBowman The WinConsole.c file was missing the copyright header, therefore the unit test checking that was failing... I'm not sure why, because I'm pretty sure that the unit tests passed for #2076 passed at least once. Did you remove it at some point?

Anyway, I moved the implementation in the Log.cpp file, and moved the `if` outside of the function call, since this will reduce to a jump after the first execution, otherwise it will have to call the function (which is more expensive) every time a log line is printed. It's a micro-optimization, but I think it is worth having it.


drdanz:
@PeterBowman I'm going to merge this, since it fixes the the CI that is currently broken, but feel free to add comments or open a new PR if you feel like something is not right.

PeterBowman:
Thank you, @drdanz, and sorry I got late! I opened https://github.com/robotology/yarp/pull/2138 to address an issue I found.

PR #2153, opened 2019-12-17
[WIP] Add cluster manager also on windows

### LINKED ISSUE #2152: Add Windows support for the cluster manager
OpenSSH client was introduced in the [Windows 10 Fall Creators Update](https://support.microsoft.com/en-us/help/4028685/windows-10-get-the-fall-creators-update), then we can remove the `#ifdef WIN32` inside the `clusterWidget`.

The problem here is to find a way to check in the code the OS version

Nicogene:
It fixes #2152 

Please review code.

TO BE TESTED.

### DIFF
--- a/src/yarpmanager/src-manager/mainwindow.cpp
+++ b/src/yarpmanager/src-manager/mainwindow.cpp
@@ -42,6 +42,7 @@
 #include <QFileDialog>
 #include <QDesktopServices>
 #include <QInputDialog>
+#include <QtCore/qoperatingsystemversion.h>
 
 #include <QWizardPage>
 #include <QLabel>
@@ -69,6 +70,8 @@
 #endif
 
 using namespace std;
+
+constexpr int32_t win10_1909{18363}; // see https://en.wikipedia.org/wiki/Windows_10#Updates_and_support
 
 
 bool isAbsolute(const char *path) {  //copied from yarp_OS ResourceFinder.cpp
@@ -170,13 +173,18 @@ MainWindow::MainWindow(QWidget *parent) :
     ui->actionAuto_reload->setChecked(true);
 
     ui->action_Manager_Window->setChecked(true);
-#ifdef WIN32
-    ui->tabWidgetLeft->tabBar()->hide();
-#else
     yarp::os::ResourceFinder& rf = yarp::os::ResourceFinder::getResourceFinderSingleton();
 
     std::string confFile = rf.findFileByName("cluster-config.xml");
-    if (!confFile.empty())
+    bool compatible{true};
+    auto current_version = QOperatingSystemVersion::current();
+    if (current_version.type() == QOperatingSystemVersion::Windows &&
+        current_version.majorVersion() < 10
+        && current_version.microVersion() < win10_1909) {
+        compatible = false;
+    }
+
+    if (compatible && !confFile.empty())
     {
         ui->clusterWidget->setConfigFile(confFile);
         ui->clusterWidget->init();
@@ -187,7 +195,6 @@ MainWindow::MainWindow(QWidget *parent) :
     {
         ui->tabWidgetLeft->tabBar()->hide();
     }
-#endif
 
 }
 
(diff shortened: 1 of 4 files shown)


src/yarpmanager/src-manager/mainwindow.cpp
```diff
@@ -70,6 +71,8 @@
 
 using namespace std;
 
+constexpr int32_t win10_1909{18363}; // see https://en.wikipedia.org/wiki/Windows_10#Updates_and_support
```
traversaro: Can we use a permanent URL? Either the Windows docs, or a link to a permanent wiki version (see https://en.wikipedia.org/wiki/Help:URL#WP)? Otherwise a Wikipedia article is always bound to change.

PeterBowman: There is a handy new [URL shortener extension](https://meta.wikimedia.org/wiki/Wikimedia_URL_Shortener) for MediaWiki: https://meta.wikimedia.org/wiki/Special:UrlShortener + https://w.wiki/E62.

drdanz:
Please note that if yarpmanager requires Qt 5.9 (which is totally fine from the [Supported Distributions](https://wiki.icub.org/wiki/YARP_Supported_Distributions) point of view), you should update also `cmake/YarpFindDependencies.cmake`.

drdanz:
@Nicogene is this still a wip?
Please add a line to the release notes to mention that Qt 5.9 or newer is required

Nicogene:
> @Nicogene is this still a wip?

Yes because I have to test somehow on windows, I am not sure how ssh works on windows.

> Please add a line to the release notes to mention that Qt 5.9 or newer is required

Sure :+1:

Nicogene:
In these days I gave a try to ssh on windows and it is required, besides build `1909` or greater, to enable the `OpenSSH Client/Server` optional features:

immagine

Once enable them I experienced some issues with `authorized_keys`, then I was able to ssh in `localhost` and `wsl` but always typing the password.
Checking on internet, I am not the only one having similar issues.
I have to check it later and understand which are the steps in order to make it work properly without password.

The cluster manager with the password is not usable, then I'd move this feature to the next release 😕

traversaro:
Comment by @lornat75 : we should probably make sure that this also works for launching GUIs.

drdanz:
> we should probably make sure that this also works for launching GUIs.

Just to clarify, this could mean:

1. windows guis should work when started from linux (**windows server**, linux client)
2. linux guis should work on windows (linux server, **windows client**)
3. windows guis should work on windows (**windows server**, **windows client**)

Moreover the gui can be displayed

4. on the local display (on the **client**)
5. on the remote display (on the **server**)

If I understood it correctly, @lornat75 request means that the `yarprun` server started on a **windows server** should be able to start GUI applications without forwarding the graphics to the client, which means that 
* We want to be sure that **1+5** and **3+5** are working
* 2 is already supported in both combinations, (2+4 requires the installation of a X server on windows)
* 1+4 are probably impossible to support due to the different technologies that are used to display the graphics on windows and linux.
* I have no idea whether 3+4 can work, but I don't think we need it

lornat75:
@drdanz I don't think we support the case GUI runs on windows and forward display on linux/windows, or at least, this is not a concern of yarp but a feature of the OS, isn't it?

About the problem I mentioned, there may be the case in which a GUI is spawned as a service and does not get access to the display (it runs, but you can't see anything), to stretch a comparison it is as if the display was > /dev/null :)

drdanz:
I just managed to enable ssh and enable the passwordless login following these 2 pages:

https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse
https://jmmv.dev/2020/10/windows-ssh-access.html

The issue with the authorized_keys seems to be that if your user is an administrator, instead of using `~/.ssh/authorized_keys`, the file to use is `C:\ProgramData\ssh\administrators_authorized_keys` and that file should not be readable by authenticated users (see the second link above).

Nicogene:
Closing since probably this feature is not needed anymore and this PR is very outdated

PR #2171, opened 2020-01-07
positionSensors added to MultipleAnalogSensorsInterfaces

randaz81:
Added a new interface `yarp::dev::IPositionSensors`, which can be used with localization sensors (e.g. UWB localization sensor) to broadcast a XYZ measurement of the device w.r.t. a fixed frame.
`MultipleAnalogSensorsServer` and `MultipleAnalogSensorsClient` have been modified accordingly.

### DIFF
--- a/doc/release/master/featureIPositionSensors.md
+++ b/doc/release/master/featureIPositionSensors.md
@@ -1,0 +1,24 @@
+featureIPositionSensors {master}
+----------------------
+
+### YARP_dev
+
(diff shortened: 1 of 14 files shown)


randaz81:
@traversaro 
> It seems fine, the only missing part is the handling of the interface in the `multipleanalogsensorsremapper`, but that can be also in a different PR, what do you think @drdanz ?

Thank you, I just added also the remapper.

PR #2186, opened 2020-01-21
Tests: Improve check_license.pl and add a very basic style check

drdanz:


Guadalupe5747863:
https://github.com/robotology/yarp/pull/2186 I aslo need my phone to be unlock from find my device I aslo have lots of errors in my internet that are not found like 500 404

PR #2223, opened 2020-03-03
yarpdataplayer is now able to reproduce ROS data types (and more!)

randaz81:
`yarpdataplayer` is now able to reproduce several ROS types, i.e.
 ```
sensor_msgs/LaserScan.
nav_msgs/Odometry
tf/tfMessage
tf2_msgs/tfMessage
geometry_msgs/Pose
geometry_msgs/Pose2D
```
Other types can be easily added just editing `utils.cpp` and `worker.cpp`. Only a couple of lines of code for message is needed! 
See also the documentation provided in the comment below.

_NOTE: (and more!) = all data types generated by yarp Thrift IDL_

### DIFF
--- a/doc/cmd_yarpdataplayer.dox
+++ b/doc/cmd_yarpdataplayer.dox
@@ -114,6 +114,16 @@ the player will successfully load all required data.
 The parts name will be taken from each subdirectory of the `/experiment1`
 forder.
 
+\section yarpdataplayer_ros Topic/ros compatibility
+
+Yarpdataplayer allows also to reproduce topics which can be subscribed by ROS nodes.
+For example, you can record a topic (e.g. /mytopic+@/mynode) as a standard yarp port using the Bottle format provided by yarpdatadumper. The Bottle format is needed because
+yarpdatadumper is currently not able to automatically detect the type of the incoming messages. 
+After the recording, you have to manually edit the info.log file created by yarpdatadumper and replace the type Bottle with the type of the ROS message you want to publish.
+For example, if you want to publish a sensor_msgs/LaserScan ROs Message, replace Bottle with sensor_msgs/LaserScan.
+That's all. When Yarpdataplayer plays the recorded data, it will open the /mytopic+@/mynode port and will publish LaserScan data. The corresponding topic will be visible also by ROS. 
+
+
 \note Currently, if the directory selected is (in this example) head,
 torso etc, the player will not load the files as it will not be able to
 find subdirs.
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by randaz81** [comment]
Added documentation in cmd_yarpdataplayer.dox, here reported for users' convenience.

`Yarpdataplayer `allows also to reproduce topics which can be subscribed by ROS nodes.
For example, you can record a topic (e.g. `/mytopic+@/mynode`) as a standard yarp port using the Bottle format provided by `yarpdatadumper`. 
The Bottle format is needed because `yarpdatadumper` is currently not able to automatically detect the type of the incoming messages. 
After the recording, you have to manually edit the `info.log` file created by yarpdatadumper and replace the type `Bottle `with the type of the ROS message you want to publish.
For example, if you want to publish a `sensor_msgs/LaserScan` ROS Message, replace `Bottle `with `sensor_msgs/LaserScan`.
That's all. When yarpdataplayer plays the recorded data, it will open the `/mytopic+@/mynode` port and will publish LaserScan data. The corresponding topic will be visible also by ROS.

**Comment by Nicogene** [comment]
> After the recording, you have to manually edit the info.log file created by yarpdatadumper and replace the type Bottle with the type of the ROS message you want to publish.

I'd open also a PR for `yarpdatadumper` for adding the ROS types, in order to not have to change the info.log manually.

**Comment by randaz81** [comment]
@nicogene it works perfectly, but unfortunately I cannot get rid of a GCC warning about a forward declaration, so I'm not merging it yet. Any help is welcome :-)

**Comment by Nicogene** [comment]
> it works perfectly, but unfortunately I cannot get rid of a GCC warning about a forward declaration,
Which warning?

**Comment by randaz81** [comment]
> Which warning?

On VS2019 I have no warnings.
On gcc 7.4.0, I have tons of warnings similar to:
```
In file included from /usr/local/src/robot/yarp/src/yarpdataplayer/include/utils.h:33:0,
                 from /usr/local/src/robot/yarp/src/yarpdataplayer/include/mainwindow.h:34,
                 from /usr/local/src/robot/yarp/src/yarpdataplayer/src/main.cpp:19:
/usr/local/src/robot/yarp/src/yarpdataplayer/include/worker.h:99:22: warning: invalid use of incomplete type ‘class Utilities’
         if (utilities->sendStrict) {
                      ^~
In file included from /usr/local/src/robot/yarp/src/yarpdataplayer/include/utils.h:33:0,
                 from /usr/local/src/robot/yarp/src/yarpdataplayer/include/mainwindow.h:34,
                 from /usr/local/src/robot/yarp/src/yarpdataplayer/src/mainwindow.cpp:19:
/usr/local/src/robot/yarp/src/yarpdataplayer/include/worker.h:44:7: note: forward declaration of ‘class Utilities’
```
But I do not know how to fix this. The original code is pretty intricate with a couple of .h files including each other. On the top of this, my recent addition of a template function which uses a class defined elsewhere generates the warning.

PR #2240, opened 2020-03-30
os: Increase the precision of printed floating point numbers

drdanz:
DBL_DIG is not enough to ensure that the number can be read back, a
better option would be to use DBL_DECIMAL_DIG, but this is available
only in C++17, hence we use DECIMAL_DIG.

#### `Bottle`

* Fixed precision for floating point numbers, to ensure that they are read as
  the original floating point written.


drdanz:
More explanations, in case someone is interested, can be found here: https://stackoverflow.com/questions/16839658

PR #2245, opened 2020-04-09
yarpview: Checkable menu issue fixed

elandini84:
Now the "checked" state of the two checkable elements in the File menu
of "yarpview" (i.e. "Save single image..." and "Save a set of
images...") turns to false when the corresponding dialogs are closed.


drdanz:
Merged, thanks.

PR #2249, opened 2020-04-23
yarp companion: QOS related commands added

elandini84:
* Two new commands have been added to the `yarp companion` related to the QOS of a given connection
  * `priority-qos` That allows to set/get the packet priority value of both the<br/>source and the destination port
  * `priority-sched` That allows to set/get the thread priority and policy values of both the<br/>source and the destination port

### DIFF
--- a/doc/release/master/feature_companion_qos_command.md
+++ b/doc/release/master/feature_companion_qos_command.md
@@ -1,0 +1,10 @@
+feature_companion_qos_command {#yarp_3_3}
+---------------
+
+### Tools
+
+#### `yarp companion`
+
+* Two new commands have been added to the `yarp companion` related to the QOS of a given connection
+  * `priority-qos` That allows to set/get the packet priority value of both the<br/>source and the destination port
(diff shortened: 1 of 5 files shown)


drdanz:
@elandini84 thanks for the patch! I didn't go through the whole code, but here are a few initial comments:

* This is a new feature, therefore it should be merged in master. I changed the base branch and rebased.

* The license check is failing because the Companion.qos.cpp doesn't have any license...
It should be this one:
```
/*
 * Copyright (C) 2006-2020 Istituto Italiano di Tecnologia (IIT)
 * All rights reserved.
 *
 * This software may be modified and distributed under the terms of the
 * BSD-3-Clause license. See the accompanying LICENSE file for details.
 */
```

* Linux/macos builds did not even start because the license is missing

* Windows builds are failing because the `editline/readline.h` file does not exist on windows. Are you actually using this file? AFAIK It is supposed to be used for interactive programs.

* I noticed a few `Please Trye again` instead of `Please try again` :wink:

drdanz:
Merged, thanks!

PR #2253, opened 2020-04-24
os: Add Clock::isClockInitialized and Network::isNetworkInitialized methods

drdanz:
##### `Time`

* Added `bool isClockInitialized()` method

##### `Network`

* Added `bool isNetworkInitialized()` method

### DIFF
--- a/doc/release/master/isNetworkInitialized.md
+++ b/doc/release/master/isNetworkInitialized.md
@@ -1,0 +1,14 @@
+isNetworkInitialized {#master}
+--------------------
+
+### Libraries
+
+#### `os`
+
+##### `Time`
+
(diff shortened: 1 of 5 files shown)


Nicogene:
@drdanz what is the difference of `isNetworkInitialized()` and the existing `checkNetwork()`?

drdanz:
@Nicogene `checkNetwork()` checks if the `yarp server` is reachable, network must be initialized before being able to use it.
`isNetworkInitialized()` is just a check to ensure that the user called `yarp::os::Network::init()` (or created a `yarp::os::Network` object. It's not meant to be used by the user, but it is a check that can be used internally to be sure that the user initialized the network before starting any network-related operation

PR #2254, opened 2020-04-24
Remove executable path from port name from LogForwarder and yarprun

drdanz:
##### `Log`

* The name of the port used to forward the output no longer contains the full
  path, but only the executable name.

#### `yarprun`

* The name of the port used to forward the output no longer contains the full
  path, but only the executable name.


Nicogene:
cc @vtikha @valegagge

PR #2256, opened 2020-04-24
YARP_OS/Log: Big refactor

drdanz:
##### `Log`

* Add `LogComponent` and component macros
  * A log component can be declared using the `YARP_LOG_COMPONENT()` macro (in
    a `.cpp` file) and eventually forward declared using
    ` YARP_DECLARE_LOG_COMPONENT()`.
  * the component can then be used in the new macros:
    - `yCTrace`
    - `yCDebug`
    - `yCInfo`
    - `yCWarning`
    - `yCError`
    - `yCTrace`
  * It is possible to set custom printing and forwarding macros for each
    component, and it is possible to enable and disable specific levels for each
    component.
  * At the moment it is not yet possible to change the defaults set in the code,
    but this will be enabled in the future

* The output forwarded is now a property and contains several useful
  information including component and system/network time, file, line,
  function. The following EXPERIMENTAL environment variables can be
  enabled to enable forwarding of some extra information:
  - `YARP_FORWARD_CODEINFO_ENABLE`
  - `YARP_FORWARD_HOSTNAME_ENABLE`
  - `YARP_FORWARD_PROCESSINFO_ENABLE`
  - `YARP_FORWARD_BACKTRACE_ENABLE`
  These environment variables could be removed in the future.

* The logger is now able to detect if it is running in yarprun and
  eventually change the output format.

* Logging is now protected by a mutex to avoid garbled output.

* The `YARP_DEBUG_LOG_ENABLE` environment variable can be set to debug
  the log of the application.

* When color is enabled, the loglevel is represented by a single colored
  character.

Depends on #2251
Depends on #2252 
Depends on #2253 
Depends on #2255


src/libYARP_os/src/yarp/os/Log.cpp
```diff
@@ -35,432 +47,809 @@
...
+    static void default_print_callback(yarp::os::Log::LogType type,
+                                       const char* msg,
+                                       const char* file,
+                                       const unsigned int line,
+                                       const char* func,
+                                       double systemtime,
+                                       double networktime,
+                                       const char* comp_name)
+    {
+        if (auto cb = current_print_callback.load()) {
```
PeterBowman, line 106: [If-with-init-statement](https://en.cppreference.com/w/cpp/language/if) is a C++17 thing, but YARP supports C++14, yet:

https://github.com/robotology/yarp/blob/4d561eefbdda1dae9870e8c19e0b59a698ddbbcf/cmake/YarpSystemCheck.cmake#L27

https://github.com/robotology/yarp/blob/f6310fb0cab38605115f20b7fe83c52fbd8f41a7/src/libYARP_os/src/CMakeLists.txt#L413

drdanz: Are you sure about this? According to the link you posted, "any init-statement must end with a semicolon `;`". From my limited knowledge of the standard, this is a c++11 (since it uses `auto`) "declaration of a single non-array variable with a brace-or-equals initializer"

Also this test works properly on most compilers supporting c++11... https://godbolt.org/z/zVK5yP

If I understand correctly, the main difference is that using the c++17 init statement, the `x` variable in the example can be used also in the else branch, and it is destroyed before returning to the upper scope, in this case the variable is destroyed at the end of the `if` branch, but since there is no `else` branch here, this should not be an issue...

PeterBowman: You are totally right and I misread that part, my bad...

PR #2257, opened 2020-04-24
logger: Interpret the new message format containing extra info

drdanz:
#### `logger`

* Interpret the new message format containing extra info.

Depends on #2256


Nicogene:
cc @vtikha @valegagge

PR #2259, opened 2020-04-24
[WIP] Use new logging almost everywhere

drdanz:
* Deprecate setVerbosity
* Deprecate --verbose and --quiet
* Get rid of old Logger

Depends on #2256


randaz81:
I'm just only a little worried about using the logger inside port components. For example, if a warning occurs inside a port and log forwarding is enabled (thus opening a new port), may it trigger an infinite recursion? But hopefully I'm wrong...

drdanz:
> I'm just only a little worried about using the logger inside port components. For example, if a warning occurs inside a port and log forwarding is enabled (thus opening a new port), may it trigger an infinite recursion? But hopefully I'm wrong...

This is possible, and that's the reason why forwarding of *core* parts is disabled by default.
At the moment, the only way to enable it, is to change the code where the components are declared, for example in `PortCore.cpp`

```c++
namespace {
YARP_LOG_COMPONENT(PORTCORE,
                   "yarp.os.impl.PortCore",
                   yarp::os::Log::InfoType,
                   yarp::os::Log::LogTypeReserved,
                   yarp::os::Log::defaultPrintCallback(),
                   nullptr)
} // namespace
```

The forward callback is set to `nullptr`, therefore it will never forward anything.

If the user modifies this `nullptr`, then it is its fault, though.

drdanz:
Closing this in favour of smaller PRs (#2263, #2264, #2265 and more to come)

PR #2271, opened 2020-05-05
Port a few carriers to the new logging system

drdanz:


drdanz:
I'm not 100% sure about this, perhaps carriers should be considered the same as YARP_os, and therefore always disable the log forwarding

PR #2282, opened 2020-05-21
yarpview: Pixel rgb value in status bar

elandini84:
* Added a checkable menu item that, if checked, shows an additional line in the `status bar` that display the color value of the pixel pointed by the mouse cursor.
  
  The string has the following format 
  ```
  Pixel("x","y") = "hexstring"
  ```
  Where  `x` and `y` are the coordinates of the pixel and `hexstring` is the hexadecimal (in the ARGB format) string representing the pixel color
  
  The additional line on the `status bar` contains also a little rectangle that will turn the same color of the currently selected pixel.

### DIFF
--- a/doc/release/master/feature_yarpview_color_picker.md
+++ b/doc/release/master/feature_yarpview_color_picker.md
@@ -1,0 +1,16 @@
+fix_yarpview_color_picker {#master}
+---------------
+
+### GUIs
+
+#### `yarpview`
+
+* Added a checkable menu item that, if checked, shows an additional line in the `status bar` that displays the color value of the pixel pointed by the mouse cursor.
+  
+  The string has the following format 
+  ```
+  Pixel("x","y") = "hexstring"
+  ```
+  Where  `x` and `y` are the coordinates of the pixel and `hexstring` is the hexadecimal (in the ARGB format) string representing the pixel color
+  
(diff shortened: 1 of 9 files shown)


elandini84:
> Thanks, that a very nice feature... I have a few questions though
> 
> 1. The value seems to be updated only when the mouse is moved, if the value of the pixel changes under the mouse, this is not updated. Is this intended?
> 2. I could not find a way to copy the value of the pixel, the only option seems to be memorizing the code, and copying it manually, but when you move the mouse, the value obviously disappears. Perhaps we should have a right click menu to copy the value on the clipboard? Or "lock" the color on the clicked value?
> 
> Anyway, LGTM, we can merge this, and eventually improve it later

Regarding the first question, the value is updated that way because I intended this feature as a way to go and check the value of a particular pixel and not also to monitor that same value over time. But if you think that it would be better that way, I can go and change it.
For the second question, I did not think about letting the user copy the value. Probably the best way is to add the pixel value to the clipboard by right clicking on the pixel.

Let me know if you think that these two changes (or one of them) are necessary to make this new feature really useful.

drdanz:
> Let me know if you think that these two changes (or one of them) are necessary to make this new feature really useful.

To be honest, I have no idea, let's just merge it this way, and eventually change it later.

drdanz:
Merged, thanks.

PR #2285, opened 2020-05-29
Fix bug of use forward/print callback user defined 

valegagge:
In this PR, I fix a bug regarding the call of forward/print callback: the default component was created using the default-forward/default-print callback instead of the user defined one; in this way, the user callbacks were not invoked.

I fix this bug, creati g the default component with the user callbacks if they are defined.

To test this patch I defined a forward user callback in embobjMotionControl
```
static void testForwarderLog(yarp::os::Log::LogType t,
                                 const char* msg,
                                 const char* file,
                                 const unsigned int line,
                                 const char* func,
                                 double systemtime,
                                 double networktime,
                                 const char* comp_name)
{
    std::string type = "type:unknown" ;
    switch(t)
    {
        case yarp::os::Log::LogType::TraceType: type = "type:Trace" ;
            break;
        case yarp::os::Log::LogType::DebugType: type = "type:Debug" ;
            break;
        case yarp::os::Log::LogType::InfoType: type = "type:Info" ;
            break;
        case yarp::os::Log::LogType::WarningType: type = "type:Warning" ;
            break;
        case yarp::os::Log::LogType::ErrorType: type = "type:Error" ;
            break;
        case yarp::os::Log::LogType::FatalType: type = "type:Fatal" ;
            break;
        default:
            break;
    };
    
    
    yarp::os::Log::LogCallback defaultcbk=yarp::os::Log::defaultForwardCallback();

    defaultcbk(t,
               msg,
               file,
               line,
               func,
               systemtime,
               networktime,
               comp_name);
    
    std::cout << "@@@TEST USER-FW-CBK: ==> " << type << "  @" <<systemtime << " " <<msg << std::endl;

}
```

and set it in the open function

```
bool embObjMotionControl::open(yarp::os::Searchable &config)
{
    // - first thing to do is verify if the eth manager is available. then i parse info about the eth board.

    yarp::os::Log::setForwardCallback(testForwarderLog);
    //yErrorCustomTime(111.22) << "embObjMotionControl::open(): first test of timed y*";
    
    ethManager = eth::TheEthManager::instance();
....
```

The output I get is this:
Screenshot from 2020-05-29 16-01-27

As you can see, each debug message is duplicated by the forward callback invocation.

### DIFF
--- a/src/libYARP_os/src/yarp/os/Log.cpp
+++ b/src/libYARP_os/src/yarp/os/Log.cpp
@@ -764,22 +764,11 @@ yarp::os::Log::LogCallback yarp::os::Log::defaultForwardCallback()
 // BEGIN Log Components
 const yarp::os::LogComponent& yarp::os::Log::defaultLogComponent()
 {
-    yarp::os::Log::LogCallback fwCbk = defaultForwardCallback();
-    if(nullptr != forwardCallback())
-    {
-        fwCbk = forwardCallback();
-    }
-    
-    yarp::os::Log::LogCallback printCbk = defaultPrintCallback();
-    if(nullptr != printCallback())
-    {
-        printCbk = printCallback();
-    }
     static const yarp::os::LogComponent component(nullptr,
                                                   defaultMinimumPrintLevel(),
                                                   defaultMinimumForwardLevel(),
-                                                  printCbk,
-                                                  fwCbk);
+                                                  printCallback(),
+                                                  forwardCallback());
     return component;
 }
 
--- a/src/libYARP_os/src/yarp/os/Log.cpp
+++ b/src/libYARP_os/src/yarp/os/Log.cpp
@@ -764,11 +764,22 @@ yarp::os::Log::LogCallback yarp::os::Log::defaultForwardCallback()
 // BEGIN Log Components
 const yarp::os::LogComponent& yarp::os::Log::defaultLogComponent()
 {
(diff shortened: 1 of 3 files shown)


drdanz:
Thanks @valegagge for catching this bug and for the patch, and sorry for the delay.

The patch fixes only part of the problem though...

The point for having `defaultForwardCallback()` (always pointing to the default callback) and `forwardCallback()` (always pointing to the callback that is currently set) is to be able to change it temporarily and restore the original behaviour anytime. This means that initializing the component to  `defaultForwardCallback()` is always wrong, because it does not allow to change it later. `forwardCallback()` will always call the current callback, and therefore it should be always used to initialize the default component. Therefore the `if` is still buggy here, if you call a `yDebug()` before calling `setForwardCallback`, the component will be initialized with the default callback, and therefore will not be changed.

The same is valid for the print callback, but also for the print and forward level.

Therefore I would just change the initialization to 

```c++
const yarp::os::LogComponent& yarp::os::Log::defaultLogComponent()
{
    static const yarp::os::LogComponent component(nullptr,
                                                  minimumPrintLevel(),
                                                  minimumForwardLevel(),
                                                  printCallback(),
                                                  forwardCallback());
    return component;
}
```

@valegagge Does this sound correct to you? (Since I know you are currently on leave, if I don't hear from you, I will just fix the PR and merge it)

drdanz:
Merged, thanks.

drdanz:
I didn't notice this was targetting `devel` :sweat:
Please use `master` next time.

PR #2294, opened 2020-06-17
os/PortCore: Do not set TOS to 0 if DSCP is an unknown string

drdanz:
Also improve debugging by adding a few debug messages


drdanz:
Travis did not report the status back, but the build was successful

PR #2306, opened 2020-07-06
devices/ovrheadset: Port to the new logging system

drdanz:
Unfortunately at the moment I don't have a way to check if this compiles and works
Can anyone with an oculus and a windows machine try this? Perhaps @GiulioRomualdi?


drdanz:
I'm merging this, if anyone tries it and has issues, please let me know.

PR #2310, opened 2020-07-09
Port dev/PolyDriver and most devices to the new logging system

drdanz:


drdanz:
Removed commits changing `transformClient` and `transformServer` to avoid conflicts with #2317 
The other changed should be safe and tested by the CI.
Merging

PR #2312, opened 2020-07-09
Add IMotor::setGearboxRatio in control board client-server pair

PeterBowman:
This patch exposes [`yarp::dev::IMotor::setGearboxRatio`](http://www.yarp.it/classyarp_1_1dev_1_1IMotor.html#aacb74f2baff8e03481adb515edfe8d3e) from wrapped devices, previously was missing.

### DIFF
--- a/doc/release/yarp_3_3/fix_controlBoard_setGearboxRatio.md
+++ b/doc/release/yarp_3_3/fix_controlBoard_setGearboxRatio.md
@@ -1,0 +1,12 @@
+fix_controlBoard_setGearboxRatio {#yarp_3_3}
+------------------
+
+### Devices
+
+#### `controlboardwrapper2`
+
+* Exposed missing `yarp::dev::IMotor::setGearboxRatio` method via RPC.
+
+#### `remote_controlboard`
+
+* Implemented missing `yarp::dev::IMotor::setGearboxRatio` method.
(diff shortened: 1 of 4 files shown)


PeterBowman:
For the record, this was successfully tested with a fake device. Messages are correctly parsed by the control board wrapper.

```cmake
# CMakeLists.txt
cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
project(yarp-rcb-imotor LANGUAGES CXX)
find_package(YARP 3.3 REQUIRED COMPONENTS os dev)
add_executable(yarp-rcb-imotor main.cpp)
target_link_libraries(yarp-rcb-imotor YARP::YARP_os YARP::YARP_init YARP::YARP_dev)
```

```c++
// main.cpp
#include <yarp/os/Network.h>
#include <yarp/os/Property.h>
#include <yarp/os/Value.h>

#include <yarp/dev/IMotor.h>
#include <yarp/dev/PolyDriver.h>

int main()
{
        yarp::os::Network yarp;

        yarp::os::Property options {{"device", yarp::os::Value("remote_controlboard")},
                                    {"local", yarp::os::Value("/yarp_rcb_imotor")},
                                    {"remote", yarp::os::Value("/fake")}};

        yarp::dev::PolyDriver device(options);

        if (!device.isValid())
        { return -1; }

        yarp::dev::IMotor * iMotor;

        if (!device.view(iMotor))
        { return -1; }

        double val;

        if (!iMotor->getGearboxRatio(0, &val))
        { return -1; }

        if (!iMotor->setGearboxRatio(0, 1.0))
        { return -1; }

        return 0;
}
```

drdanz:
The continuous integration failed for reason that seem to be unrelated to this patch...
I restarted the failed builds, just to be sure, but I'll just merge anyway if they fail again

drdanz:
Merged, thanks.

PR #2313, opened 2020-07-10
Fetch actual getInteractionModes data from stream

PeterBowman:
Typo fix. Prior to this, `getInteractionModes` was looking for control modes instead.


drdanz:
Merged, thanks.

PR #2320, opened 2020-07-15
fix_bug_checkNearToLocation

randaz81:
The method `checkNearToLocation()` now correctly checks if two orientations are similar (below a certain threshold), also considering the critical points 0,180,360,-180,-360 etc.


randaz81:
> LGTM (even though I would suggest to add an extra comment to the `#if 1`)

Added explanation, thank you.

drdanz:
Merged, thanks.

PR #2330, opened 2020-07-27
Rename test_grabber -> fakeFrameGrabber

drdanz:
### Devices

* The `test_grabber` was renamed `fakeFrameGrabber`. The old name is kept for
  compatibility, but it will be deprecated and removed in a future release.

CC-Issue #2308


drdanz:
Macos builds are broken again for other reasons, the other builds are successfull. Merging.

PR #2333, opened 2020-07-28
[Discussion] Add yarp::os::getEnvironment(), yarp::os::setEnvironment() and yarp::os::unsetEnvironment() functions and deprecate NetworkBase version.

### LINKED ISSUE #1853: Network.h contains methods that should be somewhere else
`Network` for yarp have ever been a *superclass* that does a lot of stuff, some of them not directly related to the **network** or communication.

Giving a look to `Network.h` I found that these methods:
```c++
    /**
     *
     * Read a variable from the environment.
     *
     * @param key the variable to read
     * @param found an optional variable to set to true iff variable is found
     * @return the value of the environment variable, or "" if not found
     *
     */
    static std::string getEnvironment(const char *key,
                                      bool *found = nullptr);

    /**
     *
     * Set or change an environment variable.
     *
     * @param key the variable to set or change
     * @param val the target value
     *
     */
    static void setEnvironment(const std::string& key, const std::string& val);

    /**
     *
     * Remove an environment variable.
     *
     * @param key the variable to remove
     *
     */


    /**
     *
     * Get an OS-appropriate directory separator (e.g. "/" on linux)
     *
     */
    static std::string getDirectorySeparator();

    /**
     *
     * Get an OS-appropriate path separator (e.g. ":" on linux)
     *
     */
    static std::string getPathSeparator();
```

should stay somewhere else, in a sort of utility header, what do you think about it?

This should make `Network` a lighter class, that does only what it have to do, making it easier to handle and maintain.

This is not a blocking issue, but I think it is something to not ignore because this(like other cases) causes "cross-dependencies" because some things are not in the right place.
For example see this [comment](https://github.com/robotology/yarp/pull/1852#discussion_r216931763)

drdanz:
This PR is a "Discussion PR" (which is a term I just made up), I will not merge this without a proper discussion.

-----------

#### `os`

* Add `yarp::os::getEnvironment()`, `yarp::os::setEnvironment()` and `yarp::os::unsetEnvironment()` functions.

##### `Network`

* `getEnvironment()` is now deprecated in favour of `yarp::os::getEnvironment()`
* `setEnvironment()` is now deprecated in favour of `yarp::os::setEnvironment()`
* `unsetEnvironment()` is now deprecated in favour of `yarp::os::unsetEnvironment()`

-----------

As discussed in several occasions (#1853), there is no reason for having these static methods in `yarp::os::Network`, it is actually misleading and a source of circular dependencies. `Network` should be a class related to "YARP Network", and the system environment has nothing to do with that.

This patch moves these methods in the `yarp::os` namespace and deprecates the `yarp::os::Network` and `yarp::os::NetworkBase` versions

It's quite late for the release, but this should be a rather "safe" change, but there is a deprecation involved.

In #1853 @Nicogene proposed to move these methods in the `yarp::conf::environment` namespace, but these methods depend on functions that are currently in `yarp::os` and that depend on stuff in `yarp::os::impl` therefore it is not so easy to move them outside `yarp::os`.
Therefore, even if this is probably not an ideal solution, I suggest to move these methods in the `yarp::os` namespace, removing in this way the circular dependencies. AFAIK, here is no c++ proposal for handling the environment variables yet, therefore the `yarp::conf` version could be reserved in case there will be one in the future, and could have the same API.

CC @pattacini, @vtikha, @traversaro, @randaz81, @Nicogene, @elandini84, @lornat75, @robotology/yarp-developers 

Fixes #1853

### DIFF
--- a/doc/release/master/os_getEnvironment.md
+++ b/doc/release/master/os_getEnvironment.md
@@ -1,0 +1,15 @@
+os_getEnvironment {#master}
+-----------------
+
+### Libraries
+
+#### `os`
+
+* Add `yarp::os::getEnvironment()`, `yarp::os::setEnvironment()` and
+  `yarp::os::unsetEnvironment()` functions.
+
+##### `Network`
+
+* `getEnvironment()` is now deprecated in favour of `yarp::os::getEnvironment()`
+* `setEnvironment()` is now deprecated in favour of `yarp::os::setEnvironment()`
+* `unsetEnvironment()` is now deprecated in favour of `yarp::os::unsetEnvironment()`
--- a/src/carriers/tcpros_carrier/RosLookup.cpp
+++ b/src/carriers/tcpros_carrier/RosLookup.cpp
@@ -12,6 +12,7 @@
 #include <yarp/os/Bottle.h>
 #include <yarp/os/ContactStyle.h>
 #include <yarp/os/Network.h>
+#include <yarp/os/Os.h>
 
 #include <cstdlib>
 
@@ -117,7 +118,7 @@ bool RosLookup::lookupTopic(const std::string& name) {
 }
 
 yarp::os::Contact RosLookup::getRosCoreAddressFromEnv() {
-    std::string addr = NetworkBase::getEnvironment("ROS_MASTER_URI");
+    std::string addr = yarp::os::getEnvironment("ROS_MASTER_URI");
     Contact c = Contact::fromString(addr);
     if (c.isValid()) {
         c.setCarrier("xmlrpc");
--- a/src/carriers/unix/UnixSocketCarrier.cpp
+++ b/src/carriers/unix/UnixSocketCarrier.cpp
@@ -39,20 +39,20 @@ std::string getYARPRuntimeDir()
     }
 
     // Check YARP_RUNTIME_DIR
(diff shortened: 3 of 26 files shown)


drdanz:
If you don't want to deprecate the `Network` methods right now, we could just add these functions and use them internally, without deprecating the old ones, and maybe deprecate them in the next release cycle

drdanz:
We should also investigate whether this is still necessary:

https://github.com/robotology/yarp/blob/f46557b46780f23502f59f902ef78fe1637a94af/src/libYARP_os/src/yarp/os/impl/PlatformStdlib.h#L28-L52

pattacini:
Ok for me 👍🏻

We could also proceed with the deprecation of old methods already at this early stage as I guess they're not so ubiquitous in the user codebase.

I don't have particular concerns about that.

drdanz:
I created a separate PR where I moved these methods in the `yarp::conf::environment` namespace, and I removed also `yarp::os::getenv` (which doesn't have any sense, since there is a `std::getenv` method in c++98.

The result is that the code is cleaner, and I removed some dependencies from ACE in `yarp::os`, and one dependency from `yarp::os::impl` in `yarprun`.

I vote for the `yarp::conf` version, but it is slightly less safe and will require some extra testing because it is not just moving one method from one place to another

drdanz:
Withdrawing this in favour of #2334

PR #2334, opened 2020-07-28
[Discussion] Add yarp::conf::environment::{get,set,unset}Environment() functions and deprecate NetworkBase version and yarp::os::getenv

### LINKED ISSUE #1853: Network.h contains methods that should be somewhere else

drdanz:
This is the alternative version for #2333 which moves these functions in the `yarp::conf::environment` namespace.

Please keep the discussion on #2333

---------

#### `conf`

* Added the `yarp/conf/environment.h` header.
* Added the following functions:
  * `yarp::conf::environment::getEnvironment()`
  * `yarp::conf::environment::setEnvironment()`
  * `yarp::conf::environment::unsetEnvironment()`

#### `os`

* The method `yarp::os::getenv()` is now deprecated in favour of `std::getenv()`

##### `Network`

* `getEnvironment()` is now deprecated in favour of `yarp::conf::environment::getEnvironment()`
* `setEnvironment()` is now deprecated in favour of `yarp::conf::environment::setEnvironment()`
* `unsetEnvironment()` is now deprecated in favour of `yarp::conf::environment::unsetEnvironment()`

--------

Fixes #1853

### DIFF
--- a/doc/release/master/conf_getEnvironment.md
+++ b/doc/release/master/conf_getEnvironment.md
@@ -1,0 +1,22 @@
+conf_getEnvironment {#master}
+-------------------
+
+### Libraries
+
+#### `conf`
+
+* Added the `yarp/conf/environment.h` header.
+* Added the following functions:
+  * `yarp::conf::environment::getEnvironment()`
+  * `yarp::conf::environment::setEnvironment()`
+  * `yarp::conf::environment::unsetEnvironment()`
+
+#### `os`
+
+* The method `yarp::os::getenv()` is now deprecated in favour of `std::getenv()`
+
+##### `Network`
+
+* `getEnvironment()` is now deprecated in favour of `yarp::conf::environment::getEnvironment()`
+* `setEnvironment()` is now deprecated in favour of `yarp::conf::environment::setEnvironment()`
+* `unsetEnvironment()` is now deprecated in favour of `yarp::conf::environment::unsetEnvironment()`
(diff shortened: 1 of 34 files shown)


drdanz:
Travis was successfull, but for some reason appveyor did not start

PR #2338, opened 2020-07-29
YARP sig: Fix yarp::sig::Image::synchronize method

elandini84:
##### `Image`

* Fixed the `yarp::sig::Image::synchronize` method that missed the pixel type and the orientation settings needed to properly use the `move assignment operator`
  * A `getTypeId` method has been added to `yarp::sig::ImageStore` to be able to read the <i>protected</i> value `type_id`


drdanz:
Merged, thanks.

PR #2341, opened 2020-07-31
LaserFromRosTopic

randaz81:
added `LaserFromRosTopic` device, which exposes on Yarp a lidar ROS topic (dataype `sensor_msgs::LaserScan`)
 using yarp `IRangefinder2D` interface.


drdanz:
Merged, thanks.

PR #2342, opened 2020-08-04
Fixed realsense2tracking device timestamp

randaz81:
Fixed missing timestamp for device `realsense2tracking`
added option `--timestamp yarp` : yarp timestamp will be used
added option `--timestamp realsense` : realsense timestamp will be used
default value is yarp timestamp.


drdanz:
Merged, thanks.

PR #2357, opened 2020-09-11
FakeFrameGrabber fix

randaz81:
The ball test is now displayed in the center of the image instead of the right-bottom corner. The bug was affecting windows only. The bug was introduced by a recent refactor in which some variables were modified from int to size_t


drdanz (approved):

drdanz:
Tested locally... merging

drdanz:
Merged, thanks.

PR #2194, opened 2020-02-05
Fix video generation in yarpdatadumper

pattacini:
This PR addresses and solves #2193.

I've tested it successfully running the following:
- `yarpdev --device test_grabber`
- `yarpdatadumper --type video --connect /grabber`

PR #2360, opened 2020-09-11
Fix compilation against OpenCV 2.x in yarpdatadumper

PeterBowman:
Per https://github.com/robotology/yarp/pull/2194#issuecomment-691230360:

> This change is not compatible with OpenCV 2.4.9.1 (latest `apt` package on Ubuntu 16.04). There is no `cv::VideoWriter::fourcc` function in OpenCV 2.x ([ref](https://docs.opencv.org/ref/2.4/dd/d9e/classcv_1_1VideoWriter.html), cf. [OpenCV 3.x](https://docs.opencv.org/3.4/dd/d9e/classcv_1_1VideoWriter.html)).
>
> ```
> /home/teo/repos/yarp/src/yarpdatadumper/main.cpp: In member function ‘virtual void DumpThread::run()’:
> /home/teo/repos/yarp/src/yarpdatadumper/main.cpp:489:69: error: ‘fourcc’ is not a member of ‘cv::VideoWriter’
>                  videoWriter.open(videoFile.c_str(),cv::VideoWriter::fourcc('H','F','Y','U'),
>                                                                      ^~~~~~
> ```
> (...)

In case this patch is approved (might not be since OpenCV 2.x is about to lose support, I guess), please consider backporting it to `yarp-3.3` as the change that broke compatibility with OpenCV 2.x was released in YARP 3.3.2 (https://github.com/robotology/yarp/commit/e10d44326b28ac2e1864f77dd98fcbe20a3a45a8). Besides, I noticed AppVeyor builds do **not** consume OpenCV despite version 2.4.13 being present.

### DIFF
--- a/doc/release/yarp_3_4/fix_yarpdatadumper_opencv2.md
+++ b/doc/release/yarp_3_4/fix_yarpdatadumper_opencv2.md
@@ -1,0 +1,6 @@
+fix_yarpdatadumper_opencv2 {#yarp_3_4}
+-----------
+
+### yarpdatadumper
+
+* Fixed a regression that broke support for OpenCV 2.x.
--- a/src/yarpdatadumper/main.cpp
+++ b/src/yarpdatadumper/main.cpp
@@ -486,7 +486,11 @@ public:
                 double dt=itemEnd.timeStamp.getStamp()-t0;
                 fps=(dt<=0.0)?25:int(double(sz-1)/dt);
 
+#if CV_MAJOR_VERSION >= 3
                 videoWriter.open(videoFile.c_str(),cv::VideoWriter::fourcc('H','F','Y','U'),
+#else
+                videoWriter.open(videoFile.c_str(),CV_FOURCC('H','F','Y','U'),
+#endif
                                  fps,cvSize(frameW,frameH),true);
 
                 doImgParamsExtraction=false;


pattacini (commented):
Honestly, I would rather avoid keeping this change in `yarp-3.4` and then in `master`, as OpenCV 2.x.x is very old.
Instead, why don't you edit the base branch and propose merging the PR where you need it, that is `yarp-3.3`?

pattacini:
Along the same line, we may consider getting rid of https://github.com/robotology/yarp/blob/master/src/yarpdatadumper/main.cpp#L32.

cc @Nicogene

PeterBowman:
> Instead, why don't you edit the base branch and propose merging the PR where you need it, that is `yarp-3.3`?

I actually want it in `yarp-3.4` :). I proposed a backport to `yarp-3.3` in order to avoid a gap between patch releases (YARP 3.3 used to support OpenCV 2.x and then, boom). However, it should not a problem for me to stick with YARP 3.3 for a while as we'll eventually migrate to Ubuntu 18.04/20.04 sooner or later. If that's not possible, I can live with this local patch being unmerged until then.

In case the `yarp-3.3` branch is targeted, isn't it supposed to meld into `master` anyway? I don't see any unmerged release-related (`yarp-3.x`) branches in the commit history. If I'm correct, a commit reversion would follow afterwards.

pattacini:
As long as this patch won't be merged in `master` I'm fine with it.

I was proposing `yarp-3.3` as the base branch right for this. I think that `yarp-3.3` can grow without being merged into `master` later on since it is not the latest release we have, unless we wanna patch it with fixes we'd need to merge anyway.
@drdanz correct me if I'm wrong.

Differently, `yarp-3.4` is likely going to accept fixes these days that definitely need to be merged in `master` as it is the latest version, hence pushing this patch into `yarp-3.4` will impact `master`.

We also need to be careful because of ongoing #2358, with which this PR will generate conflicts.

PeterBowman:
I have rebased against `yarp-3.3` and changed base. 

> If that's not possible, I can live with this local patch being unmerged until then.

I'm a bit worried about altering the way branching and releases are being administered because of this tiny change, therefore let me stress again I'm fine with a rejection in view of those potential conflicts in master. This PR has been also devised as a means to catch attention on the OpenCV 2.x regression, its general state and compatibility in current YARP, and the broken unit testing (I assume it used to build against OpenCV 2.x in the past).

drdanz:
My proposal is to merge this in the 3.3 and 3.4 branches, since officially we never dropped support for opencv 2, and revert this change in master, and officially drop support for opencv 2 in YARP 3.5.
I'm not planning to have a new release of the 3.3 branch, @PeterBowman is the branch enough for you or do you need a tag?

drdanz:
Actually, we no longer accept pull requests for the yarp-3.3 branch (the labels are missing, there is no folder for the changelog, and the branch is no longer merged automatically), therefore, I'd rather have this in the yarp-3.4 branch, and cherry-pick it later in the yarp-3.3 branch.
@PeterBowman do you mind rebasing this back onto the yarp-3.4 branch?
Also could you please add the release notes files in the `doc/releases/yarp-3.4` folder?

PeterBowman:
> @PeterBowman do you mind rebasing this back onto the yarp-3.4 branch?
> Also could you please add the release notes files in the `doc/releases/yarp-3.4` folder?

Sure, it's all done now.

> @PeterBowman is the branch enough for you or do you need a tag?

The branch is enough, thanks!

drdanz (approved):
LGTM. Thanks

drdanz:
Merged and cherry-picked in the YARP 3.3 branch (see 38fddc046f8573b7d03cdec3d48f869f2a142bf0). Thanks

PR #2368, opened 2020-09-22
Small refactor of the RGBDSensorClient device

### LINKED ISSUE #2349: Side effects of copying an empty/invalid yarp::sig::Image
I have observed neither base `yarp::sig::Image` nor its derived `yarp::sig::FlexImage` can be safely copied prior to populating them with pixels. In short, all boils down to:

```c++
yarp::sig::FlexImage image1, image2;
image2 = image1; // prints "*** Trying to allocate an invalid pixel type image" and quits
```

This is caused by the following lines (note the somewhat draconian `std::exit` call) and the fact that `getPixelCode` equals `VOCAB_PIXEL_INVALID`:

https://github.com/robotology/yarp/blob/b65ceeb909cd48ca7e9b68b10652c9ac827c1388/src/libYARP_sig/src/yarp/sig/Image.cpp#L349-L353

In a more realistic scenario, this bug exploded in my face while attempting a remote connection to an RGBD sensor. Steps:

1. Launch a [*RGBDSensorWrapper*](http://www.yarp.it/classRGBDSensorWrapper.html) with a wrapped subdevice of your choice (e.g. *depthCamera* or *realsense2*).
   * Don't use [*fakeDepthCamera*](http://www.yarp.it/classfakeDepthCameraDriver.html) as it sets `VOCAB_PIXEL_RGB` internally ([ref](http://www.yarp.it/classfakeDepthCameraDriver.html#a61042040605a84b66786afdce3e0a507)).
   * Please note RGD frames are represented by `yarp::sig::FlexImage`, whereas depth frames use `yarp::sig::ImageOf<yarp::sig::PixelFloat>`.
1. Start a local [*RGBDSensorClient*](http://www.yarp.it/classRGBDSensorClient.html), configure it so that it connects to the server's RGB and depth ports.
1. Retrieve both RGB and depth frames using [`yarp::dev::IRGBDSensor`'s](http://www.yarp.it/classyarp_1_1dev_1_1IRGBDSensor.html) `getRgbImage` and `getDepthImage`, respectively.

**Unless you place a small delay between steps 2. and 3., `getRgbImage` will force the application to quit**, showing that same invalid pixel type error. On the other hand, `getDepthImage` is fine. This is caused by `FlexImage` not having its internal pixel code initialized, as demonstrated earlier. Conversely, `ImageOf<T>` always knows how to interpret its `T` pixel type ([ref](https://github.com/robotology/yarp/blob/b65ceeb909cd48ca7e9b68b10652c9ac827c1388/src/libYARP_sig/src/yarp/sig/Image.h#L676-L679)), no matter if it is storing a frame or not yet.

In view of this, all *RGBDSensorClient* does is to periodically receive streamed RGB+D frames from the server device, store them locally in their respective class members, and make them available on request via `getRgbImage` and `getDepthImage`. In this last stage, the copy assignment operator is invoked (same as in my first snippet, see above). Before the first RGB frame arrives, `last_rgb` does not know what pixel type to expect and `getPixelCode()` returns `VOCAB_PIXEL_UNKNOWN`, causing the early exit. Ref.: [*RGBDSensorClient_StreamingMsgParser.cpp*](https://github.com/robotology/yarp/blob/b65ceeb909cd48ca7e9b68b10652c9ac827c1388/src/devices/RGBDSensorClient/RGBDSensorClient_StreamingMsgParser.cpp)

I propose to drop the draconian `std::exit` call and allow cloning empty/invalid `FlexImage` instances. This is not a problem with `ImageOf<T>` just because of the pixel code being known beforehand. Perhaps a more thorough review is to be considered due to:

https://github.com/robotology/yarp/blob/b65ceeb909cd48ca7e9b68b10652c9ac827c1388/src/libYARP_sig/src/yarp/sig/Image.cpp#L11

Config: Ubuntu 16.04/18.04, GCC 7.5.0, YARP 3.4 (3.4.0+8-20200715.210+gitb14899f).

drdanz:
#### `RGBDSensorClient`

* The device no longer crashes when no image was received yet (#2349).
* The timestamp is now correct.

----

Fixes #2349

### DIFF
--- a/src/devices/RGBDSensorClient/RGBDSensorClient_StreamingMsgParser.cpp
+++ b/src/devices/RGBDSensorClient/RGBDSensorClient_StreamingMsgParser.cpp
@@ -46,11 +46,6 @@ yarp::sig::ImageOf<yarp::sig::PixelFloat> FloatImageReader_Impl::getImage()
 
 
 // Streaming handler
-RGBDSensor_StreamingMsgParser::RGBDSensor_StreamingMsgParser() :
-    port_rgb(nullptr),
-    port_depth(nullptr)
-{}
-
 bool RGBDSensor_StreamingMsgParser::readRgb(yarp::sig::FlexImage &data, yarp::os::Stamp *timeStamp)
 {
     data = read_rgb.getImage();
--- a/src/devices/RGBDSensorClient/RGBDSensorClient_StreamingMsgParser.cpp
+++ b/src/devices/RGBDSensorClient/RGBDSensorClient_StreamingMsgParser.cpp
@@ -17,11 +17,13 @@ RgbImageReader_Impl::~RgbImageReader_Impl() = default;
 
 void RgbImageReader_Impl::onRead(yarp::sig::FlexImage& datum)
 {
+    std::lock_guard<std::mutex> lock(mutex);
     last_rgb = datum;
 }
 
 yarp::sig::FlexImage RgbImageReader_Impl::getImage()
 {
+    std::lock_guard<std::mutex> lock(mutex);
     return last_rgb;
 }
 
@@ -32,11 +34,13 @@ FloatImageReader_Impl::~FloatImageReader_Impl() = default;
 
 void FloatImageReader_Impl::onRead(yarp::sig::ImageOf< yarp::sig::PixelFloat> & datum)
 {
+    std::lock_guard<std::mutex> lock(mutex);
     last_depth = datum;
 }
 
 yarp::sig::ImageOf<yarp::sig::PixelFloat> FloatImageReader_Impl::getImage()
 {
+    std::lock_guard<std::mutex> lock(mutex);
     return last_depth;
 }
 
--- a/src/devices/RGBDSensorClient/RGBDSensorClient_StreamingMsgParser.cpp
+++ b/src/devices/RGBDSensorClient/RGBDSensorClient_StreamingMsgParser.cpp
@@ -18,7 +18,7 @@ RgbImageReader_Impl::~RgbImageReader_Impl() = default;
 void RgbImageReader_Impl::onRead(yarp::sig::FlexImage& datum)
 {
     std::lock_guard<std::mutex> lock(mutex);
-    last_rgb = datum;
+    std::swap(datum, last_rgb);
 }
 
 yarp::sig::FlexImage RgbImageReader_Impl::getImage()
@@ -35,7 +35,7 @@ FloatImageReader_Impl::~FloatImageReader_Impl() = default;
 void FloatImageReader_Impl::onRead(yarp::sig::ImageOf< yarp::sig::PixelFloat> & datum)
 {
     std::lock_guard<std::mutex> lock(mutex);
-    last_depth = datum;
+    std::swap(datum, last_depth);
 }
 
 yarp::sig::ImageOf<yarp::sig::PixelFloat> FloatImageReader_Impl::getImage()
--- a/src/devices/RGBDSensorClient/RGBDSensorClient_StreamingMsgParser.cpp
+++ b/src/devices/RGBDSensorClient/RGBDSensorClient_StreamingMsgParser.cpp
@@ -9,66 +9,108 @@
 
 #include "RGBDSensorClient_StreamingMsgParser.h"
 
+#include <yarp/os/Time.h>
+
 using namespace yarp::dev;
 
 // Callback reader for rgb
-RgbImageReader_Impl::RgbImageReader_Impl() = default;
-RgbImageReader_Impl::~RgbImageReader_Impl() = default;
-
-void RgbImageReader_Impl::onRead(yarp::sig::FlexImage& datum)
+void RgbImageBufferedPort::onRead(yarp::sig::FlexImage& datum)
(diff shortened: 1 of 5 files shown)


src/devices/RGBDSensorClient/RGBDSensorClient_StreamingMsgParser.cpp
```diff
@@ -9,67 +9,97 @@
...
 // Callback reader for rgb
-RgbImageReader_Impl::RgbImageReader_Impl() = default;
-RgbImageReader_Impl::~RgbImageReader_Impl() = default;
-
-void RgbImageReader_Impl::onRead(yarp::sig::FlexImage& datum)
+void RgbImageBufferedPort::onRead(yarp::sig::FlexImage& datum)
 {
-    last_rgb = datum;
+    local_arrival_time = yarp::os::Time::now();
+    std::lock_guard<std::mutex> lock(mutex);
```
PeterBowman, line 20: The lock is acquired after modifying `local_arrival_time`. Is this intentional? Note that in `getImage` this variable is accessed within the critical section, hence my doubt.

src/devices/RGBDSensorClient/RGBDSensorClient_StreamingMsgParser.cpp
```diff
@@ -9,67 +9,97 @@
...
 // callback reader for depthImage
-FloatImageReader_Impl::FloatImageReader_Impl() = default;
-FloatImageReader_Impl::~FloatImageReader_Impl() = default;
-
-void FloatImageReader_Impl::onRead(yarp::sig::ImageOf< yarp::sig::PixelFloat> & datum)
+void FloatImageBufferedPort::onRead(yarp::sig::ImageOf< yarp::sig::PixelFloat> & datum)
 {
-    last_depth = datum;
+    local_arrival_time = yarp::os::Time::now();
+    std::lock_guard<std::mutex> lock(mutex);
```
PeterBowman, line 40: Same here (first `local_arrival_time`, then mutex lock).

src/devices/RGBDSensorClient/RGBDSensorClient_StreamingMsgParser.cpp
```diff
@@ -9,67 +9,97 @@
...
-yarp::sig::ImageOf<yarp::sig::PixelFloat> FloatImageReader_Impl::getImage()
+std::tuple<bool, yarp::sig::ImageOf<yarp::sig::PixelFloat>, yarp::os::Stamp> FloatImageBufferedPort::getImage() const
 {
-    return last_depth;
+    std::lock_guard<std::mutex> lock(mutex);
+    if (local_arrival_time <= 0.0) {
+        // No image received yet
+        return std::make_tuple(false, yarp::sig::ImageOf<yarp::sig::PixelFloat>(), yarp::os::Stamp());
+    }
+    return std::make_tuple(true, last_depth, stamp);
```
PeterBowman, line 52: I believe here (and in other places like this) you should be allowed to reduce it to:

```c++
return {true, last_depth, stamp};
```

PeterBowman (commented):
Suggestion: `RGBDSensor_StreamingMsgParser::read` is not called anywhere, but it could be of use in `RGBDSensorClient::getImages` as opposed to calling both `readRgb` and `readDepth` separately.

src/devices/RGBDSensorClient/RGBDSensorClient_StreamingMsgParser.cpp
PeterBowman: You might want to disregard this comment... There is something fishy around `std::tuple`, look for "conditionally-explicit constructors" in https://en.cppreference.com/w/cpp/utility/tuple/tuple. According to https://stackoverflow.com/a/16713413, this was a compiler extension back in 2013 or so, but later (partially?) fixed in C++11 via [N4387](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4387.html).

drdanz: Yes, unfortunately N4387 is in C++17 and we still require c++14.
It's probably worth add a FIXME comment to remember changing it when we have the compiler support

Nicogene (approved):

Nicogene (commented):
The target branch should be `yarp-3.4` ? Since it is a bugifix

drdanz:
> The target branch should be `yarp-3.4` ? Since it is a bugifix

This device was modified on `master`, and it would cause conflicts, therefore I'm pushing this commit to master.

drdanz:
Fixed all suggestions (thanks @PeterBowman)

> Suggestion: `RGBDSensor_StreamingMsgParser::read` is not called anywhere, but it could be of use in `RGBDSensorClient::getImages` as opposed to calling both `readRgb` and `readDepth` separately.

While fixing this, I noticed that the timestamp returned was still generated by the client, therefore I fixed it as well.

If you are interested in having this in the YARP 3.4 branch, I can try to cherry pick the commits after merging this

PeterBowman:
For me it's perfectly fine to leave these commits in master, many thanks for fixing this :).

PR #2376, opened 2020-10-07
os/Log: Print colored components when colors are enabled

drdanz:
This isn't actually a bugfix, but it is a tiny enhancement that helps **a lot** while debugging, therefore, unless someone is against this, I'd like to merge it in the stable branch.

image

(With `YARP_COMPACT_OUTPUT`):
image

### DIFF
--- a/src/libYARP_os/src/yarp/os/Log.cpp
+++ b/src/libYARP_os/src/yarp/os/Log.cpp
@@ -281,6 +281,51 @@ inline const char* logTypeToBgColor(yarp::os::Log::LogType t)
     }
 }
 
+inline const char* compNameToColor(const char* comp_name)
+{
+    if (!comp_name || comp_name[0] == '\0' || !yarp::os::impl::LogPrivate::colored_output.load()) {
+        return "";
+    }
+
+    // Hashing the component for every log line is probably not optimal, but in
+    // using a hash table will have the hash + a search, and since the
+    // LogCallback has only the name of the component, we cannot store the color
+    // in the LogComponent without breaking the API, therefore this is the less
+    // invasive option.
+    // Anyway, this is enabled only when YARP_COLORED_OUTPUT is set, therefore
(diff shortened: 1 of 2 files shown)


Nicogene:
`YARP_COMPACT_OUTPUT` by default is `false`?

drdanz:
> `YARP_COMPACT_OUTPUT` by default is `false`?

Yes, as it was before...

src/libYARP_os/src/yarp/os/Log.cpp
```diff
@@ -281,6 +281,51 @@ inline const char* logTypeToBgColor(yarp::os::Log::LogType t)
...
+    // LogCallback has only the name of the component, we cannot store the color
+    // in the LogComponent without breaking the API, therefore this is the less
+    // invasive option.
+    // Anyway, this is enabled only when YARP_COLORED_OUTPUT is set, therefore
+    // it can be easily disabled if this slows down the execution.
+    static std::hash<std::string> hsh; // FIXME C++17: Use string_view
+    std::size_t comp_hash = hsh(comp_name) % 12;
+    switch (comp_hash) {
+    case 0:
+        return RED;
```
Nicogene, line 301: I know that it is not part of this PR but isn't better to use `constexpr char[]` instead of defines?

Nicogene (commented):
I have not a strong opinion on this, personally I dislike dark font on a dark background, and this may happen for some component, maybe we can use only bright colors?

drdanz:
> I have not a strong opinion on this, personally I dislike dark font on a dark background, and this may happen for some component, maybe we can use only bright colors?

At the moment it is using all the colors available on your terminal, but if you don't like them, you can usually modify your theme from your terminal settings... I don't think that reducing the number of colors to 7 for this reason is worth...
Also you usually have to read the debug line, not the component name, the color just helps following the execution of your program...

PR #2381, opened 2020-10-11
Minor fix of some names in GazeControl signatures

pattacini:
Actually, `od` stands for desired orientation (as per the Cartesian Control convention for example), whereas here we deal with feedback rather than setpoints, hence `o` is a better replacement.

I don't believe this tiny change is worth an entry point in the `doc/release` as signatures per se remain unchanged.


drdanz (approved):

drdanz:
Merged, thanks.

PR #2390, opened 2020-10-27
Fix ffmpeg get audio

elandini84:
* Corrected a minor error in `FfmpegGrabber.cpp` that caused a segmentation fault when using a video with audio track as a `source` for the device

* **N.B.** The device can now be used and do not crash but the actual audio grabbing has not been tested yet.


drdanz:
Merged, thanks.

PR #2403, opened 2020-11-11
tests: Fix ModifyingCarrier tests

### LINKED ISSUE #2392: ModifyingCarrier::modifyReply is not called (randomly)
As shown by the unit tests that are currently failing, sometimes the `modifyReply` method is not called as it should.
For now I'm going to disable the failing tests, but this is a serious issue for the `ModifyingCarrier` and `portmonitor` usability.

### LINKED ISSUE #2395: Valgrind failures on tests with ModifyingCarrier
There is some other race condition when using ModifyingCarrier (see also #2379) and valgrind complains about several "invalid read" and "invalid write".

This happens randomly in all the tests added in #2382 and #2386.

drdanz:
When a port uses `enableBackgroundWrite`, block on a condition variable and use the callback from the write to wake up the process, so that we can be sure that the write-read-reply process is actually over.

Fixes #2392
Fixes #2395

### DIFF
--- a/tests/libYARP_os/PortTest.cpp
+++ b/tests/libYARP_os/PortTest.cpp
@@ -36,6 +36,9 @@
 
 #include <yarp/companion/impl/Companion.h>
 
+#include <mutex>
+#include <condition_variable>
+
 #include <catch.hpp>
 #include <harness.h>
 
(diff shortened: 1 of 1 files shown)


drdanz:
Travis builds are currently blocked. This is the relative build on my personal account:
https://travis-ci.com/github/drdanz/yarp/builds/199955027

drdanz:
Travis tests on my personal account was successful.

PR #2405, opened 2020-11-11
os/impl/PortCore: Refactor state semaphore

drdanz:
Replaced by a mutex + a condition_variable + a few atomic variables.

The semaphore was used both for synchronization and for protecting critical sections. It is now clear where the threads are syncronizing, and where the mutex is protecting some data from concurrent access.

(note: better select "Hide whitespace changes" while reviewing the patch)

### DIFF
--- a/src/libYARP_os/src/yarp/os/Contactable.h
+++ b/src/libYARP_os/src/yarp/os/Contactable.h
@@ -280,6 +280,7 @@ public:
      * @param readOnly set this if you won't be modifying the properties.
      * @return the port properties (or nullptr if readOnly and none have
      *         been set)
+     * @warning Must be called in the same thread as releaseProperties
      */
     virtual Property* acquireProperties(bool readOnly) = 0;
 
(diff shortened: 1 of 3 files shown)


drdanz:
Apparently this introduces a race condition that causes a deadlock every once in a while :sob:

Nicogene:
> Apparently this introduces a race condition that causes a deadlock every once in a while 😭

This the type of PR that introduce regressions for years 😆

drdanz:
> This the type of PR that introduce regressions for years laughing

But still necessary, if we want to be able to fully understand what this code is doing...

Anyway, I fixed the regression in the latest version. I thought that atomic variables would save us from locking the mutex in a few places, but I was wrong, in a few cases two or more of these variables must be changed at the same time atomically, and that requires a mutex (it would probably be possible with an atomic bitset or something similar, but that's way beyond the goal of this patch).
Now, if the continuous integration wasn't completely broken...

PR #2408, opened 2020-11-13
grabber rename

randaz81:
Some grabber files renamed for consistency. No functional changes.


Nicogene (commented):
Ok with that, then the `ServerGrabber` `ServerGrabberResponder` classes remain unchanged?

drdanz (approved):

drdanz:
> Ok with that, then the `ServerGrabber` `ServerGrabberResponder` classes remain unchanged?

I don't have an opinion on that... I think the problem came from the fact that the folder name was really confusing.
Feel free to rename them as well, if you want, or to merge the patch.

drdanz:
Merged, thanks.

PR #2417, opened 2020-11-21
Print RGBA values as ints in PointCloud data types

PeterBowman:
For instance, `printf("%c", red);` outputs the corresponding ASCII character. We should get numeric values instead.


drdanz (approved):
Good catch! Thanks for the patch!

drdanz:
Merged, thanks.

PR #2436, opened 2020-12-18
Yarp trafficgen

randaz81:
The companion tool `yarp trafficgen` sends generated data over a yarp port. The tool allows to test the communication and the available bandwidth over the network.
E.g.
```
yarp trafficgen /port --bandwidth [Mb/s]
yarp trafficgen /port --bandwidth [Mb/s] --duration [s]
yarp trafficgen /port --period [s] --size [MB]
yarp trafficgen /port --period [s] --size [MB] --duration [s]
```

### DIFF
--- a/src/libYARP_companion/src/yarp/companion/impl/Companion.cmdTrafficGen.cpp
+++ b/src/libYARP_companion/src/yarp/companion/impl/Companion.cmdTrafficGen.cpp
@@ -1,0 +1,97 @@
+/*
+ * Copyright (C) 2006-2020 Istituto Italiano di Tecnologia (IIT)
+ * All rights reserved.
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD-3-Clause license. See the accompanying LICENSE file for details.
+ */
+
+#include <yarp/companion/impl/Companion.h>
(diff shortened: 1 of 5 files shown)


drdanz (changes requested):
Please use a lowercase name for the command like all the other tools...
I suggest yarp trafficgen

src/libYARP_companion/src/yarp/companion/impl/Companion.cmdTrafficGen.cpp
```diff
@@ -0,0 +1,131 @@
...
+    char* data_buff =nullptr;
+    std::string m_portname;
+    yarp::os::Port outport;
+    yarp::os::Bottle pp;
+
+public:
+    datasender(double period, string portname, double data_size_MB) : PeriodicThread (period)
+    {
+        m_portname=portname;
+        m_data_size_MB = data_size_MB;
```
drdanz, line 38: This should be done using the member initialization instead of assignment in the constructor

drdanz (changes requested):

drdanz (approved):
LGTM

PR #2439, opened 2020-12-21
fixed buttons of yarpmobilebasegui

randaz81:
Fixed gui buttons when the value of max linear/angular velocity is < 0


drdanz (approved):

drdanz:
Merged, thanks.

PR #2440, opened 2020-12-22
Properly set dependencies for the  libraries

### LINKED ISSUE #2438: PCL usage requirements are not properly propagated by YARP_pcl
**Background:**
I have been recently working on a YARP wrapper around several PCL cloud processing and surface reconstruction algorithms ([library](https://github.com/roboticslab-uc3m/vision/tree/67df1286484e8566c238fe750a522a6f4b41bce5/libraries/YarpCloudUtils), [rationale](https://github.com/roboticslab-uc3m/vision/issues/102)). The YARP_pcl component had come in useful in terms of YARP2PCL and PCL2YARP type conversions, although not devoid of some shortcomings on the CMake side of life.

**Minimal not working example:**

```cmake
# CMakeLists.txt
cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
project(yarp-pcl LANGUAGES CXX)
find_package(YARP 3.4 REQUIRED pcl)
add_executable(${PROJECT_NAME} main.cpp)
target_link_libraries(${PROJECT_NAME} YARP::YARP_pcl)
```

```cxx
// main.cpp
#include <yarp/pcl/Pcl.h>
auto main() -> int {}
```

```
In file included from /home/bartek/Desktop/temp/yarp-pcl/main.cpp:1:
/usr/local/include/yarp/pcl/Pcl.h:12:10: fatal error: pcl/io/pcd_io.h: No such file or directory
   12 | #include <pcl/io/pcd_io.h>
      |          ^~~~~~~~~~~~~~~~~
```

**Diagnosis:**
As expected, YARP_pcl is instructed to propagate PCL usage requirements to downstreams:

https://github.com/robotology/yarp/blob/0481f994c6e03897d038c5f1d1078145646a1772/src/libYARP_pcl/src/CMakeLists.txt#L23-L25

However, the `${PCL_INCLUDE_DIR}` and `${PCL_LIBRARIES}` variables are **empty**. These have not been populated since YARP does not perform a `find_package(PCL)` call anywhere. As a result, downstream projects are instructed to search for PCL on the system (thanks to the third instruction), but neither PCL targets/libraries nor PCL include directories are automatically propagated via `target_link_libraries(... YARP::YARP_pcl)` as they should be, thus resulting in the missing header inclusion error.

**Remark regarding YARP_cv:**
I believe this component was developed in a similar vein to YARP_cv. Please note **YARP_cv does not export OpenCV's usage requirements either**. Even though I haven't tested it thoroughly, I believe this same issue prevails after inspecting the generated *YARP_cvTargets.cmake* file. In that case, however, I tracked down the problem to the fact that there is no such thing as `OPENCV_INCLUDE_DIR` ([ref](https://github.com/robotology/yarp/blob/0481f994c6e03897d038c5f1d1078145646a1772/src/libYARP_cv/src/CMakeLists.txt#L24-L25)) on my OpenCV 4.4 installation, it should read `OpenCV_INCLUDE_DIRS` instead or, ideally, favor modern CMake targets.

Besides, YARP actually looks for OpenCV as there exists some CV-dependent code such as the *opencv_grabber* device.

**Additional suggestions:**
There is more room for improvements beyond fixing the `find_package` thing:
* In [yarp/pcl/Pcl.h](https://github.com/robotology/yarp/blob/0481f994c6e03897d038c5f1d1078145646a1772/src/libYARP_pcl/src/yarp/pcl/Pcl.h), I'd add these header inclusions:
  ```cxx
  #include <cstring> // std::memcpy, also mind prepending std:: to the existing call
  #include <pcl/point_cloud.h> // pcl::PointCloud<T>
  ```
* The following line in [CMakeLists.txt](https://github.com/robotology/yarp/blob/0481f994c6e03897d038c5f1d1078145646a1772/src/libYARP_pcl/src/CMakeLists.txt):
  ```cmake
  list(APPEND YARP_pcl_PUBLIC_DEPS PCL)
  ```
  is better written as:
  ```cmake
  list(APPEND YARP_pcl_PUBLIC_DEPS "PCL COMPONENTS common io")
  ```
  In this way, downstreams only need to search for specific PCL components as requested by YARP_pcl. Otherwise, *all* available PCL components and their related variables are needlessly populated and brought to scope.
* Care should be taken to check this against different PCL versions, but on PCL 1.10 I can confirm `PCL_INCLUDE_DIR` is empty. It should read `PCL_INCLUDE_DIRS` instead.

**Tested configuration:**
* YARP version 3.4.1+81-20201214.6+git1b0060380
* PCL 1.10 (via `apt install libpcl-dev`)
* cmake version 3.16.3
* g++ (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
* Ubuntu 20.04.1 (focal), Linux 5.4.0-58-generic

drdanz:
## Build System

* Dependencies for the `INTERFACE` libraries (YARP_cv and YARP_pcl)
  are now properly set (#2438).
* PCL is now required to build the YARP_pcl library.
* OpenCV is now required to build the YARP_cv library.

Fixes #2438

### DIFF
--- a/src/libYARP_cv/src/CMakeLists.txt
+++ b/src/libYARP_cv/src/CMakeLists.txt
@@ -21,8 +21,8 @@ target_compile_features(YARP_conf INTERFACE cxx_std_14)
 target_link_libraries(YARP_cv INTERFACE YARP::YARP_sig)
 list(APPEND YARP_cv_PUBLIC_DEPS YARP_sig)
 
-target_include_directories(YARP_cv SYSTEM INTERFACE ${OPENCV_INCLUDE_DIR})
-target_link_libraries(YARP_cv INTERFACE ${OPENCV_INCLUDE_DIR})
+target_include_directories(YARP_cv SYSTEM INTERFACE ${OpenCV_INCLUDE_DIR})
+target_link_libraries(YARP_cv INTERFACE ${OpenCV_LIBRARIES})
 list(APPEND YARP_cv_PUBLIC_DEPS OpenCV)
 
 # set_property(TARGET YARP_cv PROPERTY PUBLIC_HEADER ${YARP_cv_HDRS})
--- a/src/libYARP_pcl/src/CMakeLists.txt
+++ b/src/libYARP_pcl/src/CMakeLists.txt
@@ -20,9 +20,11 @@ target_compile_features(YARP_pcl INTERFACE cxx_std_14)
 target_link_libraries(YARP_pcl INTERFACE YARP::YARP_sig)
 list(APPEND YARP_pcl_PUBLIC_DEPS YARP_sig)
 
-target_include_directories(YARP_pcl SYSTEM INTERFACE ${PCL_INCLUDE_DIR})
-target_link_libraries(YARP_pcl INTERFACE ${PCL_LIBRARIES})
-list(APPEND YARP_pcl_PUBLIC_DEPS PCL)
+target_include_directories(YARP_pcl SYSTEM INTERFACE ${PCL_COMMON_INCLUDE_DIR}
+                                                     ${PCL_IO_INCLUDE_DIR})
+target_link_libraries(YARP_pcl INTERFACE ${PCL_COMMON_LIBRARIES}
+                                         ${PCL_IO_LIBRARIES})
+list(APPEND YARP_pcl_PUBLIC_DEPS "PCL COMPONENTS common io")
 
 # set_property(TARGET YARP_pcl PROPERTY PUBLIC_HEADER ${YARP_pcl_HDRS})
 # set_property(TARGET YARP_pcl PROPERTY PRIVATE_HEADER ${YARP_pcl_IMPL_HDRS})
--- a/cmake/YarpFindDependencies.cmake
+++ b/cmake/YarpFindDependencies.cmake
@@ -334,6 +334,9 @@ checkandset_dependency(OpenCV)
 set_property(DIRECTORY PROPERTY INCLUDE_DIRECTORIES ${OpenCV24_WORKAROUND_ORIGINAL_INCLUDE_DIRS})
 # End second part of workaround
 
+find_package(PCL COMPONENTS common io QUIET)
+checkandset_dependency(PCL)
+
 find_package(Lua QUIET)
 checkandset_dependency(Lua)
 
@@ -608,6 +611,7 @@ print_dependency(QGVCore)
 print_dependency(Libedit)
 print_dependency(SWIG)
 print_dependency(OpenCV)
+print_dependency(PCL)
 print_dependency(Lua)
 print_dependency(LibOVR)
 print_dependency(GLFW3)
--- a/doc/release/yarp_3_4/fix_interface_lib_deps.md
+++ b/doc/release/yarp_3_4/fix_interface_lib_deps.md
@@ -1,0 +1,9 @@
+fix_interface_lib_deps {#yarp_3_4}
+----------------------
+
+## Build System
+
+* Dependencies for the `INTERFACE` libraries (YARP_cv, YARP_pcl)
+  are now properly set (#2438).
+* PCL is now required to build the YARP_pcl library.
+* OpenCV is now required to build the YARP_cv library.
--- a/src/libYARP_cv/CMakeLists.txt
+++ b/src/libYARP_cv/CMakeLists.txt
@@ -6,6 +6,13 @@
 
 project(YARP_cv)
 
+# OpenCV is required, otherwise the OpenCV_INCLUDE_DIR and OpenCV_LIBRARIES
+# (used in the CMakeLists.txt file in the src direcrory) are not expanded).
+# FIXME Remove this check when OpenCV targets will be used.
+if (NOT YARP_HAS_OpenCV)
+  return()
+endif()
+
 add_subdirectory(src)
 
 include(YarpInstallBasicPackageFiles)
--- a/src/libYARP_pcl/CMakeLists.txt
+++ b/src/libYARP_pcl/CMakeLists.txt
@@ -6,6 +6,14 @@
 
 project(YARP_pcl)
 
+# PCL is required, otherwise the PCL_<component>_INCLUDE_DIR and
+# PCL_<component>_LIBRARIES
+# (used in the CMakeLists.txt file in the src direcrory) are not expanded).
+# FIXME Remove this check when PCL targets will be used.
+if (NOT YARP_HAS_PCL)
+  return()
+endif()
+
 add_subdirectory(src)
 
 include(YarpInstallBasicPackageFiles)


PeterBowman:
Thank you for the patch, @drdanz. So far I have only checked and tested YARP_pcl and I confirm it works, but let me add a few remarks:

* I don't think any of these components, especially YARP_pcl, should be disabled if their external dependency (PCL, CV, Eigen) is not found. First of all, it's not harmful at all since those are header-only libraries. Secondly, projects such as mine (https://github.com/robotology/yarp/issues/2438) won't get into trouble anyway thanks to the upstream `find_dependency(PCL)` call propagated to the dowsntream `find_package(YARP_pcl)`: no PCL found on my system = I don't even care if YARP was built with or without PCL support ("built" is a big word anyway, no YARP_pcl binaries are generated). I presume this is mostly relevant to packaged YARP distributions. I also presume you don't prepare your Windows YARP installer to fetch/bundle PCL binaries, which are quite heavy (this beast does even transitively depend on VTK and Boost!). In other words, if you do prepare a YARP distribution with PCL support, it's probably to be expected that YARP installer also installs PCL (+ VTK, Boost...) on the target system. I'd rather pick a lighter YARP installer and check whether any of my projects need to depend on YARP_pcl+PCL, then download and install PCL by myself.
* As you have noticed, `find_package(PCL)` performs a search of all known PCL components and populates a rather long list of targets with their transitive dependencies, including VTK+Boost+FLANN et al., consequently propagated to `INTERFACE_LINK_LIBRARIES` (https://github.com/robotology/yarp/issues/2438#issuecomment-749408873). YARP_pcl only needs pcl_common, pcl_io and their respective dependencies, which can be translated in the following manner:
  ```cmake
  find_package(PCL COMPONENTS common io QUIET) # in YarpFindDependencies.cmake
  list(APPEND YARP_pcl_PUBLIC_DEPS "PCL COMPONENTS common io") # in src/libYARP_pcl/src/CMakeLists.txt
  ```
* This line is buggy:
  https://github.com/robotology/yarp/blob/0481f994c6e03897d038c5f1d1078145646a1772/src/libYARP_pcl/src/CMakeLists.txt#L23
  On PCL 1.9+, you don't need it as PCL started using CMake targets and include directories are thus propagated via `target_link_libraries`. You do have to call this on PCL 1.8 (Ubuntu bionic) and earlier, but note that it should read `${PCL_INCLUDE_DIRS}` instead of `${PCL_INCLUDE_DIR}`. In fact, in the code you pasted at https://github.com/robotology/yarp/issues/2438#issuecomment-749408873 there should be PCL-related stuff in `INTERFACE_INCLUDE_DIRECTORIES`.

PeterBowman:
For the record, PCL 1.8 and 1.10 register these (inter)dependencies (from `PCLConfig.cmake`):
* Internally: pcl_io => pcl_common, pcl_octree (=> pcl_common)
* Externally:
  * pcl_common => eigen, boost
  * pcl_io => openni, openni2, pcap, png, vtk, libusb-1.0; additionally PCL 1.8 depends on: ensenso, davidSDK, dssdk

drdanz:
@PeterBowman Hi, I'm about to do a minor release, and I'm not sure what to do about this patch, I agree that these components should not be disabled if the libraries are not found, but we do have to support bionic in the 3.4 branch, and in bionic the pcl_common and pcl_io targets do not exist, and therefore the only way is to the variables for now.

I reverted the changes to the YARP_eigen library, since it uses targets, and therefore there are no variables to expand, but both YARP_pcl and YARP_opencv have this issue.

I also changed the dependency for YARP_pcl to "common" and "io" libraries and components only.

I'm not sure if we still have to support bionic for the 3.5 release, but as soon as we drop the support, we can remove this dependency.
We should also check the dependencies for the YARP_cv library.

(Side note: I think that we should also rename the library YARP_opencv for coherence... It's not a "computer vision" library, but a library for compatibility with opencv)

drdanz:
Build [22] failed for reasons that are not related to this patch.
I'm merging this, in order to make the release, @PeterBowman feel free to open a new issue (or even better a PR 😄) if you think that something is wrong with this PR.

src/libYARP_pcl/src/CMakeLists.txt
```diff
@@ -20,9 +20,11 @@ target_compile_features(YARP_pcl INTERFACE cxx_std_14)
 target_link_libraries(YARP_pcl INTERFACE YARP::YARP_sig)
 list(APPEND YARP_pcl_PUBLIC_DEPS YARP_sig)
 
-target_include_directories(YARP_pcl SYSTEM INTERFACE ${PCL_INCLUDE_DIR})
-target_link_libraries(YARP_pcl INTERFACE ${PCL_LIBRARIES})
-list(APPEND YARP_pcl_PUBLIC_DEPS PCL)
+target_include_directories(YARP_pcl SYSTEM INTERFACE ${PCL_COMMON_INCLUDE_DIR}
+                                                     ${PCL_IO_INCLUDE_DIR})
```
PeterBowman, line 24: On PCL 1.10 (using CMake targets), `INTERFACE_INCLUDE_DIRECTORIES` is defined as `"${PCL_${COMPONENT}_INCLUDE_DIRS};${PCL_CONF_INCLUDE_DIR}"`. Not only we are missing the `PCL_CONF_INCLUDE_DIR` part (I don't think this is critical here, anyway), but also note the trailing `S` in the component's include dir variable:
* `${PCL_XXX_INCLUDE_DIR}`: just this component's include header dir(s)
* `${PCL_XXX_INCLUDE_DIRS}`: same as above **plus** dependencies

On PCL 1.8 (Ubuntu bionic), my sample application (see linked issue) throws a compile error when using `PCL_XXX_INCLUDE_DIR` due to an unresolved Eigen dependency.

I think this is better written simply as `${PCL_INCLUDE_DIRS}`, thus accounting for all internal and external dependencies (pcl_io also dependes on pcl_octree).

src/libYARP_pcl/src/CMakeLists.txt
```diff
@@ -20,9 +20,11 @@ target_compile_features(YARP_pcl INTERFACE cxx_std_14)
...
+target_link_libraries(YARP_pcl INTERFACE ${PCL_COMMON_LIBRARIES}
+                                         ${PCL_IO_LIBRARIES})
```
PeterBowman, line 26: Similarly to the above comment, I'd write `${PCL_LIBRARIES}` even if dealing with PCL 1.9+ and CMake targets. The reason is that on PCL 1.10 neither `${PCL_XXX_LIBRARIES}` nor individual targets deal with 3rd party dependencies. So this is a similar issue to the include dirs counterpart, but here we'll probably see errors on link time.

PeterBowman (commented):
Sorry, @drdanz, I was struggling with different scenarios and had not come yet to a complete solution. Dealing with PCL+CMake is a bit cumbersome and I'm afraid adopting modern targets might not be a panacea. This patch is definitely an improvement, though, thank you! I'll try to review this whole issue again later on. I have dropped a few comments for future reference.

src/libYARP_cv/src/CMakeLists.txt
```diff
@@ -21,8 +21,8 @@ target_compile_features(YARP_conf INTERFACE cxx_std_14)
 target_link_libraries(YARP_cv INTERFACE YARP::YARP_sig)
 list(APPEND YARP_cv_PUBLIC_DEPS YARP_sig)
 
-target_include_directories(YARP_cv SYSTEM INTERFACE ${OPENCV_INCLUDE_DIR})
-target_link_libraries(YARP_cv INTERFACE ${OPENCV_INCLUDE_DIR})
+target_include_directories(YARP_cv SYSTEM INTERFACE ${OpenCV_INCLUDE_DIR})
```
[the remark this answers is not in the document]
drdanz: Thanks for catching this. Fixed in 121da79986ddefe384f01c31deb6d5147706a9d3

src/libYARP_pcl/src/CMakeLists.txt
drdanz: reverted, see b0443ce491b6793746eb8fcb0a586b6e2d2f4b1f

src/libYARP_pcl/src/CMakeLists.txt
drdanz: reverted see b0443ce491b6793746eb8fcb0a586b6e2d2f4b1f

drdanz:
I agree that modern targets would be better, but unfortunately I can't do it in the stable branch...
Applied @PeterBowman suggestions, thanks for the feedback.

PR #2448, opened 2021-01-07
Clear map temporary flags

randaz81:
* added method `yarp::dev::Nav2D::MapGrid2D::clearMapTemporaryFlags()`
* added the following methods to `yarp::dev::Nav2D::IMap2D` interface: `clearAllMapsTemporaryFlags()`, `clearMapTemporaryFlags(std::string map_name)`
* `Navigation2DClient` and `Navigation2DServer` updated accordingly.

PR #2449, opened 2021-01-07
Compressed MapGrid2D

randaz81:
#### This pr needs to be rebased as soon as #2448 is merged.

Relevant changes:
### yarp::dev
* added internal lossless compression for datatype `yarp::dev::Nav2D::MapGrid2D` using Zlib: the methods `yarp::os::Portable::write()`
  and `yarp::os::Portable::read()` handle the compression before sending data to the network.
* the method `MapGrid2D::enable_map_compression_over_network()` enables/disables the data compression over the network (default true).
* if Zlib library is not available, compression will be set to false.

### devices
* added rpc option `enable_maps_compression <0/1>` to `yarp::dev::Map2DServer`


src/devices/map2DServer/Map2DServer.cpp
```diff
@@ -714,6 +714,12 @@ void Map2DServer::parse_string_command(yarp::os::Bottle& in, yarp::os::Bottle& o
         m_maps_storage.clear();
         out.addString("all maps cleared");
     }
+    else if (in.get(0).asString() == "enable_maps_compression")
+    {
+        for (auto it=m_maps_storage.begin(); it!= m_maps_storage.end(); it++)
+            {it->second.enable_map_compression_over_network(in.get(1).asBool());}
+        out.addString("compression mode of all maps set to:"+ in.get(1).asString());
```
drdanz, line 721: Please add a warning when compression is not enabled

drdanz (approved):
LGTM, just a comment

src/devices/map2DServer/Map2DServer.cpp
randaz81: warning added

PR #2451, opened 2021-01-07
nested .ini files

randaz81:
`yarp::os::Property` class now implements the new tag `import <contextname> <filename>` which allows to include .ini configuration files belonging to different contexts


randaz81:
This PR fixes: https://github.com/robotology/yarp/issues/565

tests/conf/contexts/context2/c2f2.ini
```diff
@@ -0,0 +1,6 @@
+c2f2p1 15
+c2f2p2 16
+
+[c2f2s1]
+c2f2s1p1 17
+c2f2s1p2 18
```
drdanz, line 6:
```suggestion
c2f2s1p2 18

```

drdanz (approved):
LGTM, but we should spend some more time in investigating the order of include/import, and values replacements

randaz81:
> LGTM, but we should spend some more time in investigating the order of include/import, and values replacements

I put some relevant info here: https://github.com/robotology/yarp/discussions/2453

Issue #2432, opened 2020-12-11
gravityCompensator crashes in the invocation of yarp::sig::Vector::operator=(&&)


Nicogene:
This issue is for discussing the problem we discovered recently on gravityCompensator (https://github.com/robotology/icub-main/pull/702).
 
In few words it crashed **sistematically** at this line: 
 
https://github.com/robotology/icub-main/blob/78be1eff4dfda699690f9a2b4310cee6d7d39bcf/src/modules/gravityCompensator/gravityThread.cpp#L748
 

On that `yarp::sig::Vector` (`exec_torques_RA`),  as the other many `Vector`s in `gravityCompensator`, a `resize(7)` and `zero()` was invoked in the `threadInit()`.

We verified that all those vectors involved in that operation had correct sizes.

We tried then to put the result of that variable in a temporary Vector, and then copy that Vector in "our" `exec_torques_RA` and it worked smoothly. Since the temporary Vector had a name, then the copy assignment operator was invoked instead of the move one.

 Initializing the Vectors with the correct size in the gravityThread constructor(see https://github.com/robotology/icub-main/pull/702/commits/00f904632012eed4b222880f42ea035f47eadd2c) seems to fix the problem.

Under the hood of `yarp::sig::Vector`  there is a `std::vector`, since a while (https://github.com/robotology/yarp/pull/2195) and it is the first time I experience a problem like this

The implementations of copy/move assignment operators seems pretty straightforward and I don't see any criticity:

- Copy: https://github.com/robotology/yarp/blob/master/src/libYARP_sig/src/yarp/sig/Vector.h#L168-L174

- Move: https://github.com/robotology/yarp/blob/master/src/libYARP_sig/src/yarp/sig/Vector.h#L190-L194

It seems strange to me that operator=(&&) di std::vector is bugged.

It seems that doing 

```c++
std::vector<double> v;
v.resize(7);
std::fill(v.begin(), v.end(), 0.0);
```

is different from

```c++
std::vector<double> v(7, 0.0);
```

Maybe it is due to the allocation strategy is implementation-dependent, under some OS(e.g. Linux)  should allocate the memory by power of 2 elements(https://stackoverflow.com/questions/11560904/vector-memory-allocation-strategy)

 
This issue is not mainly focused on `gravityCompensator` (that is in another repo) but on the possibility that we could have a hidden issue in `yarp::sig::Vector`, and since it is widely used, it is quite dangerous.

[...]

PR #2471, opened 2021-01-28
sig/VectorOf: Do not explicitly implement copy/move ctors and assignment operators

drdanz:
Default ones are just fine (rule of zero)

CC: @Nicogene
CC-Issue: #2432

### DIFF
--- a/src/libYARP_sig/src/yarp/sig/Vector.h
+++ b/src/libYARP_sig/src/yarp/sig/Vector.h
@@ -159,41 +159,6 @@ public:
         memcpy(this->data(), p, sizeof(T)*s);
     }
 
-    VectorOf(const VectorOf& r) : VectorBase(), bytes(r.bytes) {
-    }
-
-    /**
-     * Copy operator;
-     */
-    const VectorOf<T> &operator=(const VectorOf<T>& r)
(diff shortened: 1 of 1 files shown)


Nicogene (approved):

drdanz:
The `bindings::ruby::vector` test is failing with this error:
```
<yarp>/bindings/ruby/tests/test_vector.rb:19:in `initialize': Wrong arguments for overloaded method 'Vector.new'. (ArgumentError)
Possible C/C++ prototypes are:
    Vector.new()
    Vector.new(size_t size)
    Vector.new(size_t s, double const &def)
    Vector.new(size_t s, double const *p)
    Vector.new(std::vector< double,std::allocator< double > > const &values)
        from <yarp>/bindings/ruby/tests/test_vector.rb:19:in `new'
        from <yarp>/bindings/ruby/tests/test_vector.rb:19:in `test_vector'
        from <yarp>/bindings/ruby/tests/test_vector.rb:30:in `<main>'
```

Apparently swig is not generating the method for automatically generated ctor and operators...
Perhaps we should declare them `= default` to make swig happy?
Anyway this sounds like a bug in swig

PR #2473, opened 2021-02-01
Avoid targeting too many subdevice joints by ControlBoardWrapper

PeterBowman:
For the sake of describing the problem this PR tries to solve, let me introduce and outline the following terms I've just made up related to families of methods found across several [motor interfaces](http://www.yarp.it/git-master/group__dev__iface__motor.html):

* Single-joint signature: `singleJointSignature(int joint, double value)` targets one joint
* Group signature: `groupSignature(int n_joints, const int* joints, double* values)` targets specific joints
* Whole-wrapper signature: `wholeWrapperSignature(double* values)` targets all wrapped joints

Sometimes, the last signature attempts to target way too much. Our research team has proposed a single whole-body CAN controller device for a humanoid robot that encompasses up to 40 controlled joints distributed over eight network wrappers (legs, arms, hands, head and trunk). In other words, these eight `controlboardwrapper2` instances must share a common subdevice, and none of them will be responsible of exposing more than 6 robot joints over the YARP network.

Here comes trouble: whole-wrapper mappings on the network wrapper side (e.g. `ControlBoardWrapper::xxx(double*)`) usually target whole-wrapper signatures on the subdevice side. That is, even though we just want to address 6 subdevice joints, this mapping calls a method on the subdevice that affects all 40 joints. If this action happens to produce one CAN message per joint, 34 messages are needlessly consuming bandwidth in addition to the 6 we actually require.

**Solution:** map whole-wrapper signatures on the network wrapper side to group signatures on the subdevice side, whenever possible. In this way, we can more accurately target specific joints in the whole subdevice matrix. If there is no group signature available, I propose to resort to iterating over single-joint calls (premise: better just as many times as necessary than once and too much).

**1. Proposed whole-wrapper (network wrapper side) to group (subdevice side) implementation and affected methods:**

```cxx
#include <numeric> // std::iota
#include <vector>

bool wholeWrapperSignature(double* values)
{
    int j_wrap = 0;

    for (size_t subDev_idx = 0; subDev_idx < device.subdevices.size(); subDev_idx++) {
        auto* p = device.getSubdevice(subDev_idx);

        if (p && p->iface) {
            std::vector<int> joints((p->top - p->base) + 1);
            std::iota(joints.begin(), joints.end(), p->base);

            if (!p->iface->groupSignature(joints.size(), joints.data(), &values[j_wrap])) {
                return false;
            }

            j_wrap += joints.size();
        } else {
            return false;
        }
    }

    return true;
}
```

* `getControlModes(int*)` in `IControlMode`
* `getInteractionModes(yarp::dev::InteractionModeEnum*)` in `IInteractionMode`
* `getRefSpeeds(double*)` in `IPositionControl`
* `getRefAccelerations(double*)` in `IPositionControl` and `IVelocityControl`
* `getTargetPositions(double*)` in `IPositionControl`
* `getRefPositions(double*)` in `IPositionDirect`
* `getRefVelocities(double*)` in `IVelocityControl`

**2. Proposed whole-wrapper (network wrapper side) to single-joint (subdevice side) implementation and affected methods:**

```cxx
bool wholeWrapperSignature(double* values)
{
    for (size_t l = 0; l < controlledJoints; l++) {
        int off = device.lut[l].offset;
        size_t subIndex = device.lut[l].deviceEntry;
        auto* p = device.getSubdevice(subIndex);

        if (!p || !p->iface|| !p->iface->singleJointSignature(static_cast<int>(off + p->base), &values[l])) {
            return false;
        }
    }
    
    return true;
}
```

* `getAmpStatus(int*)` in `IAmplifierControl`
* `getCurrents(double*)` in `IAmplifierControl` and `ICurrentControl`
* `getCurrentRanges(double*, double*)` in `ICurrentControl`
* `getRefCurrents(double*)` in `ICurrentControl`
* `getEncoders(double*)` in `IEncodersTimed`
* `getEncoderSpeeds(double*)` in `IEncodersTimed`
* `getEncoderAccelerations(double*)` in `IEncodersTimed`
* `getEncodersTimed(double*, double*)` in `IEncodersTimed`
* `getTemperatures(double*)` in `IMotor`
* `getMotorEncoders(double*)` in `IMotorEncoders`
* `getMotorEncoderSpeeds(double*)` in `IMotorEncoders`
* `getMotorEncoderAccelerations(double*)` in `IMotorEncoders`
* `getMotorEncodersTimed(double*, double*)` in `IMotorEncoders`
* `getPids(const PidControlTypeEnum&, double*)` in `IPidControl`
* `getPidErrors(const PidControlTypeEnum&, double*)` in `IPidControl`
* `getPidErrorLimits(const PidControlTypeEnum&, double*)` in `IPidControl`
* `getPidReferences(const PidControlTypeEnum&, Pid*)` in `IPidControl`
* `getPidOutputs(const PidControlTypeEnum&, double*)` in `IPidControl`
* `getDutyCycles(double*)` in `IPWMControl`
* `getRefDutyCycles(double*)` in `IPWMControl`
* `getRefTorques(double*)` in `ITorqueControl`
* `getTorques(double*)` in `ITorqueControl`
* `getTorqueRanges(double*, double*)` in `ITorqueControl`

**3. Miscellanea, single-joint implementations migrated to group signature**

* `setRefCurrents(const double*)` in `ICurrentControl`
* `setInteractionModes(yarp::dev::InteractionModeEnum*)` in `IInteractionMode`
* `relativeMove(const double*)` in `IPositionControl`
* `setPositions(const double*)` in `IPositionDirect`
* `setRefTorques(const double*)` in `ITorqueControl`
* `stop()` in `IPositionControl` and `IVelocityControl`

There is a bunch of other methods I'm harmonizing according to the proposed implementation, although they still map in the same way they did before. Additional changes:

* Remove superfluous checks in `SubDevice::attach` that made no sense since the deprecation of `IPositionControl2` (https://github.com/robotology/yarp/commit/b8727d4575d7238dce21fca15d8d609f96667713, https://github.com/robotology/yarp/commit/6a1ac2959cef62b18e366b5a7d980fd95656a7d1).
* Remove the now unused `printError` helper method.
* Add Eclipse and VS Code bits to .gitignore.

These changes have been tested with the code provided in [this gist](https://gist.github.com/PeterBowman/5c99113b191c255624d5f540598e6ea3). Includes console output before and after the fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,11 @@ compile_commands.json
 *.aps
 /CMakeSettings.json
 /.vs/
+/.vscode/
 /.idea/
+/.project
+/.cproject
+/.settings/
 /cmake-build*/
 .clangd/
 *.srctrlbm
--- a/src/devices/ControlBoardWrapper/ControlBoardWrapperAmplifierControl.cpp
+++ b/src/devices/ControlBoardWrapper/ControlBoardWrapperAmplifierControl.cpp
@@ -69,28 +69,17 @@ bool ControlBoardWrapperAmplifierControl::disableAmp(int j)
 
 bool ControlBoardWrapperAmplifierControl::getAmpStatus(int* st)
 {
-    int* status = new int[device.maxNumOfJointsInDevices];
-    bool ret = true;
-    for (size_t d = 0; d < device.subdevices.size(); d++) {
-        SubDevice* p = device.getSubdevice(d);
-        if (!p) {
-            ret = false;
-            break;
-        }
+    for (size_t l = 0; l < controlledJoints; l++) {
+        int off = device.lut[l].offset;
+        size_t subIndex = device.lut[l].deviceEntry;
+        auto* p = device.getSubdevice(subIndex);
 
-        if ((p->amp) && (ret = p->amp->getAmpStatus(status))) {
-            for (size_t juser = p->wbase, jdevice = p->base; juser <= p->wtop; juser++, jdevice++) {
-                st[juser] = status[jdevice];
-            }
-        } else {
-            printError("getAmpStatus", p->id, ret);
-            ret = false;
-            break;
+        if (!p || !p->amp || !p->amp->getAmpStatus(static_cast<int>(off + p->base), &st[l])) {
+            return false;
         }
     }
 
-    delete[] status;
-    return ret;
+    return true;
 }
 
 
--- a/src/devices/ControlBoardWrapper/ControlBoardWrapperCommon.cpp
+++ b/src/devices/ControlBoardWrapper/ControlBoardWrapperCommon.cpp
@@ -47,39 +47,26 @@ bool ControlBoardWrapperCommon::setRefAcceleration(int j, double acc)
 
 bool ControlBoardWrapperCommon::setRefAccelerations(const double* accs)
 {
-    bool ret = true;
-    int j_wrap = 0; // index of the joint from the wrapper side (useful if wrapper joins 2 subdevices)
+    int j_wrap = 0;
 
-    // for all subdevices
     for (size_t subDev_idx = 0; subDev_idx < device.subdevices.size(); subDev_idx++) {
-        SubDevice* p = device.getSubdevice(subDev_idx);
+        auto* p = device.getSubdevice(subDev_idx);
 
-        if (!p) {
-            return false;
-        }
+        if (p && p->pos) {
+            std::vector<int> joints((p->top - p->base) + 1);
+            std::iota(joints.begin(), joints.end(), p->base);
 
-        int wrapped_joints = static_cast<int>((p->top - p->base) + 1);
-        int* joints = new int[wrapped_joints]; // to be defined once and for all?
-
-        if (p->pos) {
-            // verione comandi su subset di giunti
-            for (int j_dev = 0; j_dev < wrapped_joints; j_dev++) {
-                joints[j_dev] = static_cast<int>(p->base + j_dev);
+            if (!p->pos->setRefAccelerations(joints.size(), joints.data(), &accs[j_wrap])) {
+                return false;
             }
 
-            ret = ret && p->pos->setRefAccelerations(wrapped_joints, joints, &accs[j_wrap]);
-            j_wrap += wrapped_joints;
+            j_wrap += joints.size();
         } else {
-            ret = false;
-        }
-
-        if (joints != nullptr) {
-            delete[] joints;
-            joints = nullptr;
+            return false;
         }
     }
 
-    return ret;
+    return true;
 }
 
 
--- a/src/devices/ControlBoardWrapper/ControlBoardWrapperCommon.cpp
+++ b/src/devices/ControlBoardWrapper/ControlBoardWrapperCommon.cpp
@@ -10,6 +10,9 @@
 
 #include "ControlBoardWrapperLogComponent.h"
 
+#include <numeric> // std::iota
+#include <vector>
+
 
 bool ControlBoardWrapperCommon::getAxes(int* ax)
 {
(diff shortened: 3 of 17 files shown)


drdanz:
@PeterBowman Thanks for the PR.
At the moment we are working on splitting the ControlBoardWrapper into `nws-yarp` and `nws-ros`, which as you can probably imagine is a quite complicated task, due to the size of the object, and to the fact that the ControlBoardWrapper contains logic (that should not be in the nws as explained in #2441), and will probably require a 3rd plugin.
Therefore I kindly ask to wait for this PR until we manage to do the split, since there might be merge conflicts...

PeterBowman:
@drdanz sure, no problem, I was actually expecting that this PR would need to be postponed because of the split. It will remain in draft state and also awaiting a mandatory consultation I want to start before taking any action. For now I find it helpful to gather here the results of that small investigation of mine and a bunch of test cases (see gist) which can be revisited in the future.

traversaro:
> At the moment we are working on splitting the ControlBoardWrapper into nws-yarp and nws-ros, which as you can probably imagine is a quite complicated task, due to the size of the object, and to the fact that the ControlBoardWrapper contains logic (that should not be in the nws as explained in #2441), and will probably require a 3rd plugin.

Slightly related: regarding the "3rd plugin"do you think that for that logic it would be sufficient to use the `controlboardremapper` device (that was added in https://github.com/robotology/yarp/pull/746), or do you think a different one will be necessary?

traversaro:
By the way, I may be wrong but I think that the optimization that @PeterBowman is proposing for the controlboardwrapper2 was already implemented for the controlboardremapper (at least for some methods, see https://github.com/robotology/yarp/issues/1922 for a related issue). A possible strategy for @PeterBowman problem is instead of having one single physical device connected to 8 `controlboardwrapper2`,  you can have that the single physical device is connected to 8 `controlboardremapper` devices, and each `controlboardremapper` is connected to a single `controlboardwrapper2` that only deals with reading and wring data on the YARP network.

traversaro:
To have a bit more graphical representation of the role of `controlboardremapper`, you can check this slide from https://github.com/vvv-school/vvv18/blob/master/material/dynamics/yarp-motor-control-for-dynamics.pdf : 
wholebody

In this slide the remapper is shown on the client side (the original reason why this logic was decoupled from the `controlboardwrapper2`), but it can be easily used as well on the server side.

PeterBowman:
Thank you for all this info and pointers, @traversaro! I have never thought of the CB remapper being used in that way, I need to analyze it carefully.

drdanz:
> Slightly related: regarding the "3rd plugin"do you think that for that logic it would be sufficient to use the `controlboardremapper` device (that was added in #746), or do you think a different one will be necessary?

I'm still trying to figure it out...
There are still a couple of obscure parts of the device that I haven't fully understood yet.

PeterBowman:
> A possible strategy for @PeterBowman problem is instead of having one single physical device connected to 8 `controlboardwrapper2`, you can have that the single physical device is connected to 8 `controlboardremapper` devices, and each `controlboardremapper` is connected to a single `controlboardwrapper2` that only deals with reading and wring data on the YARP network.

@traversaro I have modified [this gist](https://gist.github.com/PeterBowman/5c99113b191c255624d5f540598e6ea3) to give it a try. Breaking down the results into three use cases:
* ControlBoardWrapper wrapping a fake subdevice: this was the original issue this PR was meant to address.
* ControlBoardWrapper wrapping a ControlBoardRemapper wrapping a fake subdevice: I think this is aligned with your proposed workaround and I confirm it does fix the problem! There are a few methods of ControlBoardWrapper that could get use of multi-joint wrapping (see point 3 in the description of this PR), though.
* ControlBoardRemapper wrapping a fake subdevice: this makes more sense in my test scenario since I don't actually need to open any ports at all. As in the previous case, I confirm it works but here it's actually flawless, all methods map perfectly: single-joint to single-joint, group to group, whole-wrapper to group (if available, otherwise iterates over single-joint).

The workaround (second case) would not be necessary if the new nws device performs method mapping via an embedded ControlBoardRemapper instance (by the way: could/should this be turned into a library instead of being a plugin/device?).
:warning: **However, I find ControlBoardRemapper perfect the way it is implemented now. Please note that changing it as I think it is proposed in https://github.com/robotology/yarp/issues/1922#issuecomment-440956344 would make it target too many subdevice joints, again.** :warning:

I believe this proposal does no longer make any sense. I'd rather apply the workaround now on the old ControlBoardWrapper and keep looking forward to using the new nws with improved mapping (yet to come), letting the old device alone. Discussions about the new implementations should probably be held elsewhere. Thank you for your help!

drdanz:
Just for reference, after lots of debates and discussions, we are going to promote the use of `controlboardremapper` + the new nws (work in progress) in favour of `controlboardwrapper2`.
We are also considering an extra device (`controlboardwrapper3`?) that opens the remapper and yarp/ros nws, accepting the same parameters of `controlboardwrapper2`. This will make the switch very easy, and can eventually be used to start the device using `yarpdev`

PR #2487, opened 2021-02-04
dev: Rename ControlBoardPid.h to Pid.h

drdanz:
### `dev`

* The header `yarp/dev/ControlBoardPid.h` was renamed `yarp/dev/Pid.h`

### DIFF
--- a/doc/release/master/rename_controlboardpid.md
+++ b/doc/release/master/rename_controlboardpid.md
@@ -1,0 +1,8 @@
+rename_controlboardpid {#master}
+----------------------
+
+## Libraries
+
+### `dev`
+
+* The header `yarp/dev/ControlBoardPid.h` was renamed `yarp/dev/Pid.h`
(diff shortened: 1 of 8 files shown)


randaz81:
Are you sure about this? I would suggest the opposite, i.e. renaming the class from  yarp::dev::Pid to yarp::dev::ControlBoardPid. This because pids can be of many types... and this specific class, which includes stiction coefficienst, is referring to a pid to be used for a motor control board.

pattacini:
>renaming the class from yarp::dev::Pid to yarp::dev::ControlBoardPid

Just to point out that this would result in a breaking change that will impact the user code, so needed to be handled with care.

drdanz:
> Are you sure about this?

Absolutely not... I just think that the fact that the class and the file name do not correspond is an API/Usability bug.
If you think that the opposite is better, we can change the name of the class, and add a deprecated `typedef` to keep compatibility.

Anyway, changing this will end up in the interfaces (IPid), and, even though I'm not sure if having the typedef the API will break, but that will be quite confusing for sure.

drdanz:
Closing for now

PR #2508, opened 2021-02-23
Update install_mac.dox

Nicogene:
See https://github.com/robotology/yarp/discussions/2507 for reference, it removes the brew installation since the formula is not maintained anymore


drdanz (approved):
Thanks

drdanz:
Merged, thanks.

PR #2514, opened 2021-03-07
[WIP] PublisherLatched

randaz81:
* added the new class `yarp::os::PublisherLatched`
* added test to `PublisherTest`

From http://wiki.ros.org/roscpp/Overview/Publishers%20and%20Subscribers#Publisher_Options :
_"When a connection is latched, the last message published is saved and automatically sent to any future subscribers that connect. This is useful for slow-changing to static data like a map."_


drdanz:
As discussed with @randaz81, this PR requires several changes, I'm marking this as draft...

randaz81:
withdrawn because yarp ros component will be become unmaintened/deprecated

PR #2518, opened 2021-03-12
.jpg support

randaz81:
* yarp::sig::ImageFile.cpp Added support to read .jpg files
* added rpc port to FakeFrameGrabber, to change its mode at runtime


pattacini (commented):
@randaz81 a few files seem to be in PR by mistake.

drdanz (approved):

PR #2538, opened 2021-03-29
fixed bug in YarpPluginSelector::scan()

ste93:
added a static mutex in order to avoid multiple concurrent accesses to singleton instance (ResourceFinder::getResourceFinderSingleton())


drdanz (approved):
LGTM, thanks

drdanz (changes requested):
I just realized that patch this should be merged into yarp-3.4 and not into master...
Also you should add the release notes

Do you mind doing this? Let me know if you need help...

drdanz:
Another option is to use `rebase onto` see also https://github.com/robotology/community/discussions/168

drdanz (approved):
Perfect, thanks!

PR #2539, opened 2021-03-30
Make install path of Python glue .so configurable

PeterBowman:
This is a follow-up to https://github.com/robotology/yarp/pull/2523. That PR allowed users to tweak the installation path of *yarp.py*. This PR applies a similar fix for the relocation of *_yarp.so*.

In other words, after #2523 and before this patch I had (with `CMAKE_INSTALL_PYTHON3DIR=lib/python3.8/dist-packages`) this in *install_manifest.txt*:

```
/usr/local/lib/python3.8/dist-packages/yarp.py
/usr/local/lib/python3/dist-packages/_yarp.so
```

Now both files should land in the `python3.8` tree.

Closes https://github.com/robotology/yarp/issues/2511.


drdanz (approved):
LGTM

drdanz:
Merged, thanks.

PR #2540, opened 2021-04-01
yarp: Add emergency sub-command

drdanz:
Following up the discussion about IP Theft (#2520), this patch introduces a simple way to try determine which port is the `impostor` port. This method is not 100% guaranteed to work, there is a chance to make things worse, but in some cases it is worth giving it a try.

### `yarp`

* Added `emergency` sub-command
  This command calls an emergency meeting and try to determine the impostor port.
  Each open port gets a vote, and the port who receives more votes is ejected from YARP.

image

213-YARPino-sus


Nicogene (approved):
💙

PR #2555, opened 2021-04-22
New Python example proposal: a buffered reader callback

PeterBowman:
This patch adds a tiny Python example as a reference for a quite basic functionality I found missing: a typed reader callback for a buffered port. Also, a legacy Python 2 example has been removed and its Python 3 counterpart renamed (dropped version suffix).


drdanz (approved):
LGTM, thanks!

drdanz:
Merged, thanks.

PR #2562, opened 2021-05-04
UltraPython Cameras

triccyx:
UltrPython cameras support added.
The UltraPython cameras are connected to an ARM  Xilinx board. So the usbcamera device, that is the portion of SW modified/added, to this PR should work on ARM architecture. The UnitTest can be also executed on x86 architecture.
The PR contains also a Matlab file for test purpose.


src/devices/usbCamera/linux/list.cpp
```diff
@@ -1,19 +1,9 @@
...
- *
- * This library is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
- * Lesser General Public License for more details.
- *
- * You should have received a copy of the GNU Lesser General Public
- * License along with this library; if not, write to the Free Software
- * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ * This software may be modified and distributed under the terms of the
```
Nicogene, line 5: There is a reason why you changed the license?

src/devices/usbCamera/linux/list.h
```diff
@@ -1,19 +1,9 @@
...
- * version 2.1 of the License, or (at your option) any later version.
- *
- * This library is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
- * Lesser General Public License for more details.
- *
- * You should have received a copy of the GNU Lesser General Public
- * License along with this library; if not, write to the Free Software
- * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
```
Nicogene, line 16: Same as above

src/devices/usbCamera/matlabscripts/FPS_check.m
```diff
@@ -0,0 +1,47 @@
...
+title('FPS in time')
+grid on
+
+nexttile
+mytime=info(2:end,2:2)-info(2,2);
+plot(mytime,timediff(2:end,2:2));
+xlabel('Time');
+ylabel('Frame interval in sec');
+title('Frame interval in time')
+grid on
```
Nicogene: What is the purpose of this script? Usually the scripts are put in the `script` folder and installed with the `module`/`device`, but if it is just a script that you used for the development maybe it is not worth to add to the repository

src/devices/usbCamera/ultrapython_ini/ultrapython_hires.ini
```diff
@@ -0,0 +1,10 @@
+device grabberDual
+subdevice usbCamera
+period 71
+capabilities COLOR
+twoCameras false
+name /grabber
+camModel ultrapython
+framerate 1
+honorfps false
+d /dev/media0
```
Nicogene: The ini files for running the device are usually specific to the machine on which you will run it, you can put them as example in the documentation of the device (see for example https://github.com/robotology/yarp/blob/0e8c016d10d43307e7b40ba3f7c173286e7d87b0/src/devices/ServerFrameGrabberDual/ServerFrameGrabberDual.h#L105-L141)

src/devices/usbCamera/ultrapython_ini/ultrapython_lowres.ini
```diff
@@ -0,0 +1,11 @@
...
+subdevice usbCamera
+period 30
+capabilities COLOR
+twoCameras false
+name /grabber
+camModel ultrapython
+subsampling
+framerate 1
+honorfps false
+d /dev/media0
```
Nicogene: Same comment as above

src/devices/usbCamera/unittest/testUltraPytonRunning.cpp
```diff
@@ -0,0 +1,106 @@
...
+      .WillOnce(Return(-1));
+
+  // when
+  bool out = helper.step();
+
+  // then
+  EXPECT_FALSE(out);
+
+  delete interface;
+}
```
Nicogene, line 106: This comment is for all the unit test you added.

The unit test in yarp are in the test folder(maybe you can put it there https://github.com/robotology/yarp/tree/master/tests/libYARP_dev, we have to ask to @drdanz)
3. In YARP we use CATCH2 for unit test(see [ref](https://github.com/catchorg/Catch2)) I would stick to it and not introducing a different test suite(gtest).

src/devices/usbCamera/linux/.clang-format
```diff
@@ -0,0 +1,14 @@
...
+UseTab:          Always
+ColumnLimit: 200
+
+Language: Cpp
+
+DisableFormat:   false
+Standard: Cpp11
+BreakBeforeBraces: Allman
+AllowShortIfStatementsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: false
```
Nicogene: Please remove this clang-format file and restore the one you deleted

src/devices/usbCamera/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+:exclamation:_Troubleshooting_
+
+1. If vscode won't connect try to check Xilinx board file system.
+   `fsck / -y` Then restart boot Xilinx board and vscode.
+2. If vscode still won't connect try to delete, on Xilinx board, the following files:
+
+```
+rm /root/.vscode-server/.*
+```
```
Nicogene, line 287: Super that you add the README! Could you please add also some lines for the generic usbCamera? The `ultraPython` will become a subsection of the documentation

src/devices/usbCamera/CMakeLists.txt
```diff
@@ -4,19 +4,60 @@
...
+  FetchContent_MakeAvailable(googletest)
+  add_subdirectory(unittest)
+  endif()
+
+# Get rid of undef while compiling gtest -- START
+set_property(
+    DIRECTORY
+    PROPERTY COMPILE_OPTIONS ${compile_options}
+)
+unset(compile_options)
```
Nicogene, line 44: This part of gtest has to be removed, see the comment below about the use of catch

Nicogene (changes requested):
This is the first iteration of review, there are some things to be changed/adjusted

src/devices/usbCamera/unittest/testUltraPytonRunning.cpp
triccyx: Relating to UnitTest they can not be written with other than gtest/gmock. This because the code deeply use Kernel API that should be moked (with gmock so gtest is mandatory). I suggest keeping these irregular UnitTests in this folder to avoid problems with the official tests.

pattacini: If this is viable/agreed, then we need to provide a CMake infrastructure that can detect/enable the GTest part and the corresponding unit tests smoothly and transparently w/o conflicting w/ the remaining test suite.

triccyx: However, there is a flag in the CMake. If activated the CMake will download the gtest and gmock library without problems. The alternative is to remove the tests but in my opinion is a pity.

src/devices/usbCamera/ultrapython_ini/ultrapython_hires.ini
triccyx: May I use ann examples folder in the code folder?

src/devices/usbCamera/matlabscripts/FPS_check.m
triccyx: I will revert them

src/devices/usbCamera/linux/list.h
triccyx: I will revert them

src/devices/usbCamera/linux/list.cpp
triccyx: I will revert them

src/devices/usbCamera/unittest/testUltraPytonRunning.cpp
Nicogene: I see, there is the possibility to use catch2 with an external mock library(here [the example](https://github.com/sinojelly/catch2-with-mockcpp) and here is the [mockcpp library](https://github.com/sinojelly/mockcpp)) or catch with gmock(see https://github.com/matepek/catch2-with-gmock), here is the related issue https://github.com/catchorg/Catch2/issues/1659. I think that is a thing to be discussed with @drdanz

src/devices/usbCamera/README.md
triccyx: No problems for me, @pattacini all right for you?

pattacini: Yep

src/devices/usbCamera/matlabscripts/FPS_check.m
triccyx: This script is used to check if the output from the usbcamera is correct. We can call it a system test. In the future we will need more of these test for checking other specific problems on the output. It could be useful also for the customer, we can think to install it.
In my opinion, should be near code, otherway it is difficult to find/execute it.

src/devices/usbCamera/unittest/testUltraPytonRunning.cpp
triccyx: Yes... but do not forget we will need gmock. So I do not see any advantage to use a mixed and potentially problematic test architecture. However I will follow @drdanz @Nicogene indications.

src/devices/usbCamera/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+```
+ENABLE_yarpmod_usbCamera        ON
+ENABLE yarpmod_usbCameraRaw     ON
+```
+
+The only new options for UltraPython are:
+
+```
+YARP_USE_UDev                    ON
```
PeterBowman, line 200: Although this README seems to suggest that UDev is specific to UltraPython, it is actually a hard dependency for the whole usbCamera device (see CMakeLists.txt), regardless of the underlying camera model. Is this intentional?

src/devices/usbCamera/unittest/testUltraPytonRunning.cpp
Nicogene: I would like to discuss in a 30min meeting what's the most convenient way to proceed, I don't like either to include these handcrafted solutions, but I don't know if we can integrate the gtest in the current test suite that uses catch. If the test cannot run in the CI I think it is pointless to have this test, because we risk forgetting its existence.

src/devices/usbCamera/README.md
pattacini: Also, `YARP_USE_UDev` doesn't sound very expressive and self-explaining, if it's meant to be specific to the UltraPython.

Better macros? `YARP_USE_UltraPython` looks better in this sense @triccyx.

triccyx: Well, YARP_USE_UDev is not related directly to UltraPyton but to libudev. This library is a standard Linux library for device management. I would not change the macro name as the library could be used for another purpose. @PeterBowman is it possible, with some effort and some #define, to link the use of this library only to UltraPython.. but the correct way to proceed should be to use the library also for USB device.

src/devices/usbCamera/unittest/testUltraPytonRunning.cpp
triccyx: Will you organize the meeting? @Nicogene

randaz81 (changes requested):
@triccyx could you clarify the dependency between UltraPython Camera device and USB cameras device? It seems to me that the two should be well distinct plugins. If they share some common code, then both of them should derive by a common class, a pattern that is largely used in Yarp. Unfortunately, USBcamera is a bad example. It _already_ contains stuff from leopard cameras, making the code difficult to comprehend. I fear that adding a third type of camera will make the situation even worse (in the past similar situation soon or later led to a total rewrite of the device from scratch) 
Some notable examples:
- lidar devices (such as rpLidar, rpLidar2, laserHokuyo, fakeLaser) derive from Lidar2DDeviceBase
- audio devices (such as portaudioPlayer,portaudioRecorder, audioFromFileDevice, audioToFileDevice, fakeMicrophone, fakeSpeaker) derive from AudioPlayerDeviceBase, AudioRecorderDeviceBase

pattacini:
Hi @randaz81,

Just to give a bit of backdrop.

@Nicogene @triccyx correct if I'm wrong, but I think that you guys decided to go this way initially by sort of copying out the "Leopard" approach. The UltraPython cameras are a replacement for the old Leopard, which can be eventually removed.

In this context, we didn't want to touch the USBcamera (not under our responsibility and kind of off topic wrt our original goal), but @triccyx tested that they do still work as intended.

Don't know what would entail refactoring all this but it will certainly take more time. Probably, a more iterative process may pay off here.

triccyx:
> @triccyx could you clarify the dependency between UltraPython Camera device and USB cameras device? It seems to me that the two should be well distinct plugins. If they share some common code, then both of them should derive by a common class, a pattern that is largely used in Yarp. Unfortunately, USBcamera is a bad example. It _already_ contains stuff from leopard cameras, making the code difficult to comprehend. I fear that adding a third type of camera will make the situation even worse (in the past similar situation soon or later led to a total rewrite of the device from scratch)
> Some notable examples:
> 
> * lidar devices (such as rpLidar, rpLidar2, laserHokuyo, fakeLaser) derive from Lidar2DDeviceBase
> * audio devices (such as portaudioPlayer,portaudioRecorder, audioFromFileDevice, audioToFileDevice, fakeMicrophone, fakeSpeaker) derive from AudioPlayerDeviceBase, AudioRecorderDeviceBase

In order to avoid completely change USBcamera I decide for a "composition" approach instead of inheritance. Also, this helps me in the testing phase.  Please take a look at README.md section 4.

drdanz (changes requested):
Just a preliminary review.

I share the same concerns as @PeterBowman and @randaz81...

* If I understand correctly, you are adding an extra dependency (UDev) to the usbCamera driver, which is used only by the ultrapython cameras, without actually adding anything for current users. Does it makes sense to enable the UltraPython camera only if UDev is available?
* Adding a new camera to the USB camera goes somehow against the reason for having plugins... But I agree that we already have the Leopard camera inside which is already wrong. I would suggest to refactor this before merging this change, since we know that if we merge this, it is very likely stay this way until the next pull request to support the next camera (GigaMegaloTyrannosaurusRex?), when we will have the same discussion.

About the testing framework, I don't have the time to evaluate the situation right now, but consider that:
* YARP build system is supposed to work without network, therefore FetchContent is not allowed. If required you can add the framework in the external folder.
* The testing framework used in YARP is Catch2, we can consider adding a Mocking framework, but I would not be in favour to add another testing framework, since supporting 2 testing frameworks instead of one will be at least two times as much problematic.
* Unit tests for other devices are in `tests/libYARP_dev` (that's probably wrong, and at some point we might want to move them, but at the moment this is the situation).

Finally, before merging into master, please rebase the PR and ensure that there are no `merge` commits in the history. Also I would suggest have a look if there is some commit that you can squash/reword (at least the `doc` ones), since I doubt that all the 43 commits are actually useful, and several commit logs are very little descriptive. You should also probably remove the GPL-2.0 file from the history.

src/devices/usbCamera/unittest/testUltraPythonGeneral.cpp
```diff
@@ -0,0 +1,194 @@
+/*
+ * Copyright (C) 2006-2021 Istituto Italiano di Tecnologia (IIT)
+ * All rights reserved.
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD-3-Clause license. See the accompanying LICENSE file for details.
+ */
+//# @author Luca Tricerri <luca.tricerri@iit.it>
```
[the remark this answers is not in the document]
triccyx: Ok!

src/devices/usbCamera/common/USBcamera.cpp
```diff
@@ -1,19 +1,9 @@
...
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
- * Lesser General Public License for more details.
- *
- * You should have received a copy of the GNU Lesser General Public
- * License along with this library; if not, write to the Free Software
- * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ * This software may be modified and distributed under the terms of the
+ * BSD-3-Clause license. See the accompanying LICENSE file for details.
  */
```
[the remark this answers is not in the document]
triccyx: Ok!

.clang-format
```diff
@@ -1,164 +0,0 @@
----
```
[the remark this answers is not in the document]
triccyx: I will format using official yarp .clang-format

src/devices/usbCamera/common/USBcamera.cpp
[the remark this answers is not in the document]
triccyx: Done

.clang-format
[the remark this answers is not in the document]
triccyx: Done

triccyx:
Together with @drdanz, we decide on a new strategy for UltraPython camera SW. Technically speaking we agreed that the correct way to deploy UltraPython camera SW is to create a separate repository. In this way, the USB/Leopard part can be removed from the code. Also, we can use the gmock test strategy. The repo will be: https://github.com/icub-tech-iit/yarp-device-ultrapython
I have also asked @drdanz support during code review in order to maintain a correct Yarp style. A new YCM release will be used for libudev support.

PR #2569, opened 2021-05-13
Remove classes and vocabs moved in framegrabber_protocol library

drdanz:
### `dev`

* All classes and vocabs moved in framegrabber_protocol library were removed.
  This is an API break, but nobody should be using these outside of YARP.


traversaro (approved):

Nicogene (approved):

Nicogene (commented):
My only concerns is if they are used in the external devices repos ( https://github.com/robotology/yarp-device-realsense2 or the future `ultrapython`) but I think they use the parser classes

drdanz:
The are not used in the USBcamera, therefore I'm quite sure that they are not used in the ultrapython.
For the other repository, I checked using GitHub search, and they seem to be used only in yarp

Nicogene:
fyi @triccyx

Issue #2590, opened 2021-06-04
Error generation from .msg files when using yarpidl_rosmsg


plinioMoreno:
**Describe the bug**
In the use-case where the .msg files and/or .srv files are located in the ROS catkin_ws files, the script cmake/yarpIDL.cmake has errors

**To Reproduce**
I created the repository [yarp-ros-srv-msg-example](https://github.com/plinioMoreno/yarp-ros-srv-msg-example/tree/generation-issue), where the branch `generation-issue` contains the files to replicate the error, and the required packages.

**Expected behavior**
If you run the commands
```
yarpidl_rosmsg roscpp_tutorials/TwoInts.srv --out ./
yarpidl_rosmsg rospy_tutorials/HeaderString.msg --out ./
yarpidl_rosmsg tf2_msgs/FrameGraph.srv --out ./
```
The result is correct, but when you run the similar command using `yarp_idl_to_dir()`, it does not work.

**Configuration (please complete the following information):**
 - OS: Ubuntu 18.04
 - yarp version: 3.4.101+20210524.9+git10dc2b5fb+dirty
-  compiler: gcc 7.5.0

**Additional context**

PR #2591, opened 2021-06-04
Fix message generation errors

plinioMoreno:
The changes in YarpIDL.cmake fix the generation issues reported in #2590, corresponding to the use-case where files located in the ROS workspace are used to generate the Yarp IDL headers.

The changes in RosType.cpp aim to remove the extra .srv and/or .msg in the filenames. For instance, the generated filenames are xxx.msg.msg or yyy.srv.srv

### DIFF
--- a/cmake/YarpIDL.cmake
+++ b/cmake/YarpIDL.cmake
@@ -27,6 +27,7 @@
 #                   [THRIFT_NO_EDITOR]
 #                   [THRIFT_NO_DOC]
 #                   [THRIFT_DEBUG_GENERATOR]
+#                   [ROSMSG_WITH_ROS]
 #                   [VERBOSE])
 #
 # yarp_add_idl
@@ -88,7 +89,9 @@ endfunction()
 # Internal function.
 function(_YARP_IDL_ROSMSG_ARGS _prefix _out_dir _verbose _out_var)
   unset(_args)
-  list(APPEND _args --no-ros true)
+  if(NOT ${_prefix}_ROSMSG_WITH_ROS)
+    list(APPEND _args --no-ros true)
+  endif()
   list(APPEND _args --no-cache)
   if(_verbose)
     list(APPEND _args --verbose)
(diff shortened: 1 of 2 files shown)


cmake/YarpIDL.cmake
```diff
@@ -313,10 +317,6 @@ function(YARP_IDL_TO_DIR)
     elseif("${_family}" STREQUAL "rosmsg")
       get_filename_component(_rospkg_name "${_include_prefix}" NAME)
       get_filename_component(_include_prefix "${_include_prefix}" DIRECTORY)
-      if(_rospkg_name MATCHES "(msg|srv)")
-        get_filename_component(_rospkg_name "${_include_prefix}" NAME)
-        get_filename_component(_include_prefix "${_include_prefix}" DIRECTORY)
-      endif()
```
drdanz, line 319: These lines that you removed, are actually required. It ensures to get the right package name when the file has a name like `<package>/<msg|srv>/<filename>`, like most of the files from ros source code (see the files in `extern/ros`).

For example `<path>/std_msgs/msg/String.msg`, the first 2 lines would set `_rospkg_name` to `msg` and `_include_prefix` to `<path>/std_msgs`, and the second step would change them to `<path>` and `std_msgs`

drdanz (commented):
Thanks @plinioMoreno
I believe the tests might be failing for this change

drdanz:
Replaced by #2592

cmake/YarpIDL.cmake
plinioMoreno: Okay, so there is something that needs to be added.
If you run the message generation with files that are located in a ROS catkin workspace (mentioned [here](https://github.com/robotology/yarp/issues/2590), without removing the lines I removed it does not work.

In my use-case the file `tf2_msgs/FrameGraph.srv` ([this line](https://github.com/plinioMoreno/yarp-ros-srv-msg-example/blob/generation-issue/CMakeLists.txt#L22)) has the string msgs, and variable `_include_prefix` is empty because of these lines.

PR #2592, opened 2021-06-04
Fix message generation errors - yarpidl_rosmsg and YarpIDL.cmake

plinioMoreno:
The changes in YarpIDL.cmake fix the generation issues reported in #2590, corresponding to the use-case where files located in the ROS workspace are used to generate the Yarp IDL headers.

The changes in RosType.cpp aim to remove the extra .srv and/or .msg in the filenames. For instance, the generated filenames are xxx.msg.msg or yyy.srv.srv


cmake/YarpIDL.cmake
```diff
@@ -313,10 +317,6 @@ function(YARP_IDL_TO_DIR)
     elseif("${_family}" STREQUAL "rosmsg")
       get_filename_component(_rospkg_name "${_include_prefix}" NAME)
       get_filename_component(_include_prefix "${_include_prefix}" DIRECTORY)
-      if(_rospkg_name MATCHES "(msg|srv)")
-        get_filename_component(_rospkg_name "${_include_prefix}" NAME)
-        get_filename_component(_include_prefix "${_include_prefix}" DIRECTORY)
-      endif()
```
drdanz, line 319: These lines that you removed, are actually required. It ensures to get the right package name when the file has a name like `<package>/<msg|srv>/<filename>`, like most of the files from ros source code (see the files in `extern/ros`).

For example `<path>/std_msgs/msg/String.msg`, the first 2 lines would set `_rospkg_name` to `msg` and `_include_prefix` to `<path>/std_msgs`, and the second step would change them to `<path>` and `std_msgs`

drdanz (changes requested):
Hi @plinioMoreno, sorry I just noticed that the other PR was closed in favour of this one. Please note that you don't have to open a new PR if you realize that the other one has some problem, you just need to push on the same branch, and the PR will be automatically updated.

Anyway, I will copy my comments here:

 I believe the tests might be failing for this change

cmake/YarpIDL.cmake
plinioMoreno: I believe that with the changes in [these lines](url) (https://github.com/robotology/yarp/pull/2592/commits/6e86119b0a0b657cb4f308ad7719de57a597db8d#diff-f898b3600d84f6a92c58a63ac324e5e479d24e6480548cfdad42330f91be2b66R180), the removed lines are not needed anymore. I run the message generation from scratch and the messages from the folder `extern/ros` worked properly. 

If you run the message generation with files that are located in a ROS catkin workspace (mentioned [here](https://github.com/robotology/yarp/issues/2590), without removing the lines I removed it does not work.

In my use-case the file `tf2_msgs/FrameGraph.srv` ([this line](https://github.com/plinioMoreno/yarp-ros-srv-msg-example/blob/generation-issue/CMakeLists.txt#L22)) has the string msgs, and variable `_include_prefix` is empty because of these lines.

drdanz: I see, actually I believe that the problem could be the regular expression that matches `tf2_msgs` while it should match only `msg` and `srv`...
Can you try replacing it with this?
```diff
-      if(_rospkg_name MATCHES "(msg|srv)")
+      if(_rospkg_name MATCHES "^(msg|srv)$")

plinioMoreno: I will try and let you know later today.

plinioMoreno: I applied your suggestion and now it is working properly.
Thanks for the info about the pull request, I did not know that you can keep adding commits after asking for a pull request.

drdanz (approved):
LGTM now

drdanz:
Merged, thanks.

PR #2596, opened 2021-06-07
Fix including on purpose

drdanz:
### `dev`

* The `IVisualParams.h` file was deprecated in favour of `IRgbVisualParams.h`
  and `IDepthVisualParams`.
* All headers for the deprecated interfaces are now properly deprecated.

Depends on #2593, #2583, #2594, and #2595


drdanz:
Valgrind failures seem to depend on some other reason. Merging this

PR #2597, opened 2021-06-08
Drop support for Qt < 5.4

drdanz:
### `yarpview`

* Qt 5.4 or later is now requred


drdanz:
Just for reference, Qt 5.4 is available since ubuntu 15.04, which makes me think that it is about time to drop support for Qt 5.3 😉

PR #2601, opened 2021-06-10
sig: pointcloudutils added ordered and unordered pointcloud generation

ste93:
added unordered point cloud generation to remove zero points

### DIFF
--- a/src/libYARP_sig/src/yarp/sig/PointCloudUtils-inl.h
+++ b/src/libYARP_sig/src/yarp/sig/PointCloudUtils-inl.h
@@ -67,7 +67,8 @@ inline void copyColorData(yarp::sig::PointCloud<T1>& pointCloud,
 template<typename T1, typename T2>
 yarp::sig::PointCloud<T1> yarp::sig::utils::depthRgbToPC(const yarp::sig::ImageOf<yarp::sig::PixelFloat>& depth,
                                                          const yarp::sig::ImageOf<T2>& color,
-                                                         const yarp::sig::IntrinsicParams& intrinsic)
+                                                         const yarp::sig::IntrinsicParams& intrinsic,
+                                                         const yarp::sig::utils::OrganizationType organizationType)
 {
     yAssert(depth.width()  != 0);
     yAssert(depth.height() != 0);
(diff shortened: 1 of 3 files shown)


src/libYARP_sig/src/yarp/sig/PointCloudUtils-inl.h
```diff
@@ -67,31 +67,51 @@ inline void copyColorData(yarp::sig::PointCloud<T1>& pointCloud,
...
+                    point.g = color.pixel(u,v).g;
+                    point.b = color.pixel(u,v).b;
+                    pointCloud.push_back(point);
+                    pointNumber ++;
+                }
+            }
         }
     }
+    if  (organizationType == yarp::sig::utils::OrganizationType::Unorganized) {
+        pointCloud.resize(pointNumber);
```
PeterBowman, line 113: Is this actually necessary, along with the `pointNumber` variable? The cloud should already have exactly `pointNumber` elements after the last `push_back` call inside the loop.

src/libYARP_sig/src/yarp/sig/PointCloudUtils-inl.h
```diff
@@ -67,31 +67,51 @@ inline void copyColorData(yarp::sig::PointCloud<T1>& pointCloud,
...
+    }
```
Nicogene, line 114: Why do you need this resize, since you are pushing_back element by element?

src/libYARP_sig/src/yarp/sig/PointCloudUtils-inl.h
Nicogene: @PeterBowman you anticipated me 😆

ste93: I can try delete it, I think maybe is not necessary

ste93: updated

ste93:
cc @randaz81

src/libYARP_sig/src/yarp/sig/PointCloudUtils-inl.h
PeterBowman: @ste93 the `pointNumber` variable is now unused.

randaz81 (approved):

Nicogene (approved):

drdanz (approved):

PR #2612, opened 2021-06-16
sig/Image: Fix topIsLow handling in YARP 3.4

drdanz:
As explained in #2609 and #2587, this patch restores compatibility in the protocol, when an image with `topIsLow==false` is sent by YARP 3.5.

---

#### `Image`

* Fixed compatibility with images with topIsLow==false sent by YARP 3.5.

### `yarpview`

* Fixed upside down image when topIsLow is false (Qt >= 5.3.2 only).


traversaro (approved):

Nicogene (approved):

Nicogene:
If it has been tested, LGTM

PR #2613, opened 2021-06-16
sig/Image: Add move() and swap() methods

drdanz:
These 2 methods allow to swap or move 2 images of different type (like [`copy()`](https://yarp.it/git-master/classyarp_1_1sig_1_1Image.html#a7792cdf9c1192dabd7f8ad2332895897))

------

#### Image

* Added `move()` and `swap()` methods.

### DIFF
--- a/src/libYARP_sig/src/yarp/sig/Image.cpp
+++ b/src/libYARP_sig/src/yarp/sig/Image.cpp
@@ -875,6 +875,36 @@ bool Image::copy(const Image& alt)
 }
 
 
+bool Image::move(Image&& alt) noexcept
+{
+    // Cannot move an image of the wrong type inside an ImageOf that does not
+    // support it.
+    yAssert(dynamic_cast<FlexImage*>(this) ||
+            getPixelCode() == alt.getPixelCode());
+    if (&alt != this) {
+        delete static_cast<ImageStorage*>(implementation);
+        implementation = std::exchange(alt.implementation, nullptr);
+        synchronize();
+    }
+    return true;
+}
+
+
+bool Image::swap(Image& alt)
+{
+    // Cannot swap two ImageOf of different type, or an image of the wrong type
+    // inside an ImageOf that does not support it.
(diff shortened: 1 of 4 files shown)


src/libYARP_sig/src/yarp/sig/Image.cpp
```diff
@@ -875,6 +875,36 @@ bool Image::copy(const Image& alt)
...
+        implementation = std::exchange(alt.implementation, nullptr);
+        synchronize();
+    }
+    return true;
+}
+
+
+bool Image::swap(Image& alt)
+{
+    // Cannot swap two ImageOf of different type, or an an image of the wrong
```
gsaponaro, line 895: Typo: `an an`

drdanz: Fixed, thanks

src/libYARP_sig/src/yarp/sig/Image.cpp
```diff
@@ -875,6 +875,36 @@ bool Image::copy(const Image& alt)
...
+            getPixelCode() == alt.getPixelCode());
+    if (&alt != this) {
+        std::swap(alt.implementation, implementation);
+        synchronize();
+        alt.synchronize();
+    }
+    return true;
+}
+
+
```
Nicogene, line 907: What about using this `move` and `swap` for refactoring the moves operators?

drdanz: It can be done, but I think it will make the move slower (at least in debug, since a dynamic_cast is required to check the type, which you don't need).
I think they could just be simplified instead...
The move constructor could just become
```diff
 Image::Image(Image&& other) noexcept
-    : implementation(other.implementation)
+    : implementation(std::exchange(other.implementation, nullptr))
 {
-    other.implementation = nullptr;
     synchronize();
 }
```

and the move assignment operator

```diff
 Image& Image::operator=(Image&& other) noexcept
 {
-    Image moved(std::move(other));
-    std::swap(moved.implementation, implementation);
+    implementation = std::exchange(other.implementation, nullptr);
     synchronize();
     return *this;
 }
```

PR #2637, opened 2021-07-03
Link against frameTransformRC via `target_link_libraries`

PeterBowman:
On latest YARP master (unreleased 3.5), a CMake 3.12 configure run throws the following error ([logs](https://github.com/roboticslab-uc3m/vision/runs/2976739659#step:11:539)):

```
CMake Error at src/devices/frameTransformClient/CMakeLists.txt:44 (target_sources):
  Error evaluating generator expression:

    $<TARGET_OBJECTS:frameTransformRC>

  Objects of target "frameTransformRC" referenced but is not an OBJECT
  library.
```

The offending line is:

https://github.com/robotology/yarp/blob/d0058d144991446df133c540a662371fafddbde5/src/devices/frameTransformClient/CMakeLists.txt#L44

According to CMake docs, `$<TARGET_OBJECTS:...>` requires an `OBJECT` target. However, the target created by `cmrc_add_resource_library` a few lines earlier is a `STATIC` library ([ref](https://github.com/vector-of-bool/cmrc/blob/e7ba9e9417960b2a5cefc9e79e8af8b06bfde3d1/CMakeRC.cmake#L468-L471)). For some reason, this generator expression does accept `STATIC` targets in recent CMake releases (tested at CMake 3.16.3), but it fails as expected for the minimum required CMake 3.12 version (tested at CMake 3.12.4).

I believe this instruction was somewhat inspired by the way "FrameTransformUtils" was referenced at line 36, which is in fact an `OBJECT` library. I think the simplest way to handle "frameTransformRC" is via `target_link_libraries`. That being said, this target does not register any interface include directory (regarding the now-deleted and superseded `target_include_directories` command).

Follows up https://github.com/robotology/yarp/pull/2624 (cc @elandini84). I'm also removing the `NAMESPACE` option to `cmrc_add_resource_library` since it defaults to the target name anyway ("frameTransformRC", [ref](https://github.com/vector-of-bool/cmrc/blob/e7ba9e9417960b2a5cefc9e79e8af8b06bfde3d1/CMakeRC.cmake#L406), [docs](https://github.com/vector-of-bool/cmrc/tree/e7ba9e9417960b2a5cefc9e79e8af8b06bfde3d1#usage)).

### DIFF
--- a/src/devices/frameTransformClient/CMakeLists.txt
+++ b/src/devices/frameTransformClient/CMakeLists.txt
@@ -7,7 +7,6 @@
 include(CMakeRC)
 
 cmrc_add_resource_library(frameTransformRC
-                          NAMESPACE frameTransformRC
                           WHENCE robotinterface_xml
                           PREFIX config_xml
                           robotinterface_xml/ftc_full_ros.xml
@@ -40,9 +39,9 @@ if(NOT SKIP_frameTransformClient)
                                                           YARP::YARP_sig
                                                           YARP::YARP_dev
                                                           YARP::YARP_math
-                                                          YARP::YARP_robotinterface)
-  target_sources(yarp_frameTransformClient PRIVATE $<TARGET_OBJECTS:frameTransformRC>)
-  target_include_directories(yarp_frameTransformClient PRIVATE $<TARGET_PROPERTY:frameTransformRC,INTERFACE_INCLUDE_DIRECTORIES>)
+                                                          YARP::YARP_robotinterface
+                                                          frameTransformRC)
+
   list(APPEND YARP_${YARP_PLUGIN_MASTER}_PRIVATE_DEPS YARP_os
                                                       YARP_sig
                                                       YARP_dev


PeterBowman:
Static builds are failing with:

```
CMake Error: install(EXPORT "YARP_yarpmod" ...) includes target "yarp_frameTransformClient" which requires target "frameTransformRC" that is not in any export set.
CMake Error in src/devices/CMakeLists.txt:
  export called with target "yarp_frameTransformClient" which requires target
  "frameTransformRC" that is not in any export set.
```

I wonder if this was the reason behind using `$<TARGET_OBJETS:...>`. See also https://gitlab.kitware.com/cmake/cmake/-/issues/17357.

PeterBowman:
Another approach I considered before applying the `target_link_libraries` solution was:

```cmake
target_sources(yarp_frameTransformClient PRIVATE $<TARGET_PROPERTY:frameTransformRC,SOURCES>)
# as stated earlier, it should be safe to remove the `target_include_directories` line that follows
```

The `SOURCES` property comprises all XML files passed to `cmrc_add_resource_library` **and** an auto-generated *lib_.cpp* source file. The silly thing here is that frameTransformRC would be needlessly compiled and linked into a static library - we'd reuse its sources and compile them again for yarp_frameTransformClient.

drdanz:
Thanks @PeterBowman.
We made this change because of the failing `STATIC` builds, since we would have to export the extra library, which it does not make any sense to export... See also https://github.com/vector-of-bool/cmrc/issues/28
If this fails on CMake 3.12, I think the best option is to ship a patched version of the `CMakeRC` module, which produces `OBJECT` libraries instead of `STATIC`... I opened a PR https://github.com/vector-of-bool/cmrc/pull/30, I'll import it in YCM asap

PeterBowman:
Thank you, @drdanz. I'll re-run our CI after https://github.com/robotology/yarp/pull/2638 is merged.

PeterBowman:
> I'll re-run our CI after either #2638 or #2639 is merged.

I can confirm YARP master builds fine now with CMake 3.12: [CI workflow run](https://github.com/roboticslab-uc3m/kinematics-dynamics/actions/runs/1008724500). Apparently this issue didn't show up on your own CMake 3.12-specific builds since Eigen is not installed, hence no YARP_math library, which is in turn a dependency for the frame transform devices.

PR #2673, opened 2021-07-30
rplidar sdk update

randaz81:
* `extern/rplidar` updated to the latest version from https://github.com/Slamtec/rplidar_sdk
* affects yarp device `rplidar2`

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -54,7 +54,7 @@ License: BSD-3-Clause
 Files: extern/rplidar/sdk/*
 Copyright:
  2009-2014 RoboPeak Team (http://www.robopeak.com)
- 2014-2018 Shanghai Slamtec Co., Ltd. (http://www.slamtec.com)
+ 2014-2019 Shanghai Slamtec Co., Ltd. (http://www.slamtec.com)
 License: BSD-2-Clause
 
 Files: extern/sqlite/sqlite/*
--- a/doc/release/master/rplidar_sdk_udate.md
+++ b/doc/release/master/rplidar_sdk_udate.md
@@ -1,0 +1,5 @@
+rplidar_sdk_update {#master}
+-----------------------
+
+### extern
+* updated `rplidar` sdk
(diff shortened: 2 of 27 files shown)


drdanz (changes requested):
Please update the following files:
* `extern/rplidar/README.txt`
* `.reuse/dep5`

Also it is probably worth checking if the new SDK version compiles with Clang and eventually enable it, since at the moment the rpLidar2 device is disabled on Clang (see [here](https://github.com/robotology/yarp/blob/bb647b69e084c295d6d7e31aeb36202abe42ccf9/src/devices/rpLidar2/CMakeLists.txt#L8))

drdanz:
Just for reference, I think that the sdk is used only for the `rpLidar2` device, I don't think that the `rpLidar device uses it...

drdanz:
> Also it is probably worth checking if the new SDK version compiles with Clang and eventually enable it, since at the moment the rpLidar2 device is disabled on Clang (see [here](https://github.com/robotology/yarp/blob/bb647b69e084c295d6d7e31aeb36202abe42ccf9/src/devices/rpLidar2/CMakeLists.txt#L8))

Related issue: https://github.com/Slamtec/rplidar_sdk/issues/14

randaz81:
> Please update the following files:
> 
> * `extern/rplidar/README.txt`
> * `.reuse/dep5`
> 
> Also it is probably worth checking if the new SDK version compiles with Clang and eventually enable it, since at the moment the rpLidar2 device is disabled on Clang (see [here](https://github.com/robotology/yarp/blob/bb647b69e084c295d6d7e31aeb36202abe42ccf9/src/devices/rpLidar2/CMakeLists.txt#L8))

done

randaz81:
> > Also it is probably worth checking if the new SDK version compiles with Clang and eventually enable it, since at the moment the rpLidar2 device is disabled on Clang (see [here](https://github.com/robotology/yarp/blob/bb647b69e084c295d6d7e31aeb36202abe42ccf9/src/devices/rpLidar2/CMakeLists.txt#L8))
> 
> Related issue: [Slamtec/rplidar_sdk#14](https://github.com/Slamtec/rplidar_sdk/issues/14)

I think there are no updates on this issue

drdanz (approved):
LGTM now

PR #2678, opened 2021-08-09
Use `mkdir_p` instead of `mkdir` w/ `getHomeContextPath` and `getHomeRobotPath`

### LINKED ISSUE #2676: Use `mkdir_p` in lieu of `mkdir` within `getHomeContextPath`
**Describe the bug**
Unable to automatically create the path to `getHomeContextPath` when calling that function.

Have a look at:
https://github.com/robotology/yarp/blob/0c5873cdf02a6cb4bea0d375049399dd0f556a90/src/libYARP_os/src/yarp/os/ResourceFinder.cpp#L686-L694

Because of

https://github.com/robotology/yarp/blob/0c5873cdf02a6cb4bea0d375049399dd0f556a90/doc/release/v3_5_0.md#L127-L129

I think that the function `yarpdatahome` no longer creates the desired path and we use the fallback `mkdir`.
However, the correct function for creating the directory should be `mkdir_p` instead.

**To Reproduce**

This [snippet](https://github.com/robotology/icub-main/blob/master/src/tools/depth2kin/src/module.cpp#L1182-L1188) should work out also in case the dir `~/.local/share/yarp/contexts` is not available.

Anyway, I discovered that in one of the latest YARP versions the path doesn't get created. 

**Expected behavior**
When calling `getHomeContextPath`, we should be able then to open a file therein.

 - OS: Linux
 - yarp version: `3.4.101~dev`

pattacini:
This PR fixes #2676.

I've tested it successfully ✔️ 

I'm proposing this change back to yarp-3.4 to then easily port it forward as for Distro 2021.08 we would likely need to rely on a fixed `yarp-3.4.6`.

I've tried to fill in the mandatory fields for describing the fix by taking inspiration from supporting files done for other old mods.
Tell me if I'm required to do anything else.

### DIFF
--- a/doc/release/yarp_3_5/ResourceFinder_fix.md
+++ b/doc/release/yarp_3_5/ResourceFinder_fix.md
@@ -1,0 +1,12 @@
+ResourceFinder_fix {#yarp_3_5}
+-----------
+
+Important Changes
+-----------------
+
+### Libraries
+
+##### `ResourceFinder`
+
+* Use `mkdir_p` instead of `mkdir` when calling `getHomeContextPath` and `getHomeRobotPath`
+  to correctly create the missing paths. 
--- a/src/libYARP_os/src/yarp/os/ResourceFinder.cpp
+++ b/src/libYARP_os/src/yarp/os/ResourceFinder.cpp
@@ -685,10 +685,10 @@ public:
 
         std::string parentPath = getPath(yarp::conf::dirs::yarpdatahome(), "contexts", "", "");
         if (yarp::os::stat(parentPath.c_str()) != 0) {
-            yarp::os::mkdir(parentPath.c_str());
+            yarp::os::mkdir_p(parentPath.c_str());
         }
 
-        if (yarp::os::mkdir(path.c_str()) < 0 && errno != EEXIST) {
+        if (yarp::os::mkdir_p(path.c_str()) < 0 && errno != EEXIST) {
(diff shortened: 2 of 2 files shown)


drdanz:
Hello @pattacini, sorry for the delay.

Sorry, I'm confused, was the directory creation already failing with `getHomeContextPath()` or is this a regression introduced by the change to `yarp::conf::dirs::yarpdatahome()` in YARP 3.5?

pattacini:
Sorry @drdanz my fault.

I overlooked the fact that the change to `yarp::conf::dirs::yarpdatahome()` I cited in the issue was actually introduced in YARP-3.5. I was probably misled by the YARP version I checked at that time, which was `3.4.101~dev` that eventually landed in YARP-3.5.

So, after another round of tests I've verified that:
- ❌ `yarp@master` failed with the creation of the directory.
- ✔ `yarp@v3.4.6` runs successfully.

Therefore, I'm gonna update this PR accordingly (sticking to the `yarp-3.4` compare branch in order to avoid crafting a brand new one).

pattacini:
Ready for review again.

>I'm proposing this change back to yarp-3.4 to then easily port it forward as for Distro 2021.08 we would likely need to rely on a fixed yarp-3.4.6.

Of course, the above doesn't apply any longer.

drdanz (approved):

drdanz:
Merged in yarp-3.5. Thanks @pattacini!

PR #2689, opened 2021-08-19
Yarp log label

randaz81:
You can now use the new environment variable `YARP_LOG_PROCESS_LABEL` to set an extra label which helps in identifying the name of the process which is performing the log

### DIFF
--- a/doc/release/master/yarp_log_label.md
+++ b/doc/release/master/yarp_log_label.md
@@ -1,0 +1,15 @@
+yarp_log_label {#master}
+-----------
+
+Important Changes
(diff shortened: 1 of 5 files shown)


randaz81:
http://www.yarp.it/git-master/yarp_logging.html to be updated

drdanz (changes requested):
The patch looks good, but please add the documentation of this new environment variable in `doc/yarp_logging.md` and `doc/yarp_env_variables.md`

drdanz (approved):

Issue #2691, opened 2021-08-21
File names clash under Windows


pattacini:
**Describe the bug**

The following files generate names clash under case-insensitive OS like Windows:

  `src/libYARP_dev/src/idl_generated_code/yarp/dev/AudioPlayerStatus_common.h`
  `src/libYARP_dev/src/idl_generated_code/yarp/dev/audioPlayerStatus_common.h`
  `src/libYARP_dev/src/idl_generated_code/yarp/dev/AudioRecorderStatus_common.h`
  `src/libYARP_dev/src/idl_generated_code/yarp/dev/audioRecorderStatus_common.h`

**To Reproduce**
Just clone YARP anew on Windows and you'll get:

```console
$ git clone https://github.com/robotology/yarp.git
Cloning into 'yarp'...
remote: Enumerating objects: 188299, done.
remote: Counting objects: 100% (10627/10627), done.
remote: Compressing objects: 100% (4523/4523), done.
remote: Total 188299 (delta 6047), reused 9888 (delta 5765), pack-reused 177672
Receiving objects: 100% (188299/188299), 89.63 MiB | 4.76 MiB/s, done.
Resolving deltas: 100% (135372/135372), done.
Updating files: 100% (6937/6937), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'src/libYARP_dev/src/idl_generated_code/yarp/dev/AudioPlayerStatus_common.h'
  'src/libYARP_dev/src/idl_generated_code/yarp/dev/audioPlayerStatus_common.h'
  'src/libYARP_dev/src/idl_generated_code/yarp/dev/AudioRecorderStatus_common.h'
  'src/libYARP_dev/src/idl_generated_code/yarp/dev/audioRecorderStatus_common.h'
```

**Expected behavior**
We should avoid name clash from upfront as on Windows Git operates in such a way that it removes locally one group of files leaving the working tree modified.

**Configuration (please complete the following information):**
 - OS: Windows 10
 - yarp version: `master`
-  compiler: not relevant

**Additional context**

---

cc @randaz81 @drdanz

pattacini:
@randaz81 @drdanz any news on this?

pattacini:
Fixed via #2699 ✔️

PR #2699, opened 2021-08-31
fixed warning for file collision on case insensitive os

randaz81:
fix for https://github.com/robotology/yarp/issues/2691


drdanz:
@randaz81 This should be fixed in the yarp-3.5 branch... Also this should be mentioned in the release notes

drdanz:
Merged, thanks @randaz81.

PR #2700, opened 2021-08-31
Remove all 'using namespace std'

drdanz:


traversaro (approved):
Blindly trusting the PR, I did not checked all the files.

drdanz:
BTW, this patch started due to some compile error when building YARP with C++17, due to the introduction of `std::byte` that causes a conflict.
The main issue was fixed in https://github.com/robotology/yarp/commit/ac080f34334ab3ddc6acb4bc18e6457a019a2011, but I think this is a [bad practice](https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice) and we should eradicate `using namespace` from the source code.

pattacini (approved):

PR #2703, opened 2021-09-02
Enable compilation of `libYARP_math` in the Python wheels

diegoferigo:
Eigen was missing in the debian container used to build the PEP600 wheels, and `libYARP_math` automatically disabled by CMake. This PR adds the target to the process, hopefully fixing https://github.com/dic-iit/bipedal-locomotion-framework/pull/296#issuecomment-911429287.


drdanz (approved):
LGTM, but I think it would be better to have it in the yarp-3.5 branch, so that YARP 3.5.1 release will have this. @diegoferigo What do you think?

diegoferigo:
> LGTM, but I think it would be better to have it in the yarp-3.5 branch, so that YARP 3.5.1 release will have this. @diegoferigo What do you think?

Good idea, rebased on top of the 3.5 branch

Nicogene (approved):

drdanz:
Merged, thanks @diegoferigo.

PR #2708, opened 2021-09-06
Fix log under yarprun always verbose 

drdanz:
After 9a16a9a056469779f5c7ea9e00a13f7e000575e9 the custom print callback is no longer called when running under yarprun


randaz81 (approved):
tested ok!

PR #2713, opened 2021-09-12
Avoid copies between `cv::Mat` and `yarp::sig::Image` in opencv_grabber

### LINKED ISSUE #2161: opencv_grabber should use YARP_cv utilities
`opencv_grabber` converts the `cv::Mat` to a `yarp::sig::image` with a `memcpy`.

https://github.com/robotology/yarp/blob/0ac66ed8b703eecd2207e24dbff4432008761b7b/src/devices/opencv/OpenCVGrabber.cpp#L196-L200

We should use instead `fromCvMat` (introduced in #1932) that does not involves copies:

https://github.com/robotology/yarp/blob/0ac66ed8b703eecd2207e24dbff4432008761b7b/src/libYARP_cv/src/yarp/cv/Cv-inl.h#L90-L112

PeterBowman:
Resolves #2161. In contrast to the suggested solution, I am converting a `yarp::sig::Image` to a `cv::Mat` so that final allocation occurs at the YARP caller side (doing it on a local `cv::Mat` would produce memory corruption as soon as the execution flow exits `getImage`) without using YARP_cv at all since color conversion must take place after the frame is grabbed. As a side note, I noticed in-place `cv::cvtColor` produces bogus results (frames flash between BGR and RGB representations), hence the auxiliary `bgr` variable.

Disclaimer: copying still happens whenever frame transposition or resizing is requested.

In addition, I have enabled `--flip_x`, `--flip_y` and `--transpose` for data streams originating from a file on disk (using `--movie`), which also helped me test these changes.

### DIFF
--- a/doc/release/yarp_3_5/opencv_no_copy.md
+++ b/doc/release/yarp_3_5/opencv_no_copy.md
@@ -1,0 +1,9 @@
+opencv_no_copy {#yarp_3_5}
+-----------
+
+### Devices
+
+#### `opencv_grabber`
+
+* Efficiency has been improved as copies are (in most cases) no longer
+  performed in order to convert from OpenCV frames to YARP image structures.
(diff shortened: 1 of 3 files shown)


drdanz:
Thanks @PeterBowman.
@Nicogene this patch looks good to me, can you double check?

Nicogene (approved):
LGTM

drdanz:
Merged, thanks. (sorry for the delay)

PR #2715, opened 2021-09-14
Feature: frameTransform custom config

elandini84:
* Added the option to use an extern xml file (by providing its absolute path) to configure the `frameTransformServer` and the `frameTransformClient` devices.
* **NB:** this function is intended for TEST ONLY


drdanz (approved):

drdanz:
Merged, thanks.

PR #2745, opened 2021-11-02
os/Log: Add the new family of logging macros with "Id"

drdanz:
#### `Log`

* Added the new family of logging macros with "Id" (`yIDebug`).
  See [documentation](\ref yarp_logging).

#### `DeviceDriver`

* Added `id()` and `setId()` methods

#### `PolyDriver`

* The device id can be set passing the `id` option to the `open` method (or
  passing `--id` to `yarpdev`).

### `portaudio*`,

* 'id' parameter was renamed to 'dev_id'.


PeterBowman:
@drdanz I'm sorry for posting this here as off-topic, but I wanted to test the patch and have just noticed:

https://github.com/robotology/yarp/blob/d7b6df3af0c15bdc156e023da5882e3ce3893c69/CMakeLists.txt#L20

Shouldn't ongoing YARP devel version read `3.5.101` or something like that?

drdanz:
> @drdanz I'm sorry for posting this here as off-topic, but I wanted to test the patch and have just noticed:
> 
> https://github.com/robotology/yarp/blob/d7b6df3af0c15bdc156e023da5882e3ce3893c69/CMakeLists.txt#L20
> 
> Shouldn't ongoing YARP devel version read `3.5.101` or something like that?

Thanks @PeterBowman!
Good catch... This should have been handled by the release action, but probably there is still some bug

Fixed manually for now: 7eb77f263936e466e44344e6207aa721d6e914ea

PR #2747, opened 2021-11-04
Fix yarpmanager white spaces check in ports names and ports names color

elandini84:
#### `yarpmanager` + `manager`

* Targeted issue: port names containing trailing or leading white spaces not targeted as `non-existing` but throwing an error
when trying to connect.
* Now `yarpmanager` shows an error message in its `Message` text box when a port name listed in the `Connection list` widget contains white spaces. For simplicity's sake the error is thrown for spaces in any position, not only leading or trailing
* This has been done by adding a check in the [Manager::existPortFrom](https://github.com/robotology/yarp/blob/2aeab6c8e4babe651febc2ea755a5409f4cfeda7/src/libYARP_manager/src/yarp/manager/manager.cpp#L731) and [Manager::existPortTo](https://github.com/robotology/yarp/blob/2aeab6c8e4babe651febc2ea755a5409f4cfeda7/src/libYARP_manager/src/yarp/manager/manager.cpp#L745) functions
* Also, fixed the color of port names in the `Connections list` widget. If a port name is edited by the user, the new name
is automatically checked and if the port exists, the color is changed to green, if it doesn't the color is changed to red.

### DIFF
--- a/doc/release/yarp_3_5/fix_yarpmanager_white_spaces_and_colors_port_names.md
+++ b/doc/release/yarp_3_5/fix_yarpmanager_white_spaces_and_colors_port_names.md
@@ -1,0 +1,14 @@
+fix_yarpmanager_white_spaces_and_colors_port_names {#yarp_3_5}
+-------------------
+
+### Tools + Libraries
+
+#### `yarpmanager` + `manager`
+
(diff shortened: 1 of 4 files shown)


Nicogene (approved):
I like these changes, from my experience I would test this kind of PRs with very big applications, that has a lot of modules and connection in order to see if this kind of checks introduce overheads or issues

PR #2758, opened 2021-11-14
Chain most std containers, C-style arrays and tuples in LogStream

PeterBowman:
I have always found the ability to chain `std::vector` to LogStream a great shortcut. This patch expands it to most standard containers (specifically tested with `array`, `deque`, `forward_list`, `list`, `set`, `unordered_set`), including key-value associations via `std::pair` (hence it also supports `map` and `unordered_map`), as well as old C-style arrays and even `std::tuple`.

To illustrate this quickly, the following tuple:

```cxx
std::tuple<int, std::set<int>, std::vector<std::map<double, char>>> t;
t = {5, {7, 6}, {{{1.1, 'g'}, {5.5, 'f'}}, {{9.0, 'h'}}}};
```

is rendered by `yInfo() << t;` as `[INFO] {5, [6, 7], [[{1.1, g}, {5.5, f}], [{9, h}]]}`.

### DIFF
--- a/src/libYARP_os/src/yarp/os/LogStream.h
+++ b/src/libYARP_os/src/yarp/os/LogStream.h
@@ -19,7 +19,9 @@
 #include <iosfwd>
 #include <sstream>
 #include <string>
-#include <vector>
+#include <tuple>
+#include <type_traits>
+#include <utility>
 
 namespace yarp::os {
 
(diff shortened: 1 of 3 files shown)


PeterBowman:
SWIG doesn't like these new templates, that's why I'm disabling them via `SWIG_PREPROCESSOR_SHOULD_SKIP_THIS` (note this macro was previously unused). There should be no difference since (at least on Python) the `std::vector` operator overload also did not get handled before this patch.

PeterBowman:
I'd very much fancy a `std::initializer_list<T>` overload to be able to do `yInfo() << {1.0, 2.0}`, but sadly this is not possible ([SO](https://stackoverflow.com/a/11445905/10404307)).

src/libYARP_os/src/yarp/os/LogStream.h
```diff
@@ -247,26 +249,80 @@ class YARP_os_API LogStream
         return *this;
     }
 
-    template <typename T>
-    inline LogStream& operator<<(const std::vector<T>& t)
+#ifndef SWIG_PREPROCESSOR_SHOULD_SKIP_THIS
+    template <typename T1, typename T2>
+    inline LogStream& operator<<(const std::pair<T1, T2>& t)
+    {
+        stream->oss << '{' << t.first << ", " << t.second << '}';
```
PeterBowman, line 256: Perhaps I should have split this chain into `stream->oss << '{'; *this << t.first; ...`. I'll check it.

src/libYARP_os/src/yarp/os/LogStream.h
```diff
@@ -247,26 +249,80 @@ class YARP_os_API LogStream
...
+        if (!stream->nospace) {
+            stream->oss << ' ';
+        }
+        return *this;
+    }
+
+    template <typename Container, typename = typename Container::value_type>
+    inline LogStream& operator<<(const Container& cont)
```
drdanz, line 264: Perhaps this should be a C++17 template template parameter? i.e. something like;
```c++
template<typename T, typename Alloc = std::allocator<T>, template <typename, typename> class Container>
inline LogStream& operator<<(const Container<T, Alloc>& t)
```

drdanz (commented):
Thanks for this, it looks quite useful... I'll do a proper review when you mark it as ready, but meanwhile I noticed this...

src/libYARP_os/src/yarp/os/LogStream.h
PeterBowman: Yes, this needs a similar treatment as vectors in order to be able to use nested containers (e.g. `std::pair<int, std::vector<double>>`) and avoid excess space chars. Done.

src/libYARP_os/src/yarp/os/LogStream.h
```diff
@@ -247,26 +249,80 @@ class YARP_os_API LogStream
         return *this;
     }
 
-    template <typename T>
-    inline LogStream& operator<<(const std::vector<T>& t)
+#ifndef SWIG_PREPROCESSOR_SHOULD_SKIP_THIS
```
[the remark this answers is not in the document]
PeterBowman: Done, I wasn't aware of this automatic definition. You might want to delete [this line in yarp.i](https://github.com/robotology/yarp/blob/7e8148bc62d05152885879b012d98945669d14c1/bindings/yarp.i#L17), then ([added](https://github.com/robotology/yarp/blob/5313b2defd8be7237567ea56eca8dde93dfc8121/doc/release/v2_3_72_1.md#bindings) in v2.3.72.1).

src/libYARP_os/src/yarp/os/LogStream.h
PeterBowman: I'm not sure how this is related to C++17 specifically (perhaps type deduction?), I lack a proper knowledge of this standard yet. It believe it was already achievable with older standards per [this SO answer](https://stackoverflow.com/a/213811/10404307). It seems to me the less convolution, the better (see UPDATE section). In addition, [`std::set<Key, Compare, Allocator>`](https://en.cppreference.com/w/cpp/container/set), among others, might not work.

drdanz:
Really useful addition, thanks!

drdanz (approved):

drdanz:
Merged, thanks.

PR #2759, opened 2021-11-14
Add `YARP_VERSION_COMPARE` macro to yarp/conf/version.h

PeterBowman:
Heavily inspired by [pcl_config.h](https://github.com/PointCloudLibrary/pcl/blob/af3a2ea43447c9e0d2b3757fedb8f5a09d3ea93e/pcl_config.h.in#L17-L21), this patch adds a new macro in order to simplify the way YARP versions are compared in `#if` guards or similar. For instance:

```c
#if YARP_VERSION_COMPARE(>=, 3, 6, 0)
    // do something only on YARP 3.6.0 or later
#else
    // otherwise, do this instead
#endif
```

**Why?**

1. Commodity. It is easier to compose a comparable major.minor.patch version in a single statement than fiddling with three-level comparisons which involve `YARP_VERSION_MAJOR`, `YARP_VERSION_MINOR` and `YARP_VERSION_PATCH`.
2. My programmer's senses tell me that YARP 4.0 is approaching. I have planted a lot of `#if YARP_VERSION_MINOR > X` checks around our organization's codebase under the assumption that YARP major versions are not compatible ([ref](https://github.com/robotology/yarp/blob/c3b05bc6141156542228547aba78b14328c93c90/cmake/YarpDescribe.cmake#L58), [CMake docs](https://cmake.org/cmake/help/v3.16/module/CMakePackageConfigHelpers.html#generating-a-package-version-file)). This is not entirely true as I noticed you try to keep as much backward compatibility as possible, yet I messed up by not accounting for the major version in those `#if` checks (should I have done it, I'd hit 1. anyway). I think it would have been much better if I had used a macro such as the one proposed here from the very beginning.

What are your oppinions on this? Should I start a discussion elsewhere first?


drdanz (approved):

drdanz:
Nice addition! Thanks @PeterBowman

drdanz:
> My programmer's senses tell me that YARP 4.0 is approaching.

At the moment we are still in the planning phase, but your programming senses are not wrong ;)

drdanz:
> What are your oppinions on this? Should I start a discussion elsewhere first?

I agree that we should have thought about this since the beginning, but it's too late now for the old versions... but we will have it in YARP 3.6, so you can eventually check `#if !defined(YARP_VERSION_COMPARE)` to check if YARP < 3.6.

PR #2764, opened 2021-11-18
Yarpmotorgui with displayed hardware faults messages 

randaz81:
See discussion: https://github.com/robotology/community/discussions/561

* Modified yarpmotorgui to support a new page "Hardware Fault" which displays the internal hardware error
* Added new interface IJointFault with method: IJointFault::getLastJointFault(int j, int& fault, std::string& message)
* IJointFault is now implemented by RemoteControlBoard and ControlBoard_nws_yarp
* fakeMotionControl.cpp now simulates an hardware fault when it receives a TorqueCommand > 1Nm

### DIFF
--- a/bindings/yarp.i
+++ b/bindings/yarp.i
@@ -416,6 +416,7 @@ MAKE_COMMS(Bottle)
 %include <yarp/dev/ControlBoardPid.h>
 %include <yarp/dev/IControlMode.h>
 %include <yarp/dev/IInteractionMode.h>
+%include <yarp/dev/IJointFault.h>
 %include <yarp/dev/IEncodersTimed.h>
 %include <yarp/dev/IMotor.h>
 %include <yarp/dev/IMotorEncoders.h>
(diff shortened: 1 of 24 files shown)


PeterBowman (commented):
1. I wonder if `bool getLastJointFault(int j, int* fault, std::string& message)` would be more consistent with existing interfaces (note the second parameter, as a primitive type, is passed as a pointer instead of by-reference).
2. Could you please also register this new interface in [yarp.i](https://github.com/robotology/yarp/blob/7e8148bc62d05152885879b012d98945669d14c1/bindings/yarp.i)? I think it just needs an `%include`, a `CAST_POLYDRIVER_TO_INTERFACE(IJointFault)` and an extension for output parameters such as this (not tested):
   ```swig
   %extend yarp::dev::IJointFault {
       bool getLastJointFault(int j, std::vector<int>& fault, std::vector<std::string>& message) {
           return self->getLastJointFault(j, fault[0], message[0]);
       }
   }
   ```
   (change `fault[0]` to `&fault[0]` if a pointer is passed instead of a reference)

randaz81:
Thank you @PeterBowman 
Point 1: I'm trying to use & instead of * in new interfaces. It prevents users from doing bad things, such as passing a nullptr.  Of course, I cannot easily change old ones.
Point 2: done

drdanz (approved):

Issue #2725, opened 2021-10-10
Python 2 headers are used to compile Python 3 bindings on systems with legacy PCL


PeterBowman:
On Bionic with VTK 6.3, Python 2.7 headers are added to the search path by `find_package(VTK)` (through `find_package(PCL)`) and given more precedence than Python 3 headers as provided by FindPython3. This can be checked in the `CXX_INCLUDES` property in `build/bindings/python/CMakeFiles/_yarp_python.dir/flags.make`:

```
CXX_INCLUDES = -I/usr/include/vtk-6.3 -I/usr/include/freetype2 -I/usr/lib/x86_64-linux-gnu/openmpi/include/openmpi -I/usr/lib/x86_64-linux-gnu/openmpi/include/openmpi/opal/mca/event/libevent2022/libevent -I/usr/lib/x86_64-linux-gnu/openmpi/include/openmpi/opal/mca/event/libevent2022/libevent/include -I/usr/lib/x86_64-linux-gnu/openmpi/include -I/usr/include/python2.7 -I/usr/include/hdf5/openmpi -I/usr/include/libxml2 -I/usr/include/jsoncpp -I/usr/include/tcl -I/home/bartek/git/yarp/src/libYARP_conf/src -I/home/bartek/git/yarp/build-3.5/src/libYARP_conf/src -I/home/bartek/git/yarp/src/libYARP_os/src -I/home/bartek/git/yarp/src/libYARP_sig/src -I/home/bartek/git/yarp/src/libYARP_dev/src -I/home/bartek/git/yarp/build-3.5/src/libYARP_dev/src -I/home/bartek/git/yarp/src/libYARP_math/src -isystem /usr/include/python3.6m
```

Note the `-I/usr/include/python2.7`. This path has more precedence than Python 3.6 headers (see `-isystem /usr/include/python3.6m` at the very end).

Despite using a Python 3 interpreter and having those bindings installed in a Python 3 directory, upon doing `import yarp` this obscure error is thrown (I had to use `ipython` since the default interpeter simply shows an segmentation fault):

<details>
<summary>ipython3 -c 'import yarp'</summary>

```
In [1]: import yarp                                                                                                                                                                                         
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
/usr/local/lib/python3.6/dist-packages/yarp.py in swig_import_helper()
     13         try:
---> 14             return importlib.import_module(mname)
     15         except ImportError:

[...]

PR #2766, opened 2021-11-22
Give Python3 headers more precedence over 3rd-party transitive deps

PeterBowman:
On Bionic with VTK 6.3, Python 2.7 headers are added to the search path by `find_package` and given more precedence than Python 3 headers as provided by FindPython3.

Fixes https://github.com/robotology/yarp/issues/2725.


drdanz (approved):

drdanz:
Merged, thanks.

PR #2667, opened 2021-07-27
os/impl/LogForwarder: Do not use a static Bottle

drdanz:
When the `forward` method is called twice, the `Bottle::clear()` method could be called before the message is sent, hence deleting the bottle being sent.

This fixes the `LogForwarder` skipping messages when messages are sent too quickly (Fixes #2643).

----

#### `os`

* The `LogForwarder` no longer skips messages when messages are sent too
  quickly (#2643).

Issue #2643, opened 2021-07-06
Yarplogger may not report all entries


sgiraz:
**Bug description**

I happened to spot an unexpected behavior concerning the tracing of operations such as `yInfo`, `yWarning`, `yError` etc. within the `yarplogger`.

If two operations are very close each other in time, I can observe that the `yarplogger` may not keep track of their invocations. In particular, the lower the time interval the higher the probability of a missed trace.

**To Reproduce**

In order to reproduce the behavior with the smallest code snippet, I crafted a [`yInfo-test`](https://github.com/sgiraz/yInfo-test) repository where you can find a simple case study (fully documented).

The [`yInfo-test`](https://github.com/sgiraz/yInfo-test) application contains a simple for-loop that performs 10k iterations with just one call to the `yInfo` function per cycle. Note that after each iteration a [yarp::os::Time::delay](http://www.yarp.it/git-master/namespaceyarp_1_1os_1_1Time.html#aca80809e0597551745954e2547e581ee) is also called so we can vary the elapsed time between the `yInfo` calls. 

The test scenario foresees the use of `yarplogger` without the `yarprun` broker, thus relying on the `YARP_FORWARD_LOG_ENABLE=1` env var. 

After some quick tests, it came out that there exists a threshold of 0.001s. Below this value (included), the number of missing `yInfo` logged by the `yarplogger` dramatically increases.

Please, notice that the for-loop is only aimed to replicate the behavior, whereas two very close logging operations may occur wherever in the user code. 

Below, you can see some results comparing what was produced by the `yarplogger` with the expected results: 

<details>
<summary>🔘 Click to see the result of Test 1</summary>

```console
./check_log_file.sh build/yarprunlog_04_07_2021_23_46_12.log
Parsing log file...
Missing value 1707, previous was at [ Time delay ]: 0.001 Counter value: 1706
Done.
```
</details>

<details>
<summary>🔘 Click to see the result of Test 2</summary>

```console
./check_log_file.sh build/yarprunlog_04_07_2021_23_41_24.log
Parsing log file...
Missing value 3224, previous was at [ Time delay ]: 0.0001 Counter value: 3223
Missing value 3512, previous was at [ Time delay ]: 0.0001 Counter value: 3511
Missing value 3513, previous was at
Missing value 3514, previous was at
Missing value 3515, previous was at
Missing value 8594, previous was at [ Time delay ]: 0.0001 Counter value: 8593
Missing value 8595, previous was at
[...]

PR #2767, opened 2021-11-22
Fix logger lost messages

randaz81:
A possible solution for issue https://github.com/robotology/yarp/issues/2643, already addressed by  PR https://github.com/robotology/yarp/pull/2667 , but using a different, not-blocking technique.

The performance of this solution still has to be investigated.


drdanz:
Unfortunately `PortWriterBuffer` is used internally in `BufferedPort` and blocks when there is another message in the queue. Therefore this implementation has the same issue as #2667

randaz81:
replaced by #2789

PR #2772, opened 2021-12-06
added device for synchronization

ste93:
systemReady {#master}
-----------------------

#### devices
* added systemReady device that opens some ports when called. It is useful for synchronize processes opened with yarprobotinterface and yarpmanager to avoid crashes when some processes are not started

### DIFF
--- a/src/devices/systemReady/SystemReady_nws_yarp.cpp
+++ b/src/devices/systemReady/SystemReady_nws_yarp.cpp
@@ -27,12 +27,12 @@ bool SystemReady_nws_yarp::open(yarp::os::Searchable &config)
             token = current_port_property.substr(0, pos);
             std::string current_port_name = port_list.find(token).asString();
             yarp::os::Port* current_port = new yarp::os::Port;
+            port_pointers_list.push_back(current_port);
             if(!current_port->open(current_port_name)){
                 yCError(SYSTEMREADY_NWS_YARP) << "error opening " << current_port_name;
                 close();
                 return false;
             }
-            port_pointers_list.push_back(current_port);
         }
         return true;
     }
@@ -45,6 +45,7 @@ bool SystemReady_nws_yarp::close()
 {
     for (auto elem: port_pointers_list){
         elem->close();
+        delete elem;
     }
     port_pointers_list.clear();
     return true;
--- a/src/devices/systemReady/SystemReady_nws_yarp.cpp
+++ b/src/devices/systemReady/SystemReady_nws_yarp.cpp
@@ -16,9 +16,9 @@ SystemReady_nws_yarp::SystemReady_nws_yarp()
 bool SystemReady_nws_yarp::open(yarp::os::Searchable &config)
 {
     if (config.check("PORT_LIST")) {
-        yarp::os::Bottle port_list = config.findGroup("PORT_LIST");
         // skips the first one since it is PORT_LIST
-        for (int index = 1; index < port_list.size(); index++) {
+        yarp::os::Bottle port_list = config.findGroup("PORT_LIST").tail();
+        for (int index = 0; index < port_list.size(); index++) {
             std::string current_port_property = port_list.get(index).toString();
             std::string token;
             int pos;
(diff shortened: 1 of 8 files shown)


src/devices/systemReady/SystemReady_nws_yarp.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+SystemReady_nws_yarp::SystemReady_nws_yarp()
+{
+}
+
+bool SystemReady_nws_yarp::open(yarp::os::Searchable &config)
+{
+    if (config.check("PORT_LIST")) {
+        yarp::os::Bottle port_list = config.findGroup("PORT_LIST");
+        // skips the first one since it is PORT_LIST
```
PeterBowman, line 20: It is also possible to use [`yarp::os::Bottle::tail`](http://www.yarp.it/latest/classyarp_1_1os_1_1Bottle.html#a9bac31d80e659b650cb643449b93d171) as a shortcut.

src/devices/systemReady/SystemReady_nws_yarp.cpp
```diff
@@ -0,0 +1,49 @@
...
+            char delimiter = ' ';
+            pos = current_port_property.find(delimiter);
+            token = current_port_property.substr(0, pos);
+            std::string current_port_name = port_list.find(token).asString();
+            yCError(SYSTEMREADY_NWS_YARP) << port_list.get(index).asList()[0].toString();
+            yCError(SYSTEMREADY_NWS_YARP) << current_port_property;
+            yCError(SYSTEMREADY_NWS_YARP) << token;
+            yCError(SYSTEMREADY_NWS_YARP) << port_list.find(token).asString();
+            yarp::os::Port* current_port = new yarp::os::Port;
+            current_port->open(current_port_name);
```
PeterBowman, line 34: This might fail if the port cannot be opened (name conflict?). I'd add an `if` clause here and move the `push_back` call before that (to prevent a memory leak on close).

src/devices/systemReady/SystemReady_nws_yarp.cpp
```diff
@@ -0,0 +1,49 @@
...
+    return false;
+}
+
+
+bool SystemReady_nws_yarp::close()
+{
+    for (auto elem: port_pointers_list){
+        elem->close();
+        delete elem;
+    }
```
PeterBowman, line 48: It could be wise to `port_pointers_list.clear()` just in case.

src/devices/systemReady/SystemReady_nws_yarp.h
```diff
@@ -0,0 +1,60 @@
+/*
+ * SPDX-FileCopyrightText: 2006-2021 Istituto Italiano di Tecnologia (IIT)
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef YARP_SYSTEMREADY_NWS_YARP_H
+#define YARP_SYSTEMREADY_NWS_YARP_H
+
+#include <yarp/sig/Vector.h>
```
PeterBowman, line 9: This is not used, but two other includes are missing instead: `<vector>` and `<yarp/os/Port.h>`.

src/devices/systemReady/SystemReady_nws_yarp.cpp
```diff
@@ -0,0 +1,51 @@
...
+            pos = current_port_property.find(delimiter);
+            token = current_port_property.substr(0, pos);
+            std::string current_port_name = port_list.find(token).asString();
+            yarp::os::Port* current_port = new yarp::os::Port;
+            if(!current_port->open(current_port_name)){
+                yCError(SYSTEMREADY_NWS_YARP) << "error opening " << current_port_name;
+                close();
+                return false;
+            }
+            port_pointers_list.push_back(current_port);
```
PeterBowman, line 35: This should go before the `if` clause, otherwise the newly constructed `current_port` will not be accessible by the `close` handler for deletion, thus causing a memory leak.

src/devices/systemReady/SystemReady_nws_yarp.cpp
```diff
@@ -0,0 +1,51 @@
...
+    }
+    yCError(SYSTEMREADY_NWS_YARP) << "missing PORT_LIST group";
+    return false;
+}
+
+
+bool SystemReady_nws_yarp::close()
+{
+    for (auto elem: port_pointers_list){
+        elem->close();
```
PeterBowman, line 47: There used to be a `delete` here in previous revisions, it needs to be kept to avoid memory leaks.

PeterBowman (changes requested):

PR #2773, opened 2021-12-06
Improve joints tree view in yarpmotorgui

S-Dafarra:
This PR edits the tree view as follows:
- it shows each joint name
- the joint name and mode are in a separate widget
- each part uses a flow layout to have all the joints fitting in the screen
- when double click in a joint on the tree, the focus automatically moves to the corresponding joint in the tab view
- with a right click on the joint in the tree it is possible to home, idle, run or show the PID dialog of the corresponding joint.

### DIFF
--- a/src/yarpmotorgui/CMakeLists.txt
+++ b/src/yarpmotorgui/CMakeLists.txt
@@ -19,6 +19,8 @@ if(YARP_COMPILE_yarpmotorgui)
     sliderOptions.cpp
     sliderWithTarget.cpp
     startdlg.cpp
+    jointItemTree.cpp
+    partItemTree.cpp
   )
   set(yarpmotorgui_HDRS
     flowlayout.h
@@ -32,6 +34,8 @@ if(YARP_COMPILE_yarpmotorgui)
     yarpmotorgui.h
     sequencewindow.h
     startdlg.h
+    jointItemTree.h
+    partItemTree.h
   )
   set(yarpmotorgui_QRC_FILES
     res.qrc
@@ -43,6 +47,7 @@ if(YARP_COMPILE_yarpmotorgui)
     sequencewindow.ui
     startdlg.ui
     sliderOptions.ui
+    jointItemTree.ui
   )
 
   qt5_add_resources(yarpmotorgui_QRC_GEN_SRCS ${yarpmotorgui_QRC_FILES})
--- a/src/yarpmotorgui/jointItemTree.cpp
+++ b/src/yarpmotorgui/jointItemTree.cpp
@@ -14,6 +14,32 @@ JointItemTree::JointItemTree(int index, QWidget *parent) :
 {
     m_ui->setupUi(this);
     setAttribute(Qt::WA_StyledBackground, true);
+
+    this->setContextMenuPolicy(Qt::CustomContextMenu);
+
+    connect(this, SIGNAL(customContextMenuRequested(QPoint)),
+            this, SLOT(onShowContextMenu(QPoint)));
+
+    m_rightClickMenuTitle = new QAction(this);
+    m_rightClickMenuTitle->setEnabled(false);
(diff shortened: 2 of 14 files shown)


S-Dafarra:
Here is an example of usage 

https://user-images.githubusercontent.com/18591940/144997106-be3c95a3-212c-44b3-b94e-024751e6b3f4.mp4

randaz81:
Hi @S-Dafarra,
I'd say that I like the "name of the joints" and the possibility of controlling them through the new "pop-up" menu. But the box visualization seems a bit confusing to me, and I prefer the previous single-line-for-joint style...

S-Dafarra:
> Hi @S-Dafarra, I'd say that I like the "name of the joints" and the possibility of controlling them through the new "pop-up" menu. But the box visualization seems a bit confusing to me, and I prefer the previous single-line-for-joint style...

Hi @randaz81! The main reason for choosing that layout was to have the possibility of displaying all the joints at the same time. Right now, for iCub robots with all the fingers, the list of joint does not fit the screen of a normal laptop

randaz81:
PS: an idea if you want to control ll the joints from a single tab (maybe you are already aware of it). You can define an XML with an additional controlboardremapper device and an associated nws. In this way you will have a new robot part called e.g. "all-joints" that you can visualize in a single tab of the gui. 
PS2: requirements: yarp >3.5.1 + the robot configuration files should be updated to use the nws devices

S-Dafarra:
> PS: an idea if you want to control all the joints from a single tab (maybe you are already aware of it). You can define an XML with an additional controlboardremapper device and an associated nws. In this way you will have a new robot part called e.g. "all-joints" that you can visualize in a single tab of the gui. PS2: requirements: yarp >3.5.1 + the robot configuration files should be updated to use the nws devices

I think that is an interesting alternative, although it does not fix the problem of having all the joint status fitting in a single window. The main reason I tried this modification was to have a status of the whole robot without having to scroll, so that I can look at it from time to time and immediately detect if a joint is in HF or not

randaz81:
Well, actually this solves your issue, cause if just a  joint is in hardware fault, the whole "all-joints" part becomes red.

S-Dafarra:
> Well, actually this solves your issue, cause if just a joint is in hardware fault, the whole "all-joints" part becomes red.

Nope, because it often happens that one finger is in HF. It makes a difference if you have the `r_hand_finger` going in HF or the knee. It is quite common that a finger goes in HF since startup. This will avoid spotting if other joints go in HF afterwards

randaz81:
Maybe we can proceed in this way then. The left part of the GUI could be a tabbed widget that provides multiple visualizations. One tab will be the original treeview. Other tabs could be dedicated to your visualization (and extra). The whole thing works only iff you are able to keep your visualization implementation _very well separated into different files_ .  This because the code of the GUI is pretty complicated, and we need to be somehow strict to guarantee some order and readability for posterity.

S-Dafarra:
> Maybe we can proceed in this way then. The left part of the GUI could be a tabbed widget that provides multiple visualizations. One tab will be the original treeview. Other tabs could be dedicated to your visualization (and extra). The whole thing works only iff you are able to keep your visualization implementation _very well separated into different files_ . This because the code of the GUI is pretty complicated, and we need to be somehow strict to guarantee some order and readability for posterity.

Sounds good to me! Regarding the code, yes, I totally agree. Right now most of the implementation is done in separate classes. Nonetheless, the ``mainwindow`` is still the part of the code that creates the tree and does the main connections. There is still some chance of improvements there I guess

S-Dafarra:
In the meantime I opened https://github.com/robotology/yarp/pull/2774 cherry-picking some fix I did in this PR

S-Dafarra:
I also opened #2776 to reduce the code in the ``mainwindow``, porting the tree-related code to some separate files. In that PR I also added the joint name

randaz81:
Replaced by https://github.com/robotology/yarp/pull/2823

PR #2775, opened 2021-12-07
disabled failing test on windows

ste93:
there is a failing test on windows machines, the failure reason is the one below.
Screenshot from 2021-12-07 09-33-41


drdanz:
CC @PeterBowman

PeterBowman:
My bad, I have created https://github.com/robotology/yarp/pull/2777 to address this.

PR #2788, opened 2021-12-27
Actually return a reference in `VectorOf<T>::emplace_back`

PeterBowman:
This method was not returning anything contrary to what was stated in the documentation. Actually, it would have made sense to return `void` due to the signature of `std::vector<T>::emplace_back` in previous standards. In C++17 and later (hence I wouldn't think of backporting this as-is to yarp-3.5), it does return a reference instead: [cppreference.com](https://en.cppreference.com/w/cpp/container/vector/emplace_back).

Affects YARP 3.4.0+: https://github.com/robotology/yarp/pull/2195.

### DIFF
--- a/doc/release/master.md
+++ b/doc/release/master.md
@@ -322,6 +322,14 @@ Bug Fixes
 * Fixed build of lidar sdk with clang.
 
 
+### Libraries
+
+#### `sig`
+
+* Fixed `VectorOf<T>::emplace_back` returning `void`. Now it returns a
+  reference to the new element as expected.
+
+
 ### GUIs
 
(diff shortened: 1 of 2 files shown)


drdanz (approved):
LGTM, thanks!

drdanz:
I think that, since this requires C++17, a proper "backport" fix would be to make the emplace method `void` in the YARP 3.5 branch, but it is technically an API break... We are not releasing YARP 3.5.2 anyway (we are releasing YARP 3.6.0 later today), so I think it is not worth fixing it in the other branches.

PR #2789, opened 2021-12-28
Fix logger lost messages V2

randaz81:
LogForwarder: now using a separated thread (`class ThreadedPort`) to prevent the loss of log messages during stress condition


randaz81:
Related to  issue  #2643

PR #2836, opened 2022-04-28
yarprobotinterface extern-name fix

randaz81:
fixes issue https://github.com/robotology/yarp/issues/2812 about parsing lists in extern-name parameter of yarprobotinterface.


randaz81:
all tests passed, rebasing

PR #2843, opened 2022-05-18
Periodic yarp write

randaz81:
added the possibility to use `yarp write` to periodically write a message


src/libYARP_companion/src/yarp/companion/impl/Companion.cmdWrite.cpp
```diff
@@ -137,22 +186,30 @@ int Companion::write(const char *name, int ntargets, char *targets[]) {
...
-        //core.send(bot);
         port.write(bot);
     }
 
-    //core.close();
-    //core.join();
-    port.close();
+    if (period > 0)
+    {
+        writer_thread->stop();
```
PeterBowman, line 194: Memory leak: missing `delete writer_thread`.

randaz81: fixed, ty!

PR #2849, opened 2022-06-08
Changes to the devices folder tree

randaz81:


randaz81:
Superseded and split into multiple PRs, i.e.:
- https://github.com/robotology/yarp/pull/2895

randaz81:
good idea, but the PR is now obsolete becuase of several other changes and it should be reimplemented from scratch

PR #2841, opened 2022-05-05
CB_NWS encoder check

randaz81:
Check if the encoders timestamps are consistent

PR #2862, opened 2022-06-26
Refresh timestamps on consistency check failure

PeterBowman:
This bugfix is a follow-up to https://github.com/robotology/yarp/pull/2841 (cc @randaz81). The `run` loop is stuck on the consistency check after raising an error since there is no way to refresh the `times` vector again, thus rendering the state ports unusable from that point onwards. Our robot is capable of recovery on comms failure (which happens quite often) on any joint, therefore patch https://github.com/robotology/yarp/pull/2841 forces us to restart the control board. This didn't happen prior to YARP 3.7.

I might suggest later a way to disable or configure this consistency checker differently. Anyway, I believe this patch targets a bug and should be addressed first.

### DIFF
--- a/doc/release/yarp_3_7/cbw_consistency_timestamps.md
+++ b/doc/release/yarp_3_7/cbw_consistency_timestamps.md
@@ -1,0 +1,9 @@
+cbw_consistency_timestamps {#yarp_3_7}
+-----------
+
+### Devices
+
+#### `controlBoard_nws_yarp`
+
+* Encoder timestamps are now queried by the consistency checker on failure,
+  thus allowing to recover once the hardware is available again.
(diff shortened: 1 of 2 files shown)


PeterBowman:
Hello, @drdanz, you are right. Since we are moving this check to the bottom, I think there is no need to call `getEncodersTimed` again (in fact, the timestamps would be out of sync wrt. their corresponding encoder values), so I have removed this previously added part. Besides, now the `averageTime` variable definition is closer to where it's used, and the branch has been rebased.

PeterBowman:
> Just an additional curiosity about this issue: the problem should not occur unless the timestamps differ more than 1second, which is a huge time. Why does this happen on your system/robot?

In our hardware setup, there is one physical motion controller per joint, connected via CAN bus to a CPU where our centralized control board implementation resides. Often, we observe communication issues on specific (not all) controllers, but we are prepared to deal with that: relying on a heartbeat signal, the conflicting device is restarted when a reasonable timeout elapses. We are able to command working joints while some others are still trying to restart, i.e. in our case robot joints are somewhat independent from each other.

randaz81 (approved):

PR #2869, opened 2022-07-12
Throttle error message on inconsistent encoder timestamps

PeterBowman:
As a follow-up to https://github.com/robotology/yarp/pull/2862, this patch throttles the error message printed whenever an inconsistency in encoder timestamps is detected. The "Encoder timestamps are not consistent! Data will not be published." is logged every few milliseconds and bloats the terminal. I'm using an arbitrary throttle of one second to mitigate this.

@randaz81 I think the nws_ros{.h,.cpp} files at src/devices/ControlBoardWrapper have been accidentally left out after https://github.com/robotology/yarp/pull/2845.  Perhaps you want to delete them first, then I can rebase this patch accordingly. As a side note, the consistency check patch from https://github.com/robotology/yarp/pull/2841 was applied to both nws_yarp and nws_ros in the ControlBoardWrapper directory, but the latter was copied into a different location using a version prior to said changes.

As commented in https://github.com/robotology/yarp/pull/2862, I'm still considering other ways to treat (or perhaps disable) consistency checking due to my particular use case.

### DIFF
--- a/doc/release/yarp_3_8/throttle_consistency_checker.md
+++ b/doc/release/yarp_3_8/throttle_consistency_checker.md
@@ -1,0 +1,8 @@
+throttle_consistency_checker {#yarp_3_8}
+-----------
+
+### Devices
+
+#### `controlBoard_nws_yarp`
+
+* The error message upon inconsistent encoder timestamps has been throttled.
--- a/src/devices/controlBoard_nws_yarp/ControlBoard_nws_yarp.cpp
+++ b/src/devices/controlBoard_nws_yarp/ControlBoard_nws_yarp.cpp
@@ -437,7 +437,7 @@ void ControlBoard_nws_yarp::run()
     // check we are not overflowing with input messages
     constexpr int reads_for_warning = 20;
     if (inputStreamingPort.getPendingReads() >= reads_for_warning) {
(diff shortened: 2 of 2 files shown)


PeterBowman:
I'm leaving the changes to the ROS device out of this patch since the device itself has been moved to a different repository at https://github.com/robotology/yarp/commit/48a7dd0d493732b296055c37bb8dcb00dc112261.

> As commented in #2862, I'm still considering other ways to treat (or perhaps disable) consistency checking due to my particular use case.

The use case being: some (but not all) wrapped subjoints stop responding and may resume communication with the board at some later point, in which case and in the current state of things the wrapper device is temporarily muted (not publishing state) by the consistency checker. I would like to keep it alive also in this scenario. Note https://github.com/robotology/yarp/pull/2862 allowed unmuting it if **all** subdevices were made available again. Would it be acceptable in a future PR if the board keeped sending state referring to the still working subdevices, while discarding the outdated timestamps in the average value (used as port envelope)?

PeterBowman:
I have changed the base branch to yarp-3.8 (it looks like it has triggered some unwanted actions, sorry!). I have also added logging component identifiers (`yCIError(name, id(), ...)`) to reflect which wrapper has issued the error among many available. Perhaps those IDs could be extended to more places and other devices as well?

randaz81 (approved):

PR #2871, opened 2022-07-17
Allow chaining `yarp::sig::VectorOf<T>` to LogStream

PeterBowman:
Follow-up to https://github.com/robotology/yarp/pull/2758. Similarly to STL containers, C arrays, pairs and tuples, now `yarp::sig::VectorOf<T>` is also chainable to stream logging macros, e.g.:

```cxx
yarp::sig::Vector v {1.23, 5.0};
yInfo() << v;
```

(Disclaimer: one can obviously call `v.toString()` instead; this patch harmonizes YARP and STL vectors, though.)

### DIFF
--- a/doc/release/master/000_master.md
+++ b/doc/release/master/000_master.md
@@ -1,14 +1,14 @@
-YARP <yarp-3.7> (UNRELEASED)                                         {#yarp_3_7}
+YARP <yarp-3.8> (UNRELEASED)                                         {#yarp_3_8}
 ============================
 
 [TOC]
 
-YARP <yarp-3.7> Release Notes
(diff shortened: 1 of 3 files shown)


PeterBowman:

I would very much like to reflect this change somewhere, but the `doc/release/master/` directory is missing. I think it should be restored as release notes are quite an important tool for developers aiming to adapt their code to newer YARP releases.

Also, I noticed the version is no longer updated for master (experimental) branch:

https://github.com/robotology/yarp/blob/05ca89400b36bf26aea8cb4707f3029d14ccbda7/CMakeLists.txt#L18-L22

randaz81 (approved):

PR #2873, opened 2022-08-09
IAxisInfo::getAxes()

randaz81:
Added method bool getAxes(int* ax) to yarp::dev::IAxisInfo interface


PeterBowman:
Please also add this method to yarp.i:

https://github.com/robotology/yarp/blob/14df562aa253fb8d041ba98eaa7fb96d3281017a/bindings/yarp.i#L1299-L1313

Something along these lines:

```cxx
 int getAxes() {
    int buffer;
    bool ok = self->getAxes(&buffer);
    if (!ok) return 0;
    return buffer;
}
```

traversaro (dismissed):

Nicogene (commented):
I would add the method in `yarp.i` as suggested by @PeterBowman

randaz81:
> Please also add this method to yarp.i:
done

PR #2877, opened 2022-09-08
Improved integration tests in CI

randaz81:
Added integration tests to check the correct execution of the following commands:

yarp --version
yarpdev --list
yarp plugin --all


randaz81:
NB: The CI tests in this PR are expected to fail.  Indeed the purpose of this PR is to introduce new tests against a couple of errors that were not detected. These errors will be addressed later on in a another PR.

PR #2907, opened 2022-11-30
YARP_dev_tests library can be now found outside yarp

randaz81:
well... maybe.

This is experimental stuff. Do not merge.


randaz81:
superseded

PR #2921, opened 2023-01-09
soxr enabled in CI

randaz81:
soxr stuff, which was previously disabled in CI, has been enabled again


randaz81:
superseded by https://github.com/robotology/yarp/pull/2923

PR #2927, opened 2023-01-18
Check tests

randaz81:
Added check_tests.py script to verify that each device has a corresponding test


randaz81:
To be merged after https://github.com/robotology/yarp/pull/2920

PR #2931, opened 2023-01-19
Relax requirement on timestamp consistency checker

PeterBowman:
Don't account for inconsistent timestamps (e.g due to disconnection or hardware error) as long as there is at least one subdevice still reporting current data. Use valid timestamps only for obtaining the average timestamp used as envelope in broadcast state messages.

Also, throttle the inconsistency error message to avoid flooding logs.

### DIFF
--- a/doc/release/master/nws_yarp_consistency.md
+++ b/doc/release/master/nws_yarp_consistency.md
@@ -1,0 +1,15 @@
+nws_yarp_consistency {#master}
+-------------------
+
+### Devices
+
+#### `controlBoard_nws_yarp`
+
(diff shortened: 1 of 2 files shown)


PeterBowman:
On a second thought, I might want to keep publishing state even if the last read happened a while ago (e.g. some or all subdevices are currently disconnected). The main question here is: what is the desired behavior if two nodes differ significantly in the timestamp of the last successful acquisition? How is the average time used as port envelope supposed to be obtained in this case?

PeterBowman:
I have proposed a different approach at https://github.com/robotology/yarp/issues/2939.

PR #2935, opened 2023-01-23
Fix handling of `IRemoteVariables` messages via RPC

PeterBowman:
Bugfix: a wrong pointer check could prevent from parsing remote `IRemoteVariables` commands.

### DIFF
--- a/doc/release/yarp_3_7/cbw_remote_vars.md
+++ b/doc/release/yarp_3_7/cbw_remote_vars.md
@@ -1,0 +1,8 @@
+cbw_remote_vars {#yarp_3_7}
+-------------------
+
+### Devices
+
+#### `controlBoard_nws_yarp`
+
+* Fix handling of `IRemoteVariables` messages via RPC.
--- a/src/devices/ControlBoardWrapper/RPCMessagesParser.cpp
+++ b/src/devices/ControlBoardWrapper/RPCMessagesParser.cpp
@@ -1232,10 +1232,10 @@ void RPCMessagesParser::handlePWMMsg(const yarp::os::Bottle& cmd, yarp::os::Bott
 
 void RPCMessagesParser::handleRemoteVariablesMsg(const yarp::os::Bottle& cmd, yarp::os::Bottle& response, bool* rec, bool* ok)
 {
-    yCTrace(CONTROLBOARD, "Handling IRemoteCalibrator message");
+    yCTrace(CONTROLBOARD, "Handling IRemoteVariables message");
 
-    if (!rpc_IRemoteCalibrator) {
-        yCError(CONTROLBOARD, "controlBoardWrapper: I do not have a valid IRemoteCalibrator interface");
+    if (!rpc_IVar) {
+        yCError(CONTROLBOARD, "controlBoardWrapper: I do not have a valid IRemoteVariables interface");
         *ok = false;
         return;
     }


Nicogene (approved):

traversaro (approved):

randaz81 (approved):

randaz81:
Just beware, device `ControlBoardWrapper `is deprecated in yarp master (3.8).
I'm going to apply the same fix on a new commit to the new device `controlBoard_nws_yarp` (In yarp master the two devices do not share the same source code).

PeterBowman:
> I'm going to apply the same fix on a new commit to the new device `controlBoard_nws_yarp` (In yarp master the two devices do not share the same source code).

Thank you, @randaz81. I was about to ask: the now deprecated NWS's implementation had been nicely split into multiple files using virtual inheritance ([ref](https://github.com/robotology/yarp/tree/df5de7de644e4643edf333bf93a1e145c8f7da7a/src/devices/controlBoardWrapper)), but this change was lost in the new device ([ref](https://github.com/robotology/yarp/tree/df5de7de644e4643edf333bf93a1e145c8f7da7a/src/devices/controlBoard_nws_yarp)). Is this intentional? I could provide a PR (targetting 3.8.x) to restore that if needed.

If I may also ask: when is YARP 3.8.0 due? I'm hesitant to propose a discussion about reverting https://github.com/robotology/yarp/pull/2841 in 3.7.

Also, I noticed this in current master: https://github.com/robotology/yarp/commit/76fe34dfd5bb6b07557a816116cf48ae35a0e279#r97492632.

randaz81:
> Thank you, @randaz81. I was about to ask: the now deprecated NWS's implementation had been nicely split into multiple files using virtual inheritance ([ref](https://github.com/robotology/yarp/tree/df5de7de644e4643edf333bf93a1e145c8f7da7a/src/devices/controlBoardWrapper)), but this change was lost in the new device ([ref](https://github.com/robotology/yarp/tree/df5de7de644e4643edf333bf93a1e145c8f7da7a/src/devices/controlBoard_nws_yarp)). Is this intentional? I could provide a PR (in the future, targetting 3.8.x) to restore that if needed.
Hi @PeterBowman, the files you are mentioning are no longer required, The logic of the old controlboardwrapper has been split between the nws (only network communication) and another device, the controlBoardremapper (joints map handling). So the implementation of that code you are mentioning has been absorbed by the remapper.

> If I may also ask: when is YARP 3.8.0 due? We were thinking at the end of January, we are already late. We'll do our best to merge the safest and easier PR in this release.

> I'm hesitant to propose a discussion about reverting https://github.com/robotology/yarp/pull/2841 in 3.7.
This is a known issue to me and I know it's important. Let's discuss it (on a separated thread) and let's see if we can do something for this release or postpone it to Yarp 3.8.1 (eventually a couple of weeks later)

> Also, I noticed this in current master: 76fe34d#r97492632.
Thank you! I'll fix it.

PeterBowman:
> The logic of the old controlboardwrapper has been split between the nws (only network communication) and another device, the controlBoardremapper (joints map handling). So the implementation of that code you are mentioning has been absorbed by the remapper.

Oops, my bad, you are totally right, @randaz81. It looks like ControlBoardWrapper.{h/cpp} [here](https://github.com/robotology/yarp/tree/920ecde9a05d2895e9f15f51ada8b7b7ab855501/src/devices/controlBoard_nws_yarp) are no longer used, then.

> Let's discuss it (on a separated thread) and let's see if we can do something for this release or postpone it to Yarp 3.8.1

I will start an issue to cover that topic, but it is not my intention to delay the next release; it is fine for me if the solution arrives at some later point. For now, though, would you mind considering https://github.com/robotology/yarp/pull/2869 for 3.7.x (and also 3.8 by merging the patch into the new device)?

PR #2946, opened 2023-02-22
yarpdataplayer high-rate fix

arrenglover:
This pull request applies a bug-fix and an update for high-frequency to `yarpdataplayer`.

`src/libYARP_dataplayer/src/yarp/dataplayer/YarpDataplayer.cpp#L653`

The function `sendBottle(part, frame)` sends the envelope as a `Bottle(timestamp)` which is inconsistent with `yarpdatadumper` which saves the envelope as a `yarp::os::Stamp` - i.e. sequence number and timestamp. This causes invalid envelopes when reading the envelopes sent by `yarpdatadumper` as a `Stamp`.

The function is changed to `sendGenericData<Bottle>(part, frame)` as it sends the envelope as a `Stamp` by default. In the comments of the code it suggests that the functions are "identical" and therefore this was probably something overlooked, and not intentional.

This bug completely breaks all timestamps for EDPR when using `yarpdataplayer`, but I'm not sure if others have normalised using only timestamps when reading `Bottle`s from `yarpdataplayer`. However, it is inconsistent with all other functions.

`src/libYARP_dataplayer/src/yarp/dataplayer/YarpDataplayer.h#L268`

Update the thread rate of the periodic_thread to have higher temporal resolution - `2 ms` -> `0.1 ms`. In particular for event-camera datasets which can have sub millisecond packet resolution. Higher thread rate did not affect slow datasets, it may increase processing, but the CPU utilisation was `<1%` under both thread rate values. 

 `src/libYARP_dataplayer/src/yarp/dataplayer/YarpDataplayer.cpp`

`src/yarpdataplayer/src/worker.cpp`

`runNormally()` is performed in both these threads for GUI v.s. non-GUI operation. I've updated in both locations to change an 
`if(...)` -> sends only a single frame on any periodic_thread update. Multiple packets can "build-up" for high-frequency datasets or if the periodic_thread is delayed from the OS - to -
`while(..)` -> sends **all** frames of data that have not yet been sent and should have given the `virtualTime`. This change keeps multiple parts in synch much more robustly.

It made sense to restructure the `runNormally()` function to de-clutter the change in logic required to perform this change.

### DIFF
--- a/src/libYARP_dataplayer/src/yarp/dataplayer/YarpDataplayer.cpp
+++ b/src/libYARP_dataplayer/src/yarp/dataplayer/YarpDataplayer.cpp
@@ -650,10 +650,9 @@ void DataplayerWorker::run()
             ret = sendImages(part, frame);
         }
         else if (strcmp(utilities->partDetails[part].type.c_str(), "Bottle") == 0)  {
-            ret = sendBottle(part, frame);
-            // the above line can be safely replaced with sendGenericData<Bottle>.
-            // I kept it for no particular reason, thinking that maybe it could be convenient (later)
-            // to process Bottles in a different way.
+            ret = sendGenericData<Bottle>(part, frame);
+            //the above line could be replaced with sendBottle(part, frame) if 
+            //specifically different behaviour is required for a bottle.
         }
         else if (strcmp(utilities->partDetails[part].type.c_str(), "sensor_msgs/LaserScan") == 0)  {
             ret = sendGenericData<yarp::rosmsg::sensor_msgs::LaserScan>(part, frame);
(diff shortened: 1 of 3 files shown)


arrenglover:
after the conversation with @randaz81 I have further tested the update with the example dataset found here: https://yarp.it/latest/group__yarpdataplayer__example.html which includes images and joint positions.

The GUI update was being performed unnecessarily (30 Hz instead of 3Hz) fast given the changes to the thread rate. It was also coded based on a magic number (`300`) which was dependent on the thread period. I modified this to perform the GUI update at 10 Hz triggered by timestamps instead of # thread updates - so it shouldn't depend on the thread rate. `0.1`is still hardcoded but it is easier to understand the meaning.

everything tested and working.

PR #2979, opened 2023-06-22
[yarpdataplayer] Remove inertial ports for robot and iCubGui from the example app

pattacini:
~~This patch follows up on https://github.com/orgs/robotology/discussions/638#discussioncomment-6248267 and aims to update the names of the inertial ports in the `yarpdataplayer` example application.~~

---

> **Note**
Actually, the `yarpdataplayer` is unable to deal with the MAS client, hence we cleaned up those ports.
See https://github.com/robotology/yarp/pull/2979#issuecomment-1602476624.

### DIFF
--- a/data/yarpdataplayer/scripts/yarpdatadumper.xml.template
+++ b/data/yarpdataplayer/scripts/yarpdatadumper.xml.template
@@ -19,13 +19,6 @@
     </module>
     <module>
         <name>yarpdatadumper</name>
-        <parameters>--name /data/inertial</parameters>
-        <node>computer1</node>
-        <stdio></stdio>
-        <tag>yarpdatadumper</tag>
-    </module>
-    <module>
-        <name>yarpdatadumper</name>
         <parameters>--name /data/torso</parameters>
         <node>computer1</node>
         <stdio></stdio>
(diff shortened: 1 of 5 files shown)


pattacini:
Converted to draft as, while talking to @Nicogene, it turned out that the inertial ports are device-based, hence there is a bidirectional comm that is not handled by `yarpdataplayer`.

If this is the case, we may consider cleaning this example app and getting rid of the part relative to the inertials.

Nicogene:
> Converted to draft as, while talking to @Nicogene, it turned out that the inertial ports are device-based, hence there is a bidirectional comm that is not handled by `yarpdataplayer`.
> 
> If this is the case, we may consider cleaning this example app and getting rid of the part relative to the inertials.

We have to do this cleanup also on icub-main.
I confirm that w/ MASclient is not possible to use `yarpdataplayer`

cc @traversaro

pattacini:
> I confirm that w/ MASclient is not possible to use `yarpdataplayer`

I will then refactor this PR accordingly.

pattacini:
Done! [OP](https://github.com/robotology/yarp/pull/2979#issue-1769018638) updated.

Nicogene (approved):

randaz81 (approved):

PR #2990, opened 2023-07-10
FrameGrabber_nwc_yarp: fix bug that forces images to be sent through RPC calls

davidegorbani:
Together with @traversaro and @HosameldinMohamed, we found out that the `frameGrabber_nwc_yarp` is using an RPC port to get the image of a camera because `m_streamerReceiver` is not defined; we added the function `setStreamReceiver` and in this way, the object `m_streamerReceiver` is defined and we avoid continuously calling the function [getImage](https://github.com/robotology/yarp/blob/99f5a8a2c728b50300d02bada44fe8fc85df5636/src/devices/frameGrabber_nwc_yarp/FrameGrabber_nwc_yarp.cpp#L145C115-L145C115) which uses an RPC port to receive the image and it slows the process.

### DIFF
--- a/src/devices/frameGrabber_nwc_yarp/FrameGrabber_nwc_yarp.cpp
+++ b/src/devices/frameGrabber_nwc_yarp/FrameGrabber_nwc_yarp.cpp
@@ -202,6 +202,10 @@ bool FrameGrabber_nwc_yarp::open(yarp::os::Searchable& config)
         if (!streamReceiver.open(local, remote, carrier)) {
             return false;
         }
+        FrameGrabberOf_ForwarderWithStream<yarp::sig::ImageOf<yarp::sig::PixelRgb>>::setStreamReceiver(&streamReceiver);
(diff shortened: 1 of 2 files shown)


src/devices/frameGrabber_nwc_yarp/FrameGrabber_nwc_yarp.cpp
```diff
@@ -220,6 +220,11 @@ bool FrameGrabber_nwc_yarp::open(yarp::os::Searchable& config)
         yCInfo(FRAMEGRABBER_NWC_YARP) << "No remote specified. Waiting for connection";
     }
 
+    FrameGrabberOf_ForwarderWithStream<yarp::sig::ImageOf<yarp::sig::PixelRgb>>::setStreamReceiver(&streamReceiver);
```
randaz81, line 223: This block of code could be better placed at line ~202-203 after the streamReceiver.open()

randaz81 (changes requested):
see the comment below.

davidegorbani:
Thank you @randaz81 for your comment, I modified the code and added a changelog.

randaz81 (approved):

PR #3010, opened 2023-08-30
Add velocityThres param for torque control

sgiraz:
**What's new:**
- Add a new `velocityThres` parameter used during torque control for friction compensation. This parameter is used to enable the new torque control law and fine-tune the friction compensation based on velocity feedback at FW level.

> **Note**
> - Tested on a single joint setup

### DIFF
--- a/doc/release/yarp_3_9_master/added_velocityThres.md
+++ b/doc/release/yarp_3_9_master/added_velocityThres.md
@@ -1,0 +1,5 @@
+#### `lib_yarp_dev`
+
+* ITorqueControl supports four a new friction parameters: velocityThresh.
+  This parameters is used to enable the new torque control law and fine tune the the friction compensation at FW level.
+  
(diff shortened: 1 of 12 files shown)


CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ endif()
 # Main project
 project(
   YARP
-  VERSION 3.8.1
+  VERSION 3.8.2
```
randaz81, line 20: This should be never modified by a PR. It is handled by the release script

randaz81 (changes requested):

CMakeLists.txt
sgiraz: Done in https://github.com/robotology/yarp/pull/3010/commits/ad556c7cfb072dd58387086edc9791f0137e72f1

randaz81 (approved):

PR #3030, opened 2023-10-18
yarpmotorgui bugfix

sgiraz:
**What's new:**
- when the `PID` button is clicked now it works as expected without crash.

**Note:**
- Tested on `five-small-joint` setup

cc @pattacini @pattacini @AntonioConsilvio @isorrentino


randaz81:
what do you mean with "crash"? in the pr I can see changes just to the ui file

sgiraz:
> what do you mean with "crash"? in the pr I can see changes just to the ui file. And why a button is "ItemIsDragEnabled" ?

Hi @randaz81,

When I edited that .ui file in https://github.com/robotology/yarp/pull/3010, I forgot to apply all the necessary changes. This mistake causes a segmentation fault when you press the `PID` button on the `yarpmotorgui`.

randaz81 (approved):

PR #3034, opened 2023-10-24
log.cpp fix

randaz81:
fix for std::min


randaz81:
fix for #3031

pattacini:
Thanks @randaz81!

PR #3039, opened 2023-11-08
C++20 support

randaz81:
Yarp now requires cpp 20 standard


randaz81:
requires: 
- https://github.com/robotology/yarp/pull/3040
- https://github.com/robotology/yarp/pull/3042

randaz81:
Catch2 fails to compile due to MSVC 2019 bug see: https://github.com/catchorg/Catch2/issues/2174

PR #3065, opened 2023-12-29
Allow expressions in logging component parameters

PeterBowman:
As showcased in the tests, this patch aims to enable the following syntax:

```cxx
yCInfo(quiet ? LOG_COMPONENT_NULL : LOG_COMPONENT) << "message not shown if quiet=true";
```

Currently, due to macro expansion, the expression needs to be wrapped by an additional pair of parentheses:

```cxx
yCInfo((quiet ? LOG_COMPONENT_NULL : LOG_COMPONENT)) << /* ... */;
```


randaz81 (approved):

PeterBowman:
Hello, sorry for writing to a closed thread. Would it be possible to merge yarp-3.9 into master? In my code, I am doing something like this:

```cxx
#if YARP_VERSION_COMPARE(>=, 3, 9, 0)
    yCInfo(quiet ? LC2 : LC) << "message";
#else
    yCInfo((quiet ? LC2 : LC)) << "message";
#endif
```

It fails on the master branch since this patch isn't included, yet in CMake it is stated that the current version is 3.9.0.

Besides, perhaps the current master could be always assigned the next minor version beforehand? That is, set it to 3.10.100 (the 100 denoting an unreleased version). I believe this is how it used to be some time ago, perhaps I missed the change. In this way, it would be possible to safely anticipate a new feature referring to the next version in an `#if`-guard such as the above one.

PR #3074, opened 2024-01-23
Python Embedding Example

SimoneMic:
A fake device (`fakePythonSpeechTranscription`) as example on how to do a Python embedding using C++

There is some commented code that gives insight on the use of some functions, and how to test.
It's provided also a yarprobotinterface example file to ease utilization.

There are two basic idea to take-away from this code:
1- How to call a function in a `.py` module and exchange parameters and objects.
2- How to instantiate a class in Python and call its methods, exchanging parameters and objects.

This is still a draft, done to share ideas


randaz81:
Replaced by #3135

PR #3076, opened 2024-01-30
cmake_minimum_required is now 3.19

randaz81:
cmake_minimum_required is now 3.19


traversaro (approved):

Nicogene (approved):

PeterBowman:
@randaz81 I just noticed that the following lines (and similar) can be deleted now:

https://github.com/robotology/yarp/blob/8d2d4a0f9194f0c54fefea725f0b070e7561dac4/CMakeLists.txt#L7-L15

PR #3085, opened 2024-02-14
ControlBoardCouplingHandler: a device for publishing phyisical joints quantities in under-actuated system

Nicogene:
This is a first draft of the device that will expose the  publishing phyisical joints quantities in under-actuated system using the new `IJointCoupling` interface.

This PR is super draft, it just compiles, I opened the PR for agreeing w/ @traversaro @randaz81 @elandini84 on:
- The device name
- The interfaces that implement
- The interfaces that has as private attribute members

Please ignore both the `test` folder (that has been copied from `controlBoardRemapper`) and the implementation in the .cpp, I will write it once we agree on the points above.

cc @pattacini


src/devices/controlBoardCouplingHandler/ControlBoardCouplingHandler.h
```diff
@@ -0,0 +1,225 @@
...
+ * \endcode
+ *
+ * For compatibility with the controlBoard_nws_yarp, the
+ * networks keyword can also be used to select the desired joints.
+ * For more information on the syntax of the networks, see the
+ * controlBoard_nws_yarp class.
+ *
+ * \code{.unparsed}
+ *  networks (net_larm net_lhand)
+ *  joints 16
```
randaz81, line 62: Is this 16 correct?

src/devices/controlBoardCouplingHandler/ControlBoardCouplingHandler.h
```diff
@@ -0,0 +1,225 @@
...
+ *  net_larm    0 3  0 3
+ *  net_lhand   4 6  0 2
```
randaz81, line 64: can you explain a little better?

src/devices/controlBoardCouplingHandler/ControlBoardCouplingHandler.h
Nicogene: The documentation has been copied from `remoteControlBoardremapper` but it has to be completeley rewritten also because I am not sure which will be the parameters, please ignore it

randaz81:
> * The device name
> * The interfaces that implement
> * The interfaces that has as private attribute members

Name is ok for me. IEncodersTimed and IAxisInfo interfaces are ok. 
Will the device also perform some internal computation (e.g. for differential joints) or it is just a remapping of encoders data?

For the implementation, remember to use the new parser generator https://github.com/robotology/yarp/discussions/3081, I'm already going to replace all manual implementation in all devices in yarp 3.10 https://github.com/robotology/yarp/pull/3084

Nicogene:
> Will the device also perform some internal computation (e.g. for differential joints) or it is just a remapping of encoders data?

My idea was to get the actuated axis quantities via the private attributes interfaces and perform the transformations for getting the physical joints quantities (positions, velocities, accelerations) but I am open to other possibilities

Nicogene:
> > * The device name
> > * The interfaces that implement
> > * The interfaces that has as private attribute members
> 
> Name is ok for me. IEncodersTimed and IAxisInfo interfaces are ok. Will the device also perform some internal computation (e.g. for differential joints) or it is just a remapping of encoders data?
> 
> For the implementation, remember to use the new parser generator #3081, I'm already going to replace all manual implementation in all devices in yarp 3.10 #3084

Hi @randaz81, I was trying to use this nice new feature, but if I got it correctly it does support only these types of params:

| Parameter type | c++ type       | yarp::os::Value call |
|:--------------:|:--------------:|:--------------------:|
| "bool"         | bool           | .asBool()            |
| "string"       | std::string    | .asString()          |
| "int"          | int            | .asInt64()           |
| "size_t"       | size_t         | .asInt64()           |
| "float"        | float          | .asFloat32()         |
| "double"       | double         | .asFloat64()         |

How can we handle the vector/lists ? I generated the parser and then I change by hand using the vectors/lists?
Lists are often used as parameters for devices, are you planning to add it ?

cc @elandini84

Nicogene:
> > > * The device name
> > > * The interfaces that implement
> > > * The interfaces that has as private attribute members
> > 
> > 
> > Name is ok for me. IEncodersTimed and IAxisInfo interfaces are ok. Will the device also perform some internal computation (e.g. for differential joints) or it is just a remapping of encoders data?
> > For the implementation, remember to use the new parser generator #3081, I'm already going to replace all manual implementation in all devices in yarp 3.10 #3084
> 
> Hi @randaz81, I was trying to use this nice new feature, but if I got it correctly it does support only these types of params:
> Parameter type 	c++ type 	yarp::os::Value call
> "bool" 	bool 	.asBool()
> "string" 	std::string 	.asString()
> "int" 	int 	.asInt64()
> "size_t" 	size_t 	.asInt64()
> "float" 	float 	.asFloat32()
> "double" 	double 	.asFloat64()
> 
> How can we handle the vector/lists ? I generated the parser and then I change by hand using the vectors/lists? Lists are often used as parameters for devices, are you planning to add it ?
> 
> cc @elandini84

Ah ok the vectors will be added in:
- #3084 

See:
- https://github.com/robotology/yarp/pull/3084/files#diff-12cc772c34e8f961b26058a5425b0dc3834b4806ee0372cbc30c91cf747a73a7

randaz81:
To all readers, please note that the example .xml shown in  https://github.com/robotology/yarp/pull/3085#issuecomment-1983678503 is using a deprecated format and it should not be considered a good example to take inspiration from.

All list of parameters MUST be always parenthisized i.e.:

 THIS IS AN INVALID LINE:
` <param name="jntVelMax"> 100.0   100.0 100.0 100.0 100.0 100.0  100.0 100.0  100.0 100.0   100.0 100.0 </param>`
 THIS IS THE CORRECT LINE:
` <param name="jntVelMax"> (100.0   100.0 100.0 100.0 100.0 100.0  100.0 100.0  100.0 100.0   100.0 100.0 ) </param>`

the old format is currently supported because of yarp bug which might be fixed in a future release!

randaz81 (approved):

PR #3095, opened 2024-03-18
controlBoardRemapper: align the mandatory interfaces to controlBoard_nws_yarp

Nicogene:
As per `controlBoard_nws_yarp` it is mandatory only that at list is implemented `iAxisInfo` **OR** `iPositionControl`.

### DIFF
--- a/src/devices/controlBoardRemapper/ControlBoardRemapperHelpers.cpp
+++ b/src/devices/controlBoardRemapper/ControlBoardRemapperHelpers.cpp
@@ -174,26 +174,22 @@ bool RemappedSubControlBoard::attach(yarp::dev::PolyDriver *d, const std::string
     // checking minimum set of interfaces required
     if( !(pos) )
     {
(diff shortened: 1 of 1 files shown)


traversaro (approved):

randaz81 (changes requested):
It seems to me that the not-mandatory interfaces are not adequately protected elsewhere.
For example, if `vel` interface is now optional, it must have a guard `if (ivel!=nullptr)` in ControlBoardRemapper.cpp to prevent a segfault if a client is calling it

randaz81:
I'd suggest also to add a new test, which attaches to `fakeMotionControlMicro` instead of `fakeMotionControl`.
This because fakeMotionControl actually implements all interfaces, so the current test cannot catch any issue if the remapper is used with a device that does not implement all interfaces.

Nicogene:
@randaz81 @traversaro I added all the safeguards for those interfaces that are not mandatory anymore.
Moreover, I added a test on the `fakeMotionControlMicro` in order to check these new changes, finally I successfully tested the `controlBoardRemapper` in simulation.

traversaro (approved):

PR #3100, opened 2024-03-29
Improve closing behavior when attempting to close a device implementing the service interface

S-Dafarra:
I am implementing a device which can close prematurely depending on the internal state. I want ``yarpdev`` to be notified when this happens and close consequently.

In order to have this behavior, I implemented the ``IService`` interface. But then, if I press CTRL+C in the ``yarpdev`` terminal, it seems to always get stuck, and then I have to kill the ``yarpdev`` process. The last message was
```
[INFO] |yarp.dev.Drivers| [try 1 of 3] Trying to shut down /yarpdev/quit
```

After some debugging, I noticed that when the signal handler was setting the static flag "terminated" to true, it was also attempting to close immediately the terminator port. But the terminator port is receiving the termination message and some race condition seemed to occur on Windows.

If we avoid to set the ``terminated`` flag from the signal handler and instead we wait for the termination port to do its job, then ``yarpdev`` closes smoothly without issues.

cc @randaz81 @Nicogene @traversaro

### DIFF
--- a/src/libYARP_dev/src/yarp/dev/Drivers.cpp
+++ b/src/libYARP_dev/src/yarp/dev/Drivers.cpp
@@ -325,7 +325,6 @@ DriverCreator* Drivers::Private::load(const char *name) {
 }
 
 static std::string terminatorKey;
-static bool terminated = false;
 static void handler (int)
 {
     Time::useSystemClock();
@@ -343,7 +342,6 @@ static void handler (int)
     }
     if (!terminatorKey.empty()) {
         yCInfo(DRIVERS, "[try %d of 3] Trying to shut down %s", ct, terminatorKey.c_str());
-        terminated = true;
         Terminator::terminateByName(terminatorKey.c_str());
     } else {
         yCInfo(DRIVERS, "Aborting...");
@@ -529,6 +527,7 @@ int Drivers::yarpdev(int argc, char *argv[]) {
             service = nullptr;
         }
     }
+    bool terminated = false;
     while (dd.isValid() && !(terminated||(terminee&&terminee->mustQuit()))) {
         if (service!=nullptr) {
             double now = Time::now();


camillagallina:
Hello @randaz81 @Nicogene 
Did you have any chance to check this?
Thank you😊

randaz81:
checking.. I do not remember anymore :-)

randaz81:
The main issue here is that I was hesitant to make this change because I was concerned about introducing regressions elsewhere, especially since this PR is not covered by the CI.

There are two points I'd like to raise:
- There is currently no example device in YARP that uses the `IService` interface. It would be very helpful to either provide a reference to an existing device that uses it or add a simple mock device that demonstrates and tests its functionality.
- Once a dedicated test (for example, `harness_device_iServiceDevice`) is available, it will be much easier to assess the impact of this PR and verify that the proposed changes do not introduce unintended side effects.

Would you be willing to extend this PR by adding a test device?

S-Dafarra:
Thanks @randaz81!

Do you have any example of testing the closure mechanism of any yarp device?

randaz81:
Do you mean how to simulate the CTRL+C you mention in the PR description?
I'm not an expert on this, but I'd probably start trying one of the following methods that I found around in the yarp code:
```
  yarp::os::impl::raise(SIGINT);
  yarp::os::impl::raise(SIGBREAK);
  yarp::os::impl::raise(SIGTERM);
```
not SIGKILL... obviously :-)

randaz81:
This example looks perfect to be copied. This is the source code of a template device (which could implement the interface you want)
https://github.com/robotology/yarp/tree/master/src/devices/fake/fakeDeviceUnwrapped
and its test:
https://github.com/robotology/yarp/blob/master/src/devices/fake/fakeDeviceUnwrapped/tests/fakeDeviceUnwrapped_test.cpp

Following this template will automatically generate a new test in the CI. 
Hopefully, the CTRL_C event will be intercepted by the device (and not by the test suite while running on GitHub?) To be verified. In any case, having the example would be beneficial even if it does not work on the Github CI.

S-Dafarra:
> Do you mean how to simulate the CTRL+C you mention in the PR description?

Sort of. I was wondering if there was any existing test on the termination behavior of a device when sending an asynchronous SIGINT signal (i.e. pressing CTRL-C in the terminal).

I noticed that when using the ``IService`` interface there seems to be a race condition causing a deadlock. I would be happy to add a test case for this specific situation, but I would start from something existing

randaz81:
Any update on this?
Otherwise, I'll proceed with adding a test by myself

S-Dafarra:
> Otherwise, I'll proceed with adding a test by myself

That would be great, thanks a lot! I was confused about where to start

PR #3101, opened 2024-04-08
Allow JoypadControlServer to run single thread

S-Dafarra:
I developed a device that can emulate a joypad using a GUI (https://github.com/ami-iit/yarp-device-keyboard-joypad). Since on macOS the GUIs can be opened only from the main thread, I needed to allow the possibility of avoiding running the ``JoypadControlServer`` in a single threaded mode.

In order to do so, I had to implement the ``IService`` interface on both ``JoypadControlServer`` and ``DeviceBundler``. This PR also includes the modifications of https://github.com/robotology/yarp/pull/3100

cc @randaz81 @traversaro @Nicogene

### DIFF
--- a/doc/release/master.md
+++ b/doc/release/master.md
@@ -29,6 +29,7 @@ Fixes
 
 * Configuration files installed by the `yarp_configure_plugins_installation` CMake macro are now relocatable (https://github.com/robotology/yarp/issues/2445, ).
 * Improved `ffmpeg` port monitor to allow using different couples of coders/decodes
+* Improved `yarpdev` closing behavior for devices implementing the `iService` interface
 
 New Features
 ------------
--- a/src/devices/deviceBundler/DeviceBundler.cpp
+++ b/src/devices/deviceBundler/DeviceBundler.cpp
@@ -76,6 +76,12 @@ bool DeviceBundler::open(yarp::os::Searchable& config)
         return false;
     }
 
+    ret = m_pdev_wrapper.view(m_iService);
+    if (ret)
+    {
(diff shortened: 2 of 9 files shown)


traversaro (dismissed):

randaz81:
@S-Dafarra yarp already has a GUI with keybpord support to control the 2D direction of a robot: yarpmobilebasegui.
Can you attach a screenshot and maybe highlight the differences?

S-Dafarra:
> @S-Dafarra yarp already has a GUI with keybpord support to control the 2D direction of a robot: yarpmobilebasegui. Can you attach a screenshot and maybe highlight the differences?

See https://github.com/ami-iit/yarp-device-keyboard-joypad/pull/3#issuecomment-2041073183

I was not really aware of yarpmobilebasegui. I needed a drop in replacement for the SDLJoypad with a good degree of customization on the axis and buttons order

randaz81:
The main problem here is that `yarp::dev::IService` was marked in my notes: "critically flawed, to be deprecated". However, I currently do not remember the exact reason. Please allow me some time to think about this.

randaz81 (changes requested):
on hold

S-Dafarra:
I realized that in the single thread case I was not updating the variable with the last run time. Fixed with https://github.com/robotology/yarp/pull/3101/commits/b060a6e42d4eca026e030fd57a549789799a9414

randaz81:
Removed in Yarp4, replaced by `JoypadControl_nws_yarp` with a completely different implementation

Issue #3105, opened 2024-04-17
Possible access to dangling pointer in yarpbroker.cpp


S-Dafarra:
**Describe the bug**
While reading the code to understand the issue behind https://github.com/robotology/yarp/issues/3104, I came across a warning by the Visual Studio static analyzer.

In particular, the following lines
https://github.com/robotology/yarp/blob/e4762b0b4a7946bf9379b183379bd32353d54a26/src/libYARP_manager/src/yarp/manager/yarpbroker.cpp#L566-L576

return a pointer to the char buffer of a temporary string.

**Configuration (please complete the following information):**
 - OS: 
 - yarp version: 3.9.0
-  compiler:

**Additional context**

cc @randaz81

randaz81:
Probably fixed. Can you give it a check @S-Dafarra ?

PR #3112, opened 2024-06-04
fixed dangling pointer using std::string

randaz81:
Fix for: https://github.com/robotology/yarp/issues/3105
I'm aware that std::string is not a plain char*. 
..but do we need to stick to this oooooold '99 style in libyarpmanager? I think not.  
BTW, strings are used elsewhere in the same lib.

### DIFF
--- a/src/libYARP_manager/src/yarp/manager/yarpbroker.cpp
+++ b/src/libYARP_manager/src/yarp/manager/yarpbroker.cpp
@@ -529,21 +529,21 @@ bool YarpBroker::exists(const char* szport)
     return NetworkBase::exists(szport, style);
 }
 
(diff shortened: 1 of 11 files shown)


Nicogene (dismissed):

src/libYARP_manager/src/yarp/manager/yarpbroker.cpp
```diff
@@ -569,14 +568,14 @@ const char* YarpBroker::requestRpc(const char* szport, const char* request, doub
     NetworkBase::disconnect(port.getName(), szport);
     if(!response.size() || !ret) {
         port.close();
-        return nullptr;
+        return {};
     }
 
     port.close();
     return response.toString().c_str();
```
S-Dafarra, line 575: Is the ``c_str()`` needed here?

PR #3137, opened 2024-09-13
Fixed installation of yarp_sig header files

randaz81:
Added install of header files generated by thtift in yarp_sig

### DIFF
--- a/src/libYARP_dev/src/CMakeLists.txt
+++ b/src/libYARP_dev/src/CMakeLists.txt
@@ -288,10 +288,13 @@ foreach(_f IN LISTS YARP_dev_idl_GEN_FILES)
   endif()
 endforeach()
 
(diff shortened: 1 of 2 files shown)


PeterBowman:
https://github.com/robotology/yarp/blob/a53da60ada23e9741b5f0635e2250f602bd6af91/src/libYARP_sig/src/CMakeLists.txt#L192

@randaz81 shouldn't we also add private IDL headers, as in YARP_dev?

```cmake
set_property(TARGET YARP_sig PROPERTY PRIVATE_HEADER ${YARP_sig_IMPL_HDRS} ${YARP_sig_idl_IMPL_HDRS}) 
```

PeterBowman:
Would it be OK if I open a PR to patch this? It doesn't seem to cause trouble now, but I believe it might explode at some point later, and such tiny things are prone to being hard to debug.

PR #3177, opened 2025-02-12
Introduce IJointCoupling::evaluateCoupledJointsJacobian()

PasMarra:
This PR introduces `IJointCoupling::evaluateCoupledJointsJacobian()`, whose implementation should calculates the Jacobian mapping the Actuated Axes to Physical Joints velocity

### DIFF
--- a/src/libYARP_dev/src/yarp/dev/IJointCoupling.h
+++ b/src/libYARP_dev/src/yarp/dev/IJointCoupling.h
@@ -9,6 +9,7 @@
 #include <yarp/dev/api.h>
 #include <yarp/os/Vocab.h>
 #include <yarp/sig/Vector.h>
+#include <yarp/sig/Matrix.h>
 
 namespace yarp::dev {
 class IJointCoupling;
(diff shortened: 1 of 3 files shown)


Nicogene:
> Use a name more compatible with existing nomenclature, so something like evaluateJacobianFromActuatedAxesToPhysicalJointsVel ? In this way we can be ready if necessary to introduce other methods in the future (for example evaluateJacobianFromPhysicalJointsToActuatedAxeseVel).

I was wondering if it makes sense to already add `evaluateJacobianFromPhysicalJointsToActuatedAxeseVel`

randaz81:
I think these methods should be moved to a new interface: 
`IJacobianXXX` as virtual abstract.

- implementJointCoupling file will then not need any modification.

- The device that will have these capabilities will inherit from `IJacobianXXX`

- The user code will attempt to view the new interface. If not implemented the ->view() method will return false, allowing the user to detect if the device implements the interface or not.

randaz81 (approved):

PR #3188, opened 2025-03-06
Add Motor Temperature to stateExt and Convert PWM to Float

GiulioRomualdi:
After a discussion with @randaz81 and @traversaro, we decided to add motor temperature data to `stateExt`, streamed as a vector of floats. To avoid increasing the message size, we also decided to convert the PWM data from a vector of doubles to a vector of floats.  

This pull request introduces several changes to the `RemoteControlBoard` and related components to support temperature data. The modifications affect data structures, methods, and interfaces to handle temperature readings properly.  

The PR consists of two commits:  
- https://github.com/robotology/yarp/commit/e889527bc308ef3636dc727a414619af80399c2b: Modifies the message to add temperature data and converts the PWM vector to float.  
- https://github.com/robotology/yarp/commit/a435ecd640f4324df705e75c08a6a2ef568e596d: Adds support for `float32` conversion in `Bottle`. Without this, reading from the `stateExt` port results in the following error:  
  ```
  [ERROR]  |yarp.os.impl.BottleImpl| Reader failed, unrecognized object code 0
  ```

> [!NOTE]
> I also regenerated the `jointData.cpp` and `jointData.h` to be compliant with the new thrift

### DIFF
--- a/src/devices/networkWrappers/RemoteControlBoard/RemoteControlBoard.cpp
+++ b/src/devices/networkWrappers/RemoteControlBoard/RemoteControlBoard.cpp
@@ -1542,12 +1542,23 @@ bool RemoteControlBoard::getNumberOfMotors(int *num)
 
 bool RemoteControlBoard::getTemperature      (int m, double* val)
 {
-    return get1V1I1D(VOCAB_TEMPERATURE, m, val);
+    double localArrivalTime = 0.0;
+
+    extendedPortMutex.lock();
+    bool ret = extendedIntputStatePort.getLastSingle(m, VOCAB_TEMPERATURE, val, lastStamp, localArrivalTime);
+    extendedPortMutex.unlock();
+    return ret;
 }
 
 bool RemoteControlBoard::getTemperatures     (double *vals)
 {
-    return get1VDA(VOCAB_TEMPERATURES, vals);
+    double localArrivalTime = 0.0;
+
+    extendedPortMutex.lock();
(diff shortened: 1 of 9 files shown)


GiulioRomualdi:
> [!NOTE]  
> Since this PR will not be merged into YARP soon (the next YARP release is expected in the coming months), I have already prepared two branches that can be used in YARP to collect the temperature using this feature:  
> - https://github.com/ami-iit/yarp/tree/yarp-3.10.1-motor-temperature (based on YARP 3.10.1)  
> - https://github.com/ami-iit/yarp/tree/yarp-3.10-motor-temperature (based on YARP 3.10)  
>  
> Please note that this modification is a breaking change, as it alters the structure of `stateExt`. If you want to use this feature, ensure that all machines in the network involved in the experiment are switched to the same branch, recompiled, and configured accordingly. After the experiment, switch back to the original configuration.

GiulioRomualdi:
cc @valegagge @MSECode

valegagge:
Do we need to update the [robometry documentation](https://github.com/robotology/robometry?tab=readme-ov-file#telemetrydevicedumper) after this merge?
cc @Nicogene @MSECode

MSECode:
> Do we need to update the [robometry documentation](https://github.com/robotology/robometry?tab=readme-ov-file#telemetrydevicedumper) after this merge? cc @Nicogene @MSECode

It was already updated. We probably just need to remove the warning about the problem of working on a specific yarp branch. So yes.

PR #3190, opened 2025-03-12
RGBDSensorClient: fix setRgbFov

Nicogene:
As per title


randaz81:
NOTE: Already included in https://github.com/robotology/yarp/pull/3184

randaz81:
Merged because https://github.com/robotology/yarp/pull/3184 is going to be postponed to the next release

PR #3212, opened 2025-03-25
Added methods canTransform and frameExists in bindings.

S-Dafarra:
After https://github.com/robotology/yarp/pull/3169 these two methods could not be used anymore in the python bindings since they edit an input parameter by reference


randaz81:
To be merged after https://github.com/robotology/yarp/pull/3210

randaz81 (approved):

traversaro (approved):

PR #3245, opened 2025-06-27
ControlBoardRemapper: fix wrong buffer resize in resizeSubControlBoadBuffers

Nicogene:
It fixes:
- https://github.com/robotology/gz-sim-yarp-plugins/issues/260

cc @PasMarra


traversaro (approved):
Nice catch!

randaz81 (approved):

PR #3257, opened 2025-07-21
Added new interface IJointBrake

randaz81:
Added new interface IJointBrake


src/libYARP_dev/src/yarp/dev/IJointBreak.h
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * @ingroup dev_iface_motor
+ *
+ * Interface for controlling a joint equipped with brakes (hardware or simulated).
+ */
+class YARP_dev_API yarp::dev::IJointBrake
+{
+public:
+    virtual ~IJointBreak(){}
```
traversaro:
```suggestion
    virtual ~IJointBrake(){}
```

src/libYARP_dev/src/yarp/dev/IJointBreak.h
```diff
@@ -0,0 +1,78 @@
...
+    * @param j joint number
+    * @param braked is true if the joint is currently braked
+    * @return: true/false success failure.
+    */
+    virtual yarp::dev::ReturnValue isJointBraked(int j, bool& braked) const =0;
+
+    /**
+    * Enables/Disables manually the joint brake.
+    * @param j joint number
+    * @param active true to enable the joint break
```
traversaro:
```suggestion
    * @param active true to enable the joint brake
```

traversaro:
From what understood from our f2f conversation, the idea was to use the `isJointBraked` to distinguish between a Hardware fault state in which the joint is braked, from a hardware fault that is not braked. In the particular implementation we had in mind, I am not sure sure how we could implement the rest of the `set` methods, it make sense to return `ReturnValue(return_value_error_not_implemented_by_device)` on all the other methods, and just implement `isJointBraked` ? 

fyi @S-Dafarra @GiulioRomualdi @isorrentino @valegagge

src/libYARP_dev/src/yarp/dev/IJointBrake.h
```diff
@@ -0,0 +1,78 @@
...
+    /**
+    * Check is the joint brake is currently active.
+    * @param j joint number
+    * @param braked is true if the joint is currently braked
+    * @return: true/false success failure.
+    */
+    virtual yarp::dev::ReturnValue isJointBraked(int j, bool& braked) const =0;
+
+    /**
+    * Enables/Disables manually the joint brake.
```
isorrentino, line 36: Does it make sense to add more details about the expected behavior of brakes? For instance, what happens if one (manual or automatic) is set while the other is active?

src/libYARP_dev/src/yarp/dev/IJointBrake.h
```diff
@@ -0,0 +1,78 @@
...
+    * @return: true/false success failure.
+    */
+    virtual yarp::dev::ReturnValue setManualBrakeActive(int j, bool active) =0;
+
+    /**
+    * Enables/Disables the automatic joint brake. If enabled, the firmware
+    * can automatically choose when to enable the joint brake (for example when
+    * the joint is in hardware_fault mode).
+    * @param j joint number
+    * @param active true to enable the automatic joint brake
```
isorrentino: The “@param active” should be “@param enabled” to match the parameter name.

Nicogene (commented):
I am not sure about the set methods, because I am not sure the user should enable/disable the break.

Maybe it makes sense to add also the functions that acts on multiple joints, e.g.

```c++
yarp::os::ReturnValue areJointsBraked(int* axis, bool* braked_vec); 
```

randaz81:
@isorrentino

>  Does it make sense to add more details about the expected behavior of brakes? For instance, what happens if one (manual or automatic) is set while the other is active?

@Nicogene 

> I am not sure about the set methods, because I am not sure the user should enable/disable the break.

In this YARP interface, I'm trying to define some constraints while keeping it as general as possible—i.e., leaving the actual behavior to the specific firmware implementation. This approach is consistent with many other interfaces; for example, a position interface typically just exposes setPosition methods, without specifying the internal controller logic.

---
Regarding our robot, I can think of something simular to:

`isJointBraked` returns true if the joint is currently braked.
`set/getAutoBrakeEnabled` cannot be implemented
`setManualBrakeActive` cannot be implemented 
OR
`set/getAutoBrakeEnabled`  enables the firmware functionality. If enabled, the joint in hw fault is braked, otherwise the joint is free.
`setManualBrakeActive` cannot be implemented 

---
On a different robot (e.g. industrial manipulator) we can have, instead:

`setManualBrakeActive`  enable/disable the brake on request user
`set/getAutoBrakeEnabled`  cannot be implemented 
OR
`setManualBrakeActive`  enable/disable the brake on request user
`set/getAutoBrakeEnabled`  if enabled, the controller can activate the brake on its own decision (but not simultaneously with setManualBrakeActive)

These are just a few possibilities....

randaz81:
Added the label `braked` to the yarpmotorgui. The image is purely indicative, since it addresses the most general case, showing a braked joint while in position mode. In reality, in our firmware setup, it will be possible to have braked joints only when in `hwfault`.
image

To perform the test:
```
yarp server
yarpdev --device fakeMotionControl --name /robot/aaa
--robot robot --parts "(aaa)" 

yarp rpc  /robot/aaa/nws/rpc
setManualBrakeActiveRPC 0 1
```

PR #3259, opened 2025-07-24
Re-enable check for torque enable

MSECode:
This PR fix a bug that we noticed on the motorgui.
The enabling thick for the torque mode was not visible anymore. 

This fixes a regression introduced in:
- #3242


Nicogene:
Hi @randaz81 we managed to fix the regression you found after:
- #3242 

I tested it on gz-sim and seems fine, if you want we can review it on the setup

randaz81 (approved):

PR #3337, opened 2026-05-26
Quick status refresh for yarprun and yarpmanager console 

fincatomarta:
This PR improves the yarpmanager console refresh logic.
When refreshing multiple modules running on the same yarprun server, the client now requests the list of running processes only once per server, instead of issuing one request per module. This makes the refresh operation more efficient, significantly reducing the number of calls and the overall time required to update the console state.


Nicogene (commented):
I am out of YARP/yarpmanager development since some time, I didn't checked the implementation details but philosophically I agree with the idea of this implementation

randaz81:
Some comments from the Copilot reviewer will be addressed in an upcoming PR

PR #3353, opened 2026-07-05
FrameGrabberControls: propagate device operation results for SET RPCs

### LINKED ISSUE #3352: FrameGrabberControls` RPC  ` ` SET `  commands silently report success even when the underlying device call fails (yarp-3.12)
**Describe the bug**
`FrameGrabberControls_Responder`/`Forwarder` (the RPC pair backing
`IFrameGrabberControls::setFeature`, `setFeature` (2-arg), `setActive`,
`setMode`, and `setOnePush`, used by `RGBDSensorClient` and
`frameGrabber_nwc_yarp`) do not propagate the underlying device's actual
success/failure result back to the caller.

`FrameGrabberControls_Responder::respond()` correctly computes `ok` from the
real device call (e.g. `fgCtrl->setFeature(...)`), but never writes it into
the response `Bottle` for `VOCAB_SET` commands. On the client side,
`FrameGrabberControls_Forwarder::setFeature()` (and the other four SET style
methods) return `m_port.write(cmd, response)` directly, which only reflects
whether the RPC round-trip completed, not whether the device accepted the
request.

Net effect: application code can receive `true` from `setFeature()`/
`setActive()`/`setMode()`/`setOnePush()` even when the underlying device's
own implementation returned `false`.

Found while investigating #3304 (`RGBDSensorClient::setFeature()` reportedly
returning `true` without changing the camera). I could not find any code
change between v3.10.0 and v3.12.1 in the actual `setFeature()` call path
that would explain a version-to-version regression there, so this bug isn't
itself a regression, it's present identically in both versions but it is
a real defect that would produce the exact symptom described in #3304 (RPC
reports success, device silently didn't apply the change). See #3304 for
the original report and additional context.

The sibling protocols in the same file family, `RgbVisualParams` and
`DepthVisualParams`, already correctly encode and read back the device's
real result, confirming this is an omission specific to
`FrameGrabberControls`, not an intentional design choice. Also checked
`FrameGrabberControlsDC1394` (uses a different but internally consistent
convention, no bug) and `FrameGrabberOf` (no `SET` implementation at all).

**To Reproduce**
1. Build `frameGrabber_nwc_yarp` talking to `frameGrabber_nws_yarp`, wrapping
   `fakeFrameGrabber` (whose `setFeature`/`setActive`/`setMode`/`setOnePush`
   stubs always return `false`).
2. Call, e.g.:
   ```cpp
   IFrameGrabberControls* ictrl = ...;
   bool ok = ictrl->setFeature(0, 1.0);
   ```
3. Observe `ok == true`, even though `fakeFrameGrabber::setFeature()`
   returned `false`.

**Expected behavior**
`setFeature()`/`setActive()`/`setMode()`/`setOnePush()` should return
`false` whenever the underlying device's own implementation returns
`false`, matching the existing behavior of `setRgbResolution()`/
`setRgbFOV()`/etc. in `RgbVisualParams`.

**Screenshots**
None (protocol/logic bug, not visual).

- OS: N/A (reproducible on any platform; not environment-specific)
- yarp version: yarp-3.12 branch (present in v3.12.1, and likely earlier
  3.x releases using the same `framegrabber_protocol` code not checked
  further back than v3.10.0)
- compiler: N/A

**Additional context**
Not present on `master` : this protocol was replaced entirely by a
Thrift/`ReturnValue`-based RPC in PR #3184, which correctly propagates the
result. This issue and an accompanying PR are scoped to the `yarp-3.12`
stable branch only.

I have a fix ready (mirrors the existing `RgbVisualParams_Responder`/
`Forwarder` pattern) and a regression test using `fakeFrameGrabber`

will open a PR referencing this issue shortly.

cc @randaz81

gauraaansh:
### Description
`FrameGrabberControls_Responder`/`Forwarder` (the RPC pair backing
`setFeature`, `setActive`, `setMode`, `setOnePush` on `IFrameGrabberControls`)
currently discards the real result of the underlying device call:

- `FrameGrabberControls_Responder::respond()` computes `ok` from the actual
  device call (e.g. `fgCtrl->setFeature(...)`), but never writes it into the
  response `Bottle` for `VOCAB_SET` commands.
- `FrameGrabberControls_Forwarder::setFeature()` (and `setFeature2`,
  `setActive`, `setMode`, `setOnePush`) return `m_port.write(cmd, response)`
  directly, which reflects "did the RPC round-trip complete," not "did the
  device accept the request."

**Net effect:** these calls return the success of the RPC transport rather
than the result of the underlying device operation.

| Scenario                | Before  | After   |
|-------------------------|---------|---------|
| Device returns `true`   | `true`  | `true`  |
| Device returns `false`  | `true`  | `false` |

The sibling protocols in the same family, `RgbVisualParams` and
`DepthVisualParams` (`setRgbResolution`, `setRgbFOV`, `setRgbMirroring`,
`setDepthResolution`, etc.), already encode and read back the real result
correctly. This PR brings `FrameGrabberControls` in line with that existing,
working pattern with no new protocol convention is introduced.

### Testing
- `harness_device_frameGrabber_nwc_yarp`: all 170 assertions pass (both RPC
  and streaming-mode sections); confirmed 10 of them fail without this fix
  (reverted the fix, rebuilt, reran all 5 new assertions failed with
  `true == false`).
- `harness_device_RGBDSensorClient`: all 12 assertions pass unchanged.

### Why
Fixes #3352. 

This issue does not affect `master`, where the protocol has already been
replaced by the Thrift/`ReturnValue` implementation introduced in #3184.

### Changes
- `FrameGrabberControls_Responder.cpp`: each `VOCAB_SET` case now adds
  `[VOCAB_FRAMEGRABBER_CONTROL, VOCAB_SET, ok]` to the response, matching
  `RgbVisualParams_Responder`'s convention exactly.
- `FrameGrabberControls_Forwarder.cpp`: `setFeature`, `setFeature` (2-arg),
  `setActive`, `setMode`, `setOnePush` now read `response.get(2).asBool()`
  instead of returning transport-write success.
- Added a regression test in `frameGrabber_nwc_yarp_test.cpp` asserting these
  five calls return `false` when talking to `fakeFrameGrabber` (whose stub
  implementations of these methods always return `false`).

### Compatibility note
This changes the wire response shape for `VOCAB_SET` framegrabber-control
commands from empty to a 3-element Bottle. The only in-tree consumers of
`FrameGrabberControls_Forwarder` are `RGBDSensorClient` and
`frameGrabber_nwc_yarp`, both updated/tested together here. Any out-of-tree
client communicating directly with this RPC protocol (rather than using
`FrameGrabberControls_Forwarder`) may need to account for the updated
response shape.

### DIFF
--- a/doc/release/yarp_3_12/fix_framegrabbercontrols_setfeature_ack.md
+++ b/doc/release/yarp_3_12/fix_framegrabbercontrols_setfeature_ack.md
@@ -1,0 +1,12 @@
+fix_framegrabbercontrols_setfeature_ack {#yarp_3_12}
+-----------------------
+
+## Devices
+
+### `frameGrabber_nwc_yarp`, `RGBDSensorClient`
+
+* Fixed `FrameGrabberControls_Responder`/`Forwarder` silently reporting RPC
+  transport success instead of the underlying device operation result for
+  `setFeature`, `setActive`, `setMode`, and `setOnePush`. These calls now
+  correctly return `false` when the device rejects the request, matching
+  the existing `RgbVisualParams`/`DepthVisualParams` behavior (#3352).
--- a/src/devices/messages/framegrabber_protocol/yarp/proto/framegrabber/FrameGrabberControls_Forwarder.cpp
+++ b/src/devices/messages/framegrabber_protocol/yarp/proto/framegrabber/FrameGrabberControls_Forwarder.cpp
@@ -56,7 +56,8 @@ bool FrameGrabberControls_Forwarder::setFeature(int feature, double value)
     cmd.addVocab32(VOCAB_FEATURE);
     cmd.addInt32(feature);
     cmd.addFloat64(value);
-    return m_port.write(cmd, response);
+    m_port.write(cmd, response);
+    return response.get(2).asBool();
 }
 
 bool FrameGrabberControls_Forwarder::setFeature(int feature, double value1, double value2)
@@ -69,7 +70,8 @@ bool FrameGrabberControls_Forwarder::setFeature(int feature, double value1, doub
     cmd.addInt32(feature);
     cmd.addFloat64(value1);
     cmd.addFloat64(value2);
-    return m_port.write(cmd, response);
+    m_port.write(cmd, response);
+    return response.get(2).asBool();
 }
 
 bool FrameGrabberControls_Forwarder::getFeature(int feature, double* value)
@@ -124,7 +126,8 @@ bool FrameGrabberControls_Forwarder::setActive(int feature, bool onoff)
     cmd.addVocab32(VOCAB_ACTIVE);
     cmd.addInt32(feature);
     cmd.addInt32(onoff);
-    return m_port.write(cmd, response);
+    m_port.write(cmd, response);
+    return response.get(2).asBool();
 }
 
 bool FrameGrabberControls_Forwarder::getActive(int feature, bool* isActive)
@@ -192,7 +195,8 @@ bool FrameGrabberControls_Forwarder::setMode(int feature, FeatureMode mode)
     cmd.addVocab32(VOCAB_MODE);
     cmd.addInt32(feature);
     cmd.addInt32(mode);
-    return m_port.write(cmd, response);
+    m_port.write(cmd, response);
+    return response.get(2).asBool();
 }
 
 bool FrameGrabberControls_Forwarder::getMode(int feature, FeatureMode* mode)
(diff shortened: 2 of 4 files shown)


gauraaansh:
CI is failing across some jobs, but this looks pre-existing , the yarp-3.12 base commit (1ab876f36)  failed CI on its own push before this PR was made, and the failures I checked happen at the CMake Configure step, before Build  runs. 
Passing locally: harness_device_frameGrabber_nwc_yarp (170/170) and harness_device_RGBDSensorClient (12/12).

Happy to make any changes if required.

PR #3363, opened 2026-08-06
Thrift generator fix [from fork: test]

randaz81:
Fixes issue preventing the correct generation of .thrift messages containing complex data type, e.g. list<list<double>> etc.


randaz81:
replaced by: https://github.com/robotology/yarp/pull/3364

PR #3368, opened 2026-08-16
Reject misspelled sensor-name options in MAS remapper

fallenmi:
Fixes #3306.

## Summary

- Reject unknown top-level sensor-selection options that still use the documented `Names` suffix.
- List the valid `<sensorTag>Names` keys in the error log.
- Preserve unrelated YARP and `PolyDriver` options.

This prevents misspelled sensor tags such as `LinearVelocitySensorNames` from being silently ignored while the remapper opens with no matching configured sensor.

## Testing

- Reproduced the failure on `yarp-4.0`: the new regression failed because `open()` returned `true`.
- `ctest --test-dir build -R 'multipleanalogsensorsremapper' --output-on-failure` — 1/1 passed (21 assertions).
- YARP style and license checks passed.
- clang-format 20 changed-line check and `git diff --check` passed.

## AI assistance disclosure

OpenAI Codex prepared the issue reproduction, implementation, tests, and validation. This PR remains a draft for the account owner's final review.


randaz81:
MAS remapper does not use the generated parameter parser like the other devices.
We will evaluate whether the whole parser can be replaced.

PR #3369, opened 2026-08-17
Fix compilation with SWIG 4.5

traversaro:
SWIG 4.5 dropped legacy compat functions between Python 2 and Python 3, so `PyInt_AsLong` does not compile anymore. We can just use `PyInt_AsLong` that is the correct Python 3 function to call.


randaz81:
Hi @traversaro  Could you please rebase this PR?
I recently fixed an issue on master with a force push, which broke the history.

randaz81:
@traversaro I cannot push force on your branch. Could you rebase commit [1c877c3](https://github.com/robotology/yarp/pull/3369/commits/1c877c369b25d7bf8e905f0620cf6038e29bd801)?

PR #3375, opened 2026-08-25
ControlBoardRemapper param parser cleanup and refactor

randaz81:
In `controlboardRemapper`: dropped the old deprecated logic (inherited by the old controlboardwrapper device) supporting the indices to define the remapping logic; i.e., this syntax is no more legit:
```
 *  networks (net_larm net_lhand)
 *  joints 16
 *  net_larm    0 3  0 3
 *  net_lhand   4 6  0 2
```
Now only axis names (`axesNames` parameter) can be used, i..e.
```
axesNames (joint1 joint2 joint3)
```
This syntax is cleaner and error-proof. In the previous syntax, that was already deprecated but unfortunately still used, changing something in the device board can invalidate the indices of the remapping device. Moreover, a mismatch in the joint values can lead to segfaults or undefined behavior.

---

### This is a breaking change for all robot configuration files that have not adopted the new logic yet.

---

ControlBoardRemapper and RemoteControlBoardRemapper were refactored to support ParamParser.


Nicogene:
cc @valegagge that is working on robots-configuration porting

cc @pattacini

Issue #3370, opened 2026-08-20
Purpose of `yarp::dev::IJoypadController::getStickDoF` after latest refactor


PeterBowman:
PR https://github.com/robotology/yarp/pull/3336 heavily refactored [`yarp::dev::IJoypadController`](https://github.com/robotology/yarp/blob/65e90e42a5c8cc3a43e87b54191a5f704cb06622/src/libYARP_dev/src/yarp/dev/IJoypadController.h). Among other changes, the `getStick` method no longer accepts a `yarp::sig::Vector`, preferring the new `StickData` Thrift-generated message structure instead. It seems to be generated twice:

- src/libYARP_dev/src/idl: [.thrift](https://github.com/robotology/yarp/blob/fcef5d167b761fbb8e578c395ec24808d759fd45/src/libYARP_dev/src/idl/JoypadData.thrift#L27-L34), [.h](https://github.com/robotology/yarp/blob/fcef5d167b761fbb8e578c395ec24808d759fd45/src/libYARP_dev/src/idl_generated_code/yarp/dev/StickData.h), [.cpp](https://github.com/robotology/yarp/blob/fcef5d167b761fbb8e578c395ec24808d759fd45/src/libYARP_dev/src/idl_generated_code/yarp/dev/StickData.cpp)
- src/devices/messages/IJoypadControlMsg: [.thrift](https://github.com/robotology/yarp/blob/fcef5d167b761fbb8e578c395ec24808d759fd45/src/devices/messages/IJoypadControlMsg/IJoypadControlMsgs.thrift#L26-L30), [.h](https://github.com/robotology/yarp/blob/fcef5d167b761fbb8e578c395ec24808d759fd45/src/devices/messages/IJoypadControlMsg/idl_generated_code/yarp/dev/StickData.h), [.cpp](https://github.com/robotology/yarp/blob/fcef5d167b761fbb8e578c395ec24808d759fd45/src/libYARP_dev/src/idl_generated_code/yarp/dev/StickData.cpp)

However, the `getStickDoF` method still exists:

https://github.com/robotology/yarp/blob/65e90e42a5c8cc3a43e87b54191a5f704cb06622/src/libYARP_dev/src/yarp/dev/IJoypadController.h#L87-L94

[...]

PR #3376, opened 2026-08-25
Fix IJoypadControl: removed getStickDoF

randaz81:
remove method `IJoypadControl::getStickDoF(size_t  stick_id, size_t& DoF)`

Related to: https://github.com/robotology/yarp/issues/3370


randaz81:
Added method  yarp::dev::ReturnValue getAllAxes(std::vector<double>& values)

PR #3386, opened 2026-09-08
Fix signed overflow when the Stamp sequence counter wraps

vpfkfl753:
Both `Stamp::update()` overloads increment the sequence number before checking whether it has reached the limit. At `INT_MAX`, this causes signed integer overflow before the counter can wrap to zero.

Check the limit before incrementing. Adds rollover tests for both overloads. The Stamp and Header tests pass with `-fsanitize=signed-integer-overflow -fno-sanitize-recover=signed-integer-overflow`; both new cases fail on the original code.

### DIFF
--- a/doc/release/yarp_4_0/stamp_sequence_rollover.md
+++ b/doc/release/yarp_4_0/stamp_sequence_rollover.md
@@ -1,0 +1,4 @@
+stamp_sequence_rollover {#yarp_4_0}
+-----------------------
+
+* Fixed signed overflow when `yarp::os::Stamp` wraps its sequence number to zero.
--- a/src/libYARP_os/src/yarp/os/Stamp.cpp
+++ b/src/libYARP_os/src/yarp/os/Stamp.cpp
@@ -125,18 +125,20 @@ void yarp::os::Stamp::update()
 {
     double now = Time::now();
 
-    sequenceNumber++;
-    if (sequenceNumber > getMaxCount() || sequenceNumber < 0) {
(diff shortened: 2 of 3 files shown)


vpfkfl753:
Rebased onto current `master` (`d02afa6`). The Stamp fix and regression tests are unchanged.

vpfkfl753:
Sorry, I missed the target branch in my earlier rebase. The PR now targets `yarp-4.0`, and the release note is in `doc/release/yarp_4_0/`.

The Stamp fix and tests are unchanged. Rebuilt on that branch and reran Stamp/Header with signed-integer-overflow sanitization: all 105 assertions pass.

randaz81 (approved):

vpfkfl753:
Thanks for the review. I checked the CI failures. `os::StampTest` passes in the Ubuntu 22.04 Valgrind job. In the [Ubuntu 26.04 Valgrind job](https://github.com/robotology/yarp/actions/runs/34264818911/job/103216473795), all 27 assertions, including both rollover cases, pass; Valgrind then reports leaks in the dynamic-loader path. The no-ACE job also passes the assertions and reports a loader leak.

The other failures include missing MSVC/Qt5 files, unavailable Homebrew `gts` bottles, interrupted dependency downloads, and timeouts/Thrift failures in other tests. The [PR Analysis failure](https://github.com/robotology/yarp/actions/runs/34615257207/job/103317555375) is a missing `codacy-coverage-data` artifact after the coverage build failed.

PR #3388, opened 2026-09-11
Feature/python binding logger

elandini84:
* `LogComponent` added to bindings to improve yarp logging system integration


randaz81 (approved):

randaz81 (changes requested):
Looks like there are some errors, maybe related to Java on Ubunt22.04. To be investigated

